PR #125, opened 2017-12-12
Init implementation for the Open API validation

xuzhg:
@PerthCharern @darrelmiller 

**General design:**

1. Can validation each part individually. for example, Developer create a <OpenApiContact> object, he can call the validate api to validate the created object.

2. Validate rule can be customized. For example, developer can create their own validation rules.

**Detail:**

1. OpenApiElementValidator.cs has the public apis

2. .\Validations\Visitors has the visitor classes for each element.

3. .\Validations\Rules has the rule classes for each element.

### DIFF
--- a/src/Microsoft.OpenApi/Validations/ValidationContext.cs
+++ b/src/Microsoft.OpenApi/Validations/ValidationContext.cs
@@ -1,0 +1,80 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using Microsoft.OpenApi.Validations.Rules;
+
+namespace Microsoft.OpenApi.Validations
+{
+    /// <summary>
+    /// The validation context.
+    /// </summary>
+    public class ValidationContext
+    {
+        private readonly IList<ValidationError> _errors = new List<ValidationError>();
+
+        /// <summary>
+        /// Initializes the <see cref="ValidationContext"/> class.
+        /// </summary>
+        /// <param name="ruleSet"></param>
+        public ValidationContext(ValidationRuleSet ruleSet)
+        {
+            RuleSet = ruleSet ?? throw Error.ArgumentNull(nameof(ruleSet));
+        }
+
+        /// <summary>
+        /// Gets the rule set.
+        /// </summary>
+        public ValidationRuleSet RuleSet { get; }
+
+        /// <summary>
+        /// Gets the validation errors.
+        /// </summary>
+        public IEnumerable<ValidationError> Errors
+        {
+            get
+            {
+                return _errors;
+            }
+        }
+
+        /// <summary>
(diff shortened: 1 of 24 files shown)


PerthCharern:
Darrel, I don't think extending rules is out of scope per se. One purpose of the core + readers libs is to be the "base" for other things that can be built (e.g. more types of readers), and in this case, types of validation rules. Given that we make the "default rule set" public and with the overload Sam is providing for the entry point, I think we can allow our customers to build their own rules if needs be. They might want to restrict certain sets of strings (path strings, etc.) to enforce consistency just among their sets of APIs.

---
In reply to: [351419521](https://github.com/Microsoft/OpenAPI.NET/pull/125#issuecomment-351419521) [](ancestors = 351419521)

PerthCharern:
I understand though that this extensibility is kind of a stretch (since extensions in other dimensions like the reader still adheres strictly to OpenAPI spec), but I don't see the harm of providing this potentiality.

---
In reply to: [351804977](https://github.com/Microsoft/OpenAPI.NET/pull/125#issuecomment-351804977) [](ancestors = 351804977,351419521)

src/Microsoft.OpenApi/Validations/ValidationContext.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        internal void Push(string segment)
+        {
+            this._path.Push(segment);
+        }
+
+        internal void Pop()
+        {
+            this._path.Pop();
+        }
```
PerthCharern: Can we make _path a property and simply call Push/Pop on _path itself? I don't think we need to make these methods on the context.

src/Microsoft.OpenApi/Validations/ValidationError.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// </summary>
+        public string ErrorMessage { get; private set; }
+
+        /// <summary>
+        /// Returns the whole error message.
+        /// </summary>
+        /// <returns>The error string.</returns>
+        public override string ToString()
+        {
+            return "ErroCode: " + ErrorCode + ", " + ErrorPath + " | " + ErrorMessage;
```
PerthCharern: >"ErroCode: " + ErrorCode + ", " + ErrorPath + " | " + ErrorMessage; [](start = 19, length = 67)

This is a little awkward. Maybe we don't need this?

src/Microsoft.OpenApi/Validations/ValidationContext.cs
xuzhg: @PerthCharern I make them as internal and the methods itself can let me coding less character. 😆

src/Microsoft.OpenApi/Validations/ValidationError.cs
xuzhg: Do you think it's helpful for debug?

xuzhg:
@PerthCharern Thanks! Ok, Let me continue on this PR (even though I think #133 is better. 😆 ).

xuzhg:
@PerthCharern @darrelmiller I have to close this because the origin link failed. 

I created a new PR to continue: https://github.com/Microsoft/OpenAPI.NET/pull/152

PR #128, opened 2017-12-12
Write full objects in components / Fix components missing issue when V2 document is parsed

### LINKED ISSUE #120: Writing out components that are referenced, writes out $ref, not full object

PerthCharern:
Fixes #120 

- Writing anything in Components or Definitions/Parameters/SecuritySchemes/etc will always yield full objects, regardless of whether Reference exists.

- Unit tests for all the Referenceable objects. Tests verify that both ways of writing an object (without Reference or with Reference) work.

- Ensure that Components is not overwritten in `OpenApiV2Deserializer`.


xuzhg:
@PerthCharern  please take a look at : #157

PR #133, opened 2017-12-13
Add the validator attribute, validation rule and validator

xuzhg:
@PerthCharern @darrelmiller
General design:
Can validation each part individually. for example, Developer create a object, he can call the validate api to validate the created object.
Validate rule can be customized. For example, developer can create their own validation rules.
Detail:
OpenApiElementValidator.cs has the public apis
.\Validations\Validator has the validator classes for each element.
.\Validations\Rules has the rule classes for each element.
Each model class has an attribute to identify the validator.


PerthCharern:
Given that both #125 and #133 are both equally extensible, I would go with #125 since it's easier to understand and maintain. I'll go ahead and make specific comments in #125.

xuzhg:
Close this because we are moving to #152.

PR #147, opened 2017-12-14
Update NuGet-related sections in csproj for nuget.org release

PerthCharern:


PerthCharern:
This is based on the documentation here:

https://docs.microsoft.com/en-us/nuget/schema/msbuild-targets#pack-target

PR #152, opened 2017-12-15
Implementation for the Open API validation

xuzhg:
See #125:

General design:
Can validation each part individually. for example, Developer create a object, he can call the validate api to validate the created object.
Validate rule can be customized. For example, developer can create their own validation rules.
Detail:
OpenApiElementValidator.cs has the public apis
.\Validations\Visitors has the visitor classes for each element.
.\Validations\Rules has the rule classes for each element.

### DIFF
--- a/src/Microsoft.OpenApi/Validations/OpenApiElementValidator.cs
+++ b/src/Microsoft.OpenApi/Validations/OpenApiElementValidator.cs
@@ -1,7 +1,5 @@
-﻿// ------------------------------------------------------------
-//  Copyright (c) Microsoft Corporation.  All rights reserved.
-//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
-// ------------------------------------------------------------
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
 
 using System.Collections.Generic;
 using System.Linq;
--- a/src/Microsoft.OpenApi/Validations/OpenApiElementValidator.cs
+++ b/src/Microsoft.OpenApi/Validations/OpenApiElementValidator.cs
@@ -26,7 +26,7 @@ namespace Microsoft.OpenApi.Validations
         }
 
         /// <summary>
-        /// Validate the Open API element.
+        /// Validate the Open API element and output the whole validation errors.
         /// </summary>
         /// <typeparam name="T">The Open API element type.</typeparam>
         /// <param name="element">The Open API element.</param>
@@ -40,7 +40,7 @@ namespace Microsoft.OpenApi.Validations
         }
 
         /// <summary>
-        ///  Validate the Open API element.
+        ///  Validate the Open API element with the given rule set and output the whole validation errors.
         /// </summary>
         /// <typeparam name="T">The Open API element type.</typeparam>
         /// <param name="element">The Open API element.</param>
--- a/src/Microsoft.OpenApi/Validations/OpenApiElementValidator.cs
+++ b/src/Microsoft.OpenApi/Validations/OpenApiElementValidator.cs
@@ -1,0 +1,75 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// ------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.OpenApi.Interfaces;
+using Microsoft.OpenApi.Validations.Rules;
+
+namespace Microsoft.OpenApi.Validations
+{
+    /// <summary>
+    /// The public APIs to validate the Open API element.
+    /// </summary>
+    public static class OpenApiElementValidator
+    {
+        /// <summary>
+        /// Validate the Open API element.
+        /// </summary>
+        /// <typeparam name="T">The Open API element type.</typeparam>
+        /// <param name="element">The Open API element.</param>
+        /// <returns>True means no errors, otherwise with errors.</returns>
+        public static bool Validate<T>(this T element) where T : IOpenApiElement
+        {
+            IEnumerable<ValidationError> errors;
+            return element.Validate(out errors);
+        }
+
+        /// <summary>
+        /// Validate the Open API element.
+        /// </summary>
+        /// <typeparam name="T">The Open API element type.</typeparam>
+        /// <param name="element">The Open API element.</param>
+        /// <param name="errors">The output errors.</param>
+        /// <returns>True means no errors, otherwise with errors.</returns>
+        public static bool Validate<T>(this T element, out IEnumerable<ValidationError> errors)
+            where T : IOpenApiElement
+        {
+            ValidationRuleSet ruleSet = ValidationRuleSet.DefaultRuleSet;
+            return element.Validate(ruleSet, out errors);
+        }
+
+        /// <summary>
+        ///  Validate the Open API element.
+        /// </summary>
+        /// <typeparam name="T">The Open API element type.</typeparam>
+        /// <param name="element">The Open API element.</param>
+        /// <param name="ruleSet">The input rule set.</param>
+        /// <param name="errors">The output errors.</param>
+        /// <returns>True means no errors, otherwise with errors.</returns>
+        public static bool Validate<T>(this T element, ValidationRuleSet ruleSet, out IEnumerable<ValidationError> errors)
+            where T : IOpenApiElement
+        {
+            errors = null;
+            if (element == null)
+            {
+                return true;
+            }
+
+            if (ruleSet == null)
+            {
+                ruleSet = ValidationRuleSet.DefaultRuleSet;
+            }
+
+            ValidationContext context = new ValidationContext(ruleSet);
+
+            context.Validate(element);
+
+            errors = context.Errors;
+
+            return !errors.Any();
+        }
+    }
+}
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiExternalDocsRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiExternalDocsRules.cs
@@ -10,12 +10,13 @@ namespace Microsoft.OpenApi.Validations.Rules
     /// <summary>
     /// The validation rules for <see cref="OpenApiExternalDocs"/>.
     /// </summary>
+    [OpenApiRule]
     internal static class OpenApiExternalDocsRules
     {
         /// <summary>
         /// Validate the field is required.
         /// </summary>
-        public static readonly ValidationRule<OpenApiExternalDocs> FieldIsRequired =
+        public static ValidationRule<OpenApiExternalDocs> FieldIsRequired =>
             new ValidationRule<OpenApiExternalDocs>(
                 (context, item) =>
                 {
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiExternalDocsRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiExternalDocsRules.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Properties;
+
+namespace Microsoft.OpenApi.Validations.Rules
+{
+    /// <summary>
+    /// The validation rules for <see cref="OpenApiExternalDocs"/>.
+    /// </summary>
+    internal static class OpenApiExternalDocsRules
+    {
+        /// <summary>
+        /// Validate the field is required.
+        /// </summary>
+        public static readonly ValidationRule<OpenApiExternalDocs> FieldIsRequired =
+            new ValidationRule<OpenApiExternalDocs>(
+                (context, item) =>
+                {
+                    // title
+                    context.Push("url");
+                    if (item.Url == null)
+                    {
+                        ValidationError error = new ValidationError(ErrorReason.Required, context.PathString,
+                            String.Format(SRResource.Validation_FieldIsRequired, "url", "External Documentation"));
+                        context.AddError(error);
+                    }
+                    context.Pop();
+                });
+
+        // add more rule.
+    }
+}
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiExternalDocsRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiExternalDocsRules.cs
@@ -20,7 +20,7 @@ namespace Microsoft.OpenApi.Validations.Rules
             new ValidationRule<OpenApiExternalDocs>(
                 (context, item) =>
                 {
-                    // title
+                    // url
                     context.Push("url");
                     if (item.Url == null)
                     {
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
@@ -20,7 +20,7 @@ namespace Microsoft.OpenApi.Validations.Rules
             new ValidationRule<OpenApiResponse>(
                 (context, response) =>
                 {
-                    // title
+                    // description
                     context.Push("description");
                     if (String.IsNullOrEmpty(response.Description))
                     {
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Properties;
+
+namespace Microsoft.OpenApi.Validations.Rules
+{
+    /// <summary>
+    /// The validation rules for <see cref="OpenApiResponse"/>.
+    /// </summary>
+    internal static class OpenApiResponseRules
+    {
+        /// <summary>
+        /// Validate the field is required.
+        /// </summary>
+        public static readonly ValidationRule<OpenApiResponse> FieldIsRequired =
+            new ValidationRule<OpenApiResponse>(
+                (context, response) =>
+                {
+                    // title
+                    context.Push("description");
+                    if (String.IsNullOrEmpty(response.Description))
+                    {
+                        ValidationError error = new ValidationError(ErrorReason.Required, context.PathString,
+                            String.Format(SRResource.Validation_FieldIsRequired, "description", "response"));
+                        context.AddError(error);
+                    }
+                    context.Pop();
+                });
+
+        // add more rule.
+    }
+}
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
@@ -10,12 +10,13 @@ namespace Microsoft.OpenApi.Validations.Rules
     /// <summary>
     /// The validation rules for <see cref="OpenApiResponse"/>.
     /// </summary>
+    [OpenApiRule]
     internal static class OpenApiResponseRules
     {
         /// <summary>
         /// Validate the field is required.
         /// </summary>
-        public static readonly ValidationRule<OpenApiResponse> FieldIsRequired =
+        public static ValidationRule<OpenApiResponse> FieldIsRequired =>
             new ValidationRule<OpenApiResponse>(
                 (context, response) =>
                 {
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiRuleAttribute.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiRuleAttribute.cs
@@ -1,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System;
+
+namespace Microsoft.OpenApi.Validations.Rules
+{
+    /// <summary>
+    /// The Validator attribute.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
+    internal class OpenApiRuleAttribute : Attribute
+    {
+    }
+}
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Properties;
+
+namespace Microsoft.OpenApi.Validations.Rules
+{
+    /// <summary>
+    /// The validation rules for <see cref="OpenApiTag"/>.
+    /// </summary>
+    public static class OpenApiTagRules
+    {
+        /// <summary>
+        /// REQUIRED.
+        /// </summary>
+        public static readonly ValidationRule<OpenApiTag> FieldIsRequired =
+            new ValidationRule<OpenApiTag>(
+                (context, tag) =>
+                {
+                    context.Push("name");
+                    if (String.IsNullOrEmpty(tag.Name))
+                    {
+                        ValidationError error = new ValidationError(ErrorReason.Required, context.PathString,
+                            String.Format(SRResource.Validation_FieldIsRequired, "name", "Tag"));
+                        context.AddError(error);
+                    }
+                    context.Pop();
+                });
+
+        // add more rules
+    }
+}
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
@@ -14,7 +14,7 @@ namespace Microsoft.OpenApi.Validations.Rules
     public static class OpenApiTagRules
     {
         /// <summary>
-        /// REQUIRED.
+        /// Validate the field is required.
         /// </summary>
         public static ValidationRule<OpenApiTag> FieldIsRequired =>
             new ValidationRule<OpenApiTag>(
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
@@ -10,12 +10,13 @@ namespace Microsoft.OpenApi.Validations.Rules
     /// <summary>
     /// The validation rules for <see cref="OpenApiTag"/>.
     /// </summary>
+    [OpenApiRule]
     public static class OpenApiTagRules
     {
         /// <summary>
         /// REQUIRED.
         /// </summary>
-        public static readonly ValidationRule<OpenApiTag> FieldIsRequired =
+        public static ValidationRule<OpenApiTag> FieldIsRequired =>
             new ValidationRule<OpenApiTag>(
                 (context, tag) =>
                 {
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
@@ -24,7 +24,7 @@ namespace Microsoft.OpenApi.Validations.Rules
                     if (String.IsNullOrEmpty(tag.Name))
                     {
                         ValidationError error = new ValidationError(ErrorReason.Required, context.PathString,
-                            String.Format(SRResource.Validation_FieldIsRequired, "name", "Tag"));
+                            String.Format(SRResource.Validation_FieldIsRequired, "name", "tag"));
                         context.AddError(error);
                     }
                     context.Pop();
--- a/src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
@@ -11,7 +11,7 @@ namespace Microsoft.OpenApi.Validations.Rules
         /// Input string must be in the format of an email address
         /// </summary>
         /// <param name="input">The input string.</param>
-        /// <returns></returns>
+        /// <returns>True if it's an email address. Otherwise False.</returns>
         public static bool IsEmailAddress(this string input)
         {
             if (String.IsNullOrEmpty(input))
--- a/src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
@@ -1,0 +1,40 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.OpenApi.Validations.Rules
+{
+    internal static class RuleHelpers
+    {
+        /// <summary>
+        /// Input string must be in the format of an email address
+        /// </summary>
+        /// <param name="input">The input string.</param>
+        /// <returns></returns>
+        public static bool IsEmailAddress(this string input)
+        {
+            if (String.IsNullOrEmpty(input))
+            {
+                return false;
+            }
+
+            var splits = input.Split('@');
+            if (splits.Length != 2)
+            {
+                return false;
+            }
+
+            if (String.IsNullOrEmpty(splits[0]) || String.IsNullOrEmpty(splits[1]))
+            {
+                return false;
+            }
+
+            // Add more rules.
+
+            return true;
+        }
+    }
+}
--- a/src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
@@ -1,7 +1,5 @@
-﻿// ------------------------------------------------------------
-//  Copyright (c) Microsoft Corporation.  All rights reserved.
-//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
-// ------------------------------------------------------------
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
 
 using System;
 
--- a/src/Microsoft.OpenApi/Validations/ValidationContext.cs
+++ b/src/Microsoft.OpenApi/Validations/ValidationContext.cs
@@ -1,7 +1,5 @@
-﻿// ------------------------------------------------------------
-//  Copyright (c) Microsoft Corporation.  All rights reserved.
-//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
-// ------------------------------------------------------------
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
 
 using System;
 using System.Collections.Generic;
@@ -72,7 +70,7 @@ namespace Microsoft.OpenApi.Validations
         {
             get
             {
-                return String.Join("/", _path);
+                return "#/" + String.Join("/", _path);
             }
         }
         #endregion
--- a/src/Microsoft.OpenApi/Validations/ValidationContext.cs
+++ b/src/Microsoft.OpenApi/Validations/ValidationContext.cs
@@ -1,0 +1,80 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// ------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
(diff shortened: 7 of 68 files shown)


src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            OpenApiTagRules.FieldIsRequired,
+
+            OpenApiResponseRules.FieldIsRequired,
+
+            OpenApiOAuthFlowRules.FieldIsRequired,
+
+            OpenApiExtensibleRules.ExtensionNameMustStartWithXDash,
+            // add more default rules.
+        };
```
PerthCharern: This will be difficult to track. People will forget to put the new rules in here since this list is in a different file than the rules themselves.

I am thinking of this:

- Add DefaultRuleSet for every OpenApiXXXRules file
- This main DefaultRuleSet is a concatenation of all DefaultRuleSet's

What do you think?

PerthCharern: In fact, we might not need this concatenated list at all since it will be broken down in Add() below anyway.

---
In reply to: [157320269](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157320269) [](ancestors = 157320269)

xuzhg: 1. Basically, I want to use the attribute for each "DOM" classes same as #133. 
    Let me think it more.

2. Why do you think it will break the Add() method? That's in deed will call "Add()" method.

darrelmiller:
Can I make a suggestion for doing these kind of design reviews?  It would be helpful to see the code changes for a couple of examples rather than a complete implementation.  Trying to review changes to 60 files in the GitHub UI is harder than it should be.

xuzhg:
@darrelmiller Yes.  🤣 

Can you use the CodeFlow for the large PR? 
You can try to run the following command at your local cmd window to open the code flow:

```cmd
\\codeflow\public\cf.cmd openGitHubPR -webUrl https://github.com/Microsoft/OpenAPI.NET/pull/152
```

src/Microsoft.OpenApi/Validations/OpenApiElementValidator.cs
```diff
@@ -0,0 +1,73 @@
...
+        public static bool Validate<T>(this T element, out IEnumerable<ValidationError> errors)
+            where T : IOpenApiElement
+        {
+            ValidationRuleSet ruleSet = ValidationRuleSet.DefaultRuleSet;
+            return element.Validate(ruleSet, out errors);
+        }
+
+        /// <summary>
+        ///  Validate the Open API element.
+        /// </summary>
```
darrelmiller: Ok, so here's the deal!  If I have to put up with these horrible comment blocks, I would like to see comments that actually justify their existence.  So let's make the comment explain why I would pick one overload over the other so that I don't have to guess the reason based on what parameters are there.

xuzhg: Great point. I will modify it.

---
In reply to: [157324617](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157324617) [](ancestors = 157324617)

src/Microsoft.OpenApi/Validations/ValidationContext.cs
```diff
@@ -0,0 +1,78 @@
...
+        private readonly IList<ValidationError> _errors = new List<ValidationError>();
+
+        /// <summary>
+        /// Initializes the <see cref="ValidationContext"/> class.
+        /// </summary>
+        /// <param name="ruleSet"></param>
+        public ValidationContext(ValidationRuleSet ruleSet)
+        {
+            RuleSet = ruleSet ?? throw Error.ArgumentNull(nameof(ruleSet));
+        }
```
darrelmiller: This is a personal preference, but I really don't like this syntax.  I'd much rather see a guard clause and a separate assignment.  Could we even use the Code Contracts library that would do static analysis?

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
darrelmiller: If you wanted to create an OpenApiValidationRuleAttribute and annotate each rule with that attribute and then have a discovery mechanism that searched the assembly to find all classes with that attribute, then I would have no objection to that.

---
In reply to: [157321750](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157321750) [](ancestors = 157321750)

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Linq;
+using Microsoft.OpenApi.Interfaces;
+
+namespace Microsoft.OpenApi.Validations.Visitors
+{
+    /// <summary>
+    /// The base visitor class for Open Api elements.
+    /// </summary>
+    internal abstract class VisitorBase<T> : IVisitor where T : IOpenApiElement
+    {
```
darrelmiller: This class has no state and we have an IVisitor interface. Is there any reason these two methods couldn't just be library functions?  I don't think we need a base class for all the visitors.

xuzhg: Each visitor has two parts:

1. Visit itself's property
2. Go to visit it's Children.


Where:
a. Visit() method is for the #1, it has the same logic for all elements. So, put its at the base class.

b. Next() method is for the #2, it has different logic for all elements. So, each element visitor class should override Next() method to list the children. 

---
In reply to: [157326815](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157326815) [](ancestors = 157326815)

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
darrelmiller: Overall I like the direction of this PR.  I think by adding a simple marker attribute to each rule, you can reflect over the rules and built a default ruleset.  You should be able to determine the target type from the generic return type of the rule.

This should address the concern brought up by @PerthCharern  that someone might forget to add the rule to the ruleset.

src/Microsoft.OpenApi/Validations/OpenApiElementValidator.cs
xuzhg: modified.

---
In reply to: [157324939](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157324939) [](ancestors = 157324939,157324617)

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
PerthCharern: @xuzhg I didn't mean it will break the Add() method. I meant the list didn't have to be created because it would be enumerated in the Add() method anyway, so we can find a way to make it less complicated.

Anyway, I like @darrelmiller 's suggestion here. That would put the concern of when each rule should be used back to the rule itself.

xuzhg:
@darrelmiller @PerthCharern I added the attribute for Rule classes. Would you please take a look at https://github.com/Microsoft/OpenAPI.NET/pull/152/commits/da5f90a05a0c3c931cb8ad214c2dda9b0191aee3

xuzhg:
@darrelmiller @PerthCharern I also added the attribute for visitor classes. Would you please take a look at :

https://github.com/Microsoft/OpenAPI.NET/pull/152/commits/4a6ca43701b589a359794a9fa99e3ff6e1c0d33a

src/Microsoft.OpenApi/Validations/Visitors/OpenApiVisitorSet.cs
```diff
@@ -14,39 +16,23 @@ namespace Microsoft.OpenApi.Validations.Visitors
...
-            { typeof(OpenApiServerVariable), new ServerVariableVisitor() },
-            { typeof(OpenApiServer), new ServerVisitor() },
-            { typeof(OpenApiTag), new TagVisitor() },
-            { typeof(OpenApiXml), new XmlVisitor() }
-        };
+            get
+            {
+                if (_visitors == null)
+                {
+                    _visitors = new Lazy<IDictionary<Type, IVisitor>>(() => BuildVisitorSet(), isThreadSafe: false).Value;
```
xuzhg: Maybe Lazy is unnecessary?

src/Microsoft.OpenApi/Validations/Rules/OpenApiRuleAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT license.
+
+using System;
+
+namespace Microsoft.OpenApi.Validations.Rules
+{
+    /// <summary>
+    /// The Validator attribute.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
```
PerthCharern: >Class [](start = 37, length = 5)

Why class attribute vs field/property attribute?

xuzhg: It's easy to target "Rule" Class. 

If it's target on the Rule property, we should go through all the properties in all the types, it's time consuming.

src/Microsoft.OpenApi/Validations/ValidationContext.cs
PerthCharern: Will we have to enforce this though? It's a valid and acceptable C# syntax as far as I know, and I'm not sure if we should enforce it one way or another when someone submits a PR.

---
In reply to: [157325008](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157325008) [](ancestors = 157325008)

PerthCharern:
You just meant the logic to traverse all the Visitors, not a C# attribute, right? I want to make sure I'm not missing something.

---
In reply to: [352585589](https://github.com/Microsoft/OpenAPI.NET/pull/152#issuecomment-352585589) [](ancestors = 352585589)

src/Microsoft.OpenApi/Validations/Visitors/OpenApiVisitorSet.cs
PerthCharern: I would say the underlying operation is not that time consuming/complex, so I think it's unnecessary. But I don't really mind using it if you prefer.

---
In reply to: [157624690](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157624690) [](ancestors = 157624690)

xuzhg:
@PerthCharern For the visitor, Yes. I have the interface `IVisitor`. That's enough.

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        [Fact]
+        public void GetVisitorThrowsUnknowElementType()
+        {
+            // Arrange & Act
+            Action test = () => OpenApiVisitorSet.GetVisitor(typeof(OpenApiVisitorSetTests));
+
+            // Assert
+            var exception = Assert.Throws<OpenApiException>(test);
+            Assert.Equal("Can not find visitor type registered for type 'Microsoft.OpenApi.Validations.Visitors.Tests.OpenApiVisitorSetTests'.",
```
PerthCharern: >Can not find visitor type registered for type [](start = 26, length = 45)

This should be a public const in the main project and referenced here.

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.OpenApi.Validations.Visitors.Tests
+{
+    public class OpenApiVisitorSetTests
+    {
+        [Fact]
+        public void VisitorsPropertyReturnsTheCorrectVisitorList()
+        {
+            for (int i = 0; i < 5; i++) // 5 is just a magic number
+            {
```
PerthCharern: Not sure what the loop helps with. Maybe remove?

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
```diff
@@ -0,0 +1,55 @@
...
+                // Arrange & Act
+                var visitors = OpenApiVisitorSet.Visitors;
+
+                // Assert
+                Assert.NotNull(visitors);
+                Assert.NotEmpty(visitors);
+                Assert.Equal(29, visitors.Count);
```
PerthCharern: >Assert.Equal(29, visitors.Count); [](start = 16, length = 33)

Provide comment

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            var exception = Assert.Throws<OpenApiException>(test);
+            Assert.Equal("Can not find visitor type registered for type 'Microsoft.OpenApi.Validations.Visitors.Tests.OpenApiVisitorSetTests'.",
+                exception.Message);
+        }
+
+        [Theory]
+        [InlineData(typeof(OpenApiDocument), typeof(DocumentVisitor))]
+        [InlineData(typeof(OpenApiInfo), typeof(InfoVisitor))]
+        [InlineData(typeof(OpenApiXml), typeof(XmlVisitor))]
+        [InlineData(typeof(OpenApiComponents), typeof(ComponentsVisitor))]
```
PerthCharern: >[InlineData(typeof(OpenApiComponents), typeof(ComponentsVisitor))] [](start = 8, length = 66)

Can we add all types here? It's a one time thing, and it will help ensure no mistakes in the real code.

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -0,0 +1,149 @@
...
+                    if (validationRuleType.IsAssignableFrom(property.PropertyType))
+                    {
+                        var propertyValue = property.GetValue(null); // static property
+                        ValidationRule rule = propertyValue as ValidationRule;
+                        if (rule != null)
+                        {
+                            ruleSet.Add(rule);
+                        }
+                    }
+                }
```
PerthCharern: It's a little risky to me that we rely on the type and not an attribute assigned to each property.

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            IOpenApiExtensible extensbile = element as IOpenApiExtensible;
+            if (extensbile != null)
+            {
+                var rules = context.RuleSet.Where(r => r.ElementType == typeof(IOpenApiExtensible));
+                foreach (var rule in rules)
+                {
+                    rule.Evaluate(context, extensbile);
+                }
+            }
```
PerthCharern: Why is this not done in Visit above?

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// The base visitor class for Open Api elements.
+    /// </summary>
+    internal abstract class VisitorBase<T> : IVisitor where T : IOpenApiElement
+    {
+        /// <summary>
+        /// Visit the element itself.
+        /// </summary>
+        /// <param name="context">The validation context.</param>
+        /// <param name="item">The element.</param>
+        public void Visit(ValidationContext context, object item)
```
PerthCharern: >object [](start = 53, length = 6)

Can this be T?

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        /// <summary>
+        /// Visit the children.
+        /// </summary>
+        /// <param name="context">The validation context.</param>
+        /// <param name="element">The element.</param>
+        protected virtual void Next(ValidationContext context, T element)
+        {
+            IOpenApiExtensible extensbile = element as IOpenApiExtensible;
```
PerthCharern: >extensbile [](start = 31, length = 10)

typo, extensible

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
PerthCharern: It is okay for now though. I don't have a better way than to tag all properties, which is something we want to avoid.

---
In reply to: [157650194](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157650194) [](ancestors = 157650194)

test/Microsoft.OpenApi.Tests/Validations/OpenApiInfoValidationTests.cs
```diff
@@ -0,0 +1,31 @@
...
+
+            // Act
+            bool result = info.Validate(out errors);
+
+            // Assert
+            Assert.False(result);
+            Assert.NotNull(errors);
+
+            Assert.Equal(new[] { "The field 'url' in 'info' object is REQUIRED.",
+                "The field 'version' in 'info' object is REQUIRED."}, errors.Select(e => e.ErrorMessage));
```
PerthCharern: These strings in all test classes should come from the main project, so that we don't need to keep track of the strings in two places.

src/Microsoft.OpenApi/Validations/ValidationError.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// </summary>
+        public string ErrorMessage { get; private set; }
+
+        /// <summary>
+        /// Returns the whole error message.
+        /// </summary>
+        /// <returns>The error string.</returns>
+        public override string ToString()
+        {
+            return "ErroCode: " + ErrorCode + ", " + ErrorPath + " | " + ErrorMessage;
```
PerthCharern: >ErroCode [](start = 20, length = 8)

typo, ErrorCode

test/Microsoft.OpenApi.Tests/Validations/OpenApiResponseValidationTests.cs
```diff
@@ -0,0 +1,29 @@
...
+            OpenApiResponse response = new OpenApiResponse();
+
+            // Act
+            bool result = response.Validate(out errors);
+
+            // Assert
+            Assert.False(result);
+            Assert.NotNull(errors);
+            ValidationError error = Assert.Single(errors);
+            Assert.Equal("The field 'description' in 'response' object is REQUIRED.", error.ErrorMessage);
```
PerthCharern: Should we check that the path and the reason are also as expected?

src/Microsoft.OpenApi/Validations/Visitors/PathItemVisitor.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Visit the children of the <see cref="OpenApiPathItem"/>.
+        /// </summary>
+        /// <param name="context">The validation context.</param>
+        /// <param name="pathItem">The <see cref="OpenApiPathItem"/>.</param>
+        protected override void Next(ValidationContext context, OpenApiPathItem pathItem)
+        {
+            Debug.Assert(context != null);
+            Debug.Assert(pathItem != null);
```
PerthCharern: We haven't been using this anywhere else. Can we be consistent and use the normal null checks?

src/Microsoft.OpenApi/Validations/Rules/OpenApiExternalDocsRules.cs
```diff
@@ -0,0 +1,36 @@
...
+    internal static class OpenApiExternalDocsRules
+    {
+        /// <summary>
+        /// Validate the field is required.
+        /// </summary>
+        public static ValidationRule<OpenApiExternalDocs> FieldIsRequired =>
+            new ValidationRule<OpenApiExternalDocs>(
+                (context, item) =>
+                {
+                    // title
```
PerthCharern: >title [](start = 23, length = 5)

url

src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
```diff
@@ -0,0 +1,36 @@
...
+    internal static class OpenApiResponseRules
+    {
+        /// <summary>
+        /// Validate the field is required.
+        /// </summary>
+        public static ValidationRule<OpenApiResponse> FieldIsRequired =>
+            new ValidationRule<OpenApiResponse>(
+                (context, response) =>
+                {
+                    // title
```
PerthCharern: >title [](start = 23, length = 5)

description

src/Microsoft.OpenApi/Validations/Rules/OpenApiTagRules.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.OpenApi.Validations.Rules
+{
+    /// <summary>
+    /// The validation rules for <see cref="OpenApiTag"/>.
+    /// </summary>
+    [OpenApiRule]
+    public static class OpenApiTagRules
+    {
+        /// <summary>
+        /// REQUIRED.
```
PerthCharern: >REQUIRED [](start = 12, length = 8)

change comment

src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Microsoft.OpenApi.Validations.Rules
+{
+    internal static class RuleHelpers
+    {
+        /// <summary>
+        /// Input string must be in the format of an email address
+        /// </summary>
+        /// <param name="input">The input string.</param>
+        /// <returns></returns>
```
PerthCharern: remove or fill in

src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
```diff
@@ -0,0 +1,38 @@
...
+        public static bool IsEmailAddress(this string input)
```
PerthCharern: >IsEmailAddress [](start = 27, length = 14)

The accepted answer here looks reasonable to me:

[https://stackoverflow.com/questions/1365407/c-sharp-code-to-validate-email-address](https://stackoverflow.com/questions/1365407/c-sharp-code-to-validate-email-address)

It is though a little more forgiving than it has to be, but I believe it is strict enough for our purpose.

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
darrelmiller: Part of the reason I initially used the visitor pattern was so that we could define the traversal of the OpenAPI DOM just once and then have different visitors for different purposes.  Validation is the obvious Visitor, but code generation could be another.  Either client code gen, or server code scaffolding.  
The Visitor related classes in this PR are specifically tied to validation via the ValidationContext.  Even the abstract base class and IVisitor interface.

If you look at the Visitor pattern https://en.wikipedia.org/wiki/Visitor_pattern here, the logic to walk the object graph is not in the Visitor, it is in the object graph itself.  Rather than embed the Accept methods into our DOM, I created the walker to house all of the traversal logic.

Also, you will see the Visitor implements all the visit methods for all the elements it is interested in.  This allows the Visitor to hold shared state whilst doing the traversal.  It also plays the role of the context object.

With the current approach, ignoring the fact that it is tied to validation, each new "visiting" capability needs to create classes for each and every element and then pass the context between them.

I'm going to see if I can use my walker and validationVisitor to execute your rules classes and we can compare the end result.

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
xuzhg: @PerthCharern 

Yes. I can use the
`
Assert.Equal(`String.Format(SRResource.xxxxx, typeof(xxx).FullName)`, exception.Message);
`
However, it looks not good than the plain text in the test project. Because it can help us verify the error message thrown as our expected.  What do you think?

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
xuzhg: I just want to test multiple retrieving the property "Visitors" can work as expected.

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
xuzhg: :) My design is 

1. Visit() method is used to verify the `itself` properities
2. Next() method is used to verify the `Children` element.

So, I think Extension belongs to the `Children` scope. What do you think?

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
xuzhg: Resolved.

src/Microsoft.OpenApi/Validations/Rules/OpenApiExternalDocsRules.cs
xuzhg: Thx.

---
In reply to: [157651258](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157651258) [](ancestors = 157651258)

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
xuzhg: Sure!

---
In reply to: [157663995](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157663995) [](ancestors = 157663995)

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
xuzhg: :) My design is 
1. Visit() method is used to verify the itself properities
2. Next() method is used to verify the Children element.
So, I think Extension belongs to the Children scope. What do you think?

---
In reply to: [157650373](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157650373) [](ancestors = 157650373)

src/Microsoft.OpenApi/Validations/Visitors/PathItemVisitor.cs
xuzhg: @Perth, 

Yes. However, it's a protected method. It's only be called from "Visit(...)" method.  So, I think it's only to add such Debug assert. 

And, yes. Maybe I should add the null check at the "Visit(...)" method.


---
In reply to: [157651170](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157651170) [](ancestors = 157651170)

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
xuzhg: @PerthCharern 

Yes. I can use the

Assert.Equal(String.Format(SRResource.xxxxx, typeof(xxx).FullName), exception.Message);

However, it looks not good than the plain text in the test project. Because it can help us verify the error message thrown as our expected.  What do you think?

---
In reply to: [157649686](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157649686) [](ancestors = 157649686)

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
darrelmiller: I was able to incorporate your Rules classes into the OpenApiValidator like this,

```C#
   public class OpenApiValidator : OpenApiVisitorBase
    {
        readonly ValidationRuleSet _ruleSet;
        readonly ValidationContext _context;

        /// <summary>
        /// Create a vistor that will validate an OpenAPIDocument
        /// </summary>
        /// <param name="ruleSet"></param>
        public OpenApiValidator(ValidationRuleSet ruleSet = null)
        {
            _ruleSet = ruleSet ?? ValidationRuleSet.DefaultRuleSet;
            _context = new ValidationContext(_ruleSet);
        }

        public override void Visit(OpenApiDocument item) => Validate(item);
        public override void Visit(OpenApiInfo item) => Validate(item);
        public override void Visit(OpenApiContact item) => Validate(item);
        public override void Visit(OpenApiResponse item) => Validate(item);


        public IEnumerable<ValidationError> Errors => _context.Errors;

        private void Validate<T>(T item)
        {
            if (item == null) return;  // Required fields should be checked by higher level objects
            var rules = _ruleSet.Where(r => r.ElementType == typeof(T));
            foreach (var rule in rules)
            {
                rule.Evaluate(_context, item);
            }
        }
    }
}
```
Obviously this doesn't have the Visit methods for all types yet, and some types may be more complex than just the simple expression.  Also, the walker class would need to build the property path.  
However, using this approach, we need one class for walking the DOM (we could create other walkers if other visitors need to traverse in a different way) and one class for dispatching the Visit methods to the rules.  This makes implementing other types of Visitors easier by only requiring one new class.  
My understanding is that this approach would remove the need for : 
- the Visitor class for each element
- the OpenApiVisitorSet
- the ValidationContextExtensions

I'm still not quite sure how the ValidationMap and ValidateCollection extensions fit into this approach.

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
xuzhg: I tried, But i can't.

---
In reply to: [157650480](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157650480) [](ancestors = 157650480)

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
xuzhg: It's ok. Please share your thoughts and codes. We can discuss to select one or just merge together. 

---
In reply to: [157841020](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157841020) [](ancestors = 157841020)

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
xuzhg: I just want to test multiple retrieving the property "Visitors" can work as expected.

---
In reply to: [157649800](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157649800) [](ancestors = 157649800)

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
xuzhg: Added.

---
In reply to: [157649917](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157649917) [](ancestors = 157649917)

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
xuzhg: Yes. Basically, that's convention. It's ok if we follow up this convention. 

---
In reply to: [157650606](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157650606) [](ancestors = 157650606,157650194)

test/Microsoft.OpenApi.Tests/Validations/OpenApiResponseValidationTests.cs
xuzhg: Good catch. I added such here.

Such test cases is helping us to understand the design. I'd like to make sure we can finalize the design first.

---
In reply to: [157651079](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157651079) [](ancestors = 157651079)

test/Microsoft.OpenApi.Tests/Validations/OpenApiInfoValidationTests.cs
xuzhg: Same as others:

Yes. I can use the

Assert.Equal(String.Format(SRResource.xxxxx, typeof(xxx).FullName), exception.Message);

However, it looks not good than the plain text in the test project. Because it can help us verify the error message thrown as expected.  What do you think?

---
In reply to: [157650687](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157650687) [](ancestors = 157650687)

xuzhg:
For the visitor, Yes. I have the interface IVisitor. That's enough.

---
In reply to: [352597979](https://github.com/Microsoft/OpenAPI.NET/pull/152#issuecomment-352597979) [](ancestors = 352597979,352585589)

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
xuzhg: Changed the design to use reflect. 

---
In reply to: [157320436](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157320436) [](ancestors = 157320436,157320269)

xuzhg: Changed.

---
In reply to: [157326026](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157326026) [](ancestors = 157326026,157321750)

src/Microsoft.OpenApi/Validations/ValidationContext.cs
xuzhg: It's ok. And it's VS recommendation.

---
In reply to: [157626597](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157626597) [](ancestors = 157626597,157325008)

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
xuzhg: Changed.

---
In reply to: [157327739](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157327739) [](ancestors = 157327739)

src/Microsoft.OpenApi/Validations/Visitors/OpenApiVisitorSet.cs
xuzhg: Ok.

---
In reply to: [157634816](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157634816) [](ancestors = 157634816,157624690)

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
PerthCharern: I think we should go with that instead of duplicating string literals. We are essentially checking that the exception thrown has the "right string", not that it is spelled correctly, etc.

I would say we get better verification from referring to the constant string in SRResource.

---
In reply to: [157832135](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157832135) [](ancestors = 157832135)

src/Microsoft.OpenApi/Validations/Visitors/PathItemVisitor.cs
PerthCharern: protected class is practically a public class (since our customer can technically just create a class that inherits the PathItemVisitor and calls the method).

I think we should do the null check here.

---
In reply to: [157838274](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157838274) [](ancestors = 157838274,157651170)

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
PerthCharern: Why does Extension fall into the Children scope? Isn't it one of the properties?

---
In reply to: [157838009](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157838009) [](ancestors = 157838009,157650373)

src/Microsoft.OpenApi/Validations/Visitors/VisitorBase.cs
```diff
@@ -0,0 +1,52 @@
...
+    internal abstract class VisitorBase<T> : IVisitor where T : IOpenApiElement
+    {
+        /// <summary>
+        /// Visit the element itself.
+        /// </summary>
+        /// <param name="context">The validation context.</param>
+        /// <param name="item">The element.</param>
+        public void Visit(ValidationContext context, object item)
+        {
+            Debug.Assert(item is T, "item should be " + typeof(T));
```
PerthCharern: >Debug.Assert(item is T, "item should be " + typeof(T)); [](start = 12, length = 55)

Remove Debug.Assert and use a normal check

test/Microsoft.OpenApi.Tests/Validations/Visitors/OpenApiVisitorSetTests.cs
xuzhg: Changed.

---
In reply to: [157882006](https://github.com/Microsoft/OpenAPI.NET/pull/152#discussion_r157882006) [](ancestors = 157882006,157832135)

xuzhg:
@PerthCharern @darrelmiller

src/Microsoft.OpenApi/Validations/ValidationContext.cs
darrelmiller: It's also VS recommendation to sort System. namespaces after Microsoft. namespaces.  There be dragons with that argument ;-)

darrelmiller: And nobody likes the idea of using Code Contracts so we actually get static analysis for these issues too?

PR #155, opened 2017-12-15
Fix V2 parameter deserializer to recognize formData instead of form

PerthCharern:
- Fix V2 parameter deserializer to recognize formData instead of form
- Basic unit tests for all types of possible parameters in V2.


test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
```diff
@@ -65,6 +65,21 @@
         <None Update="V2Tests\Samples\minimal.v3.yaml">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </None>
+        <None Update="V2Tests\Samples\OpenApiParameter\formDataParameter.yaml">
+          <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
xuzhg: Can we embed this files with the test.dll?

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiParameterTests.cs
```diff
@@ -0,0 +1,180 @@
...
+    public class OpenApiParameterTests
+    {
+        private const string SampleFolderPath = "V2Tests/Samples/OpenApiParameter";
+
+        [Fact]
+        public void ParseBodyParameterShouldSucceed()
+        {
+            using (var stream = File.OpenRead(Path.Combine(SampleFolderPath, "bodyParameter.yaml")))
+            {
+                var yamlStream = new YamlStream();
```
xuzhg: // Arrange

xuzhg: Same as others

test/Microsoft.OpenApi.Readers.Tests/V2Tests/Samples/OpenApiParameter/headerParameter.yaml
```diff
@@ -0,0 +1,10 @@
+﻿# https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject
```
xuzhg: Do we include this line?

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiParameterTests.cs
```diff
@@ -0,0 +1,180 @@
...
+                yamlStream.Load(new StreamReader(stream));
+                var yamlNode = yamlStream.Documents.First().RootNode;
```
xuzhg: A lot of codes are duplicated. Think to add a private method to wrapper?

test/Microsoft.OpenApi.Readers.Tests/V2Tests/Samples/OpenApiParameter/headerParameter.yaml
PerthCharern: I think we should point to the source we took the YAML files from.

test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
PerthCharern: I'm not sure about the pros and cons of this. @darrelmiller also pointed this out earlier.

What's the disadvantage of not embedding a test file?

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiParameterTests.cs
PerthCharern: I'll look into this now.

test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
darrelmiller: @PerthCharern It just means you have to carry those files around to make the tests work.  If they are embedded in the DLL then the tests are self-contained.

PerthCharern: I think I understand the concerns now. At first I thought using a non-embedded resource with relative path should work, but that might not be the case since the test assembly can technically be run from anywhere. 

Let's make this change in a separate PR for all resources.

PerthCharern:
@darrelmiller @xuzhg 

I'll take care of embedding the all test files in a separate PR. Please let me know if you guys have more comments about the changes here.

PR #160, opened 2017-12-18
Support v3.0.1

### LINKED ISSUE #159: Updates needed to support v3.0.1 (and future patches)
1. I did a diff from 3.0.0 to 3.0.1 and didn't see changes to the model itself, so there shouldn't be any major deserialization/serialization needed in this library.
2. Recognize 3.0.1 in the openapi field in the OpenAPI Object. It can still use all the existing V3 deserialization logic.
3. Create new enum value OpenApi3_0_1. (I'm going back and forth about this. This last number is simply a patch to the spec. Should we support enum for every single patch explicitly, or should we always try to output the latest patch for a given Major.Minor)

PerthCharern:
- Output as V3.0.1
- Recognize any V3.0.X when deserializing

Fix #159

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
@@ -11,7 +11,7 @@ using Microsoft.OpenApi.Writers;
 namespace Microsoft.OpenApi.Models
 {
     /// <summary>
-    /// Describes an Open API Document. See: https://swagger.io/specification
+    /// Describes an OpenAPI object (OpenAPI document). See: https://swagger.io/specification
     /// </summary>
     public class OpenApiDocument : IOpenApiSerializable, IOpenApiExtensible
     {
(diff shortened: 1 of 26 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -69,7 +69,7 @@ public void SerializeAsV3(IOpenApiWriter writer)
             writer.WriteStartObject();
 
             // openapi
-            writer.WriteProperty(OpenApiConstants.OpenApi, "3.0.0");
```
xuzhg: So, it's changed to 3.0.1?

src/Microsoft.OpenApi/OpenApiSpecVersion.cs
```diff
@@ -4,18 +4,18 @@
...
         /// </summary>
         OpenApi2_0,
 
         /// <summary>
-        /// Open Api v3.0.0
-        /// </summary>
-        OpenApi3_0_0
+        /// Represents all patches of OpenAPI V3.0 spec (e.g. 3.0.0, 3.0.1)
+        /// </summary> 
+        OpenApi3_0
```
xuzhg: why can not we add new one?

darrelmiller: A patch should not represent a functional change to the spec.  We are expecting to put out patch releases to the spec monthly.  We don't want to update this enum on a monthly basis.

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
PerthCharern: Yes, we do have v3.0.1 now.

Issue #156, opened 2017-12-15
Embed the test YAML/JSON files into the DLL


PerthCharern:
Stems out from #115

xuzhg:
@PerthCharern @darrelmiller Please review #164. Thanks.

xuzhg:
Merged at https://github.com/Microsoft/OpenAPI.NET/pull/164. Thanks!

PR #164, opened 2017-12-21
Embedded the test files into project and make test self-contain

xuzhg:
See #156 

Embedded the test files into the test project and make the test self-contain.

### DIFF
--- a/test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
+++ b/test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
@@ -13,7 +13,161 @@
         <AssemblyOriginatorKeyFile>..\..\src\Microsoft.OpenApi.snk</AssemblyOriginatorKeyFile>
     </PropertyGroup>
     <ItemGroup>
+        <None Remove="ReferenceService\Samples\multipleReferences.v2.yaml" />
+        <None Remove="V2Tests\Samples\basic.v2.yaml" />
+        <None Remove="V2Tests\Samples\basic.v3.yaml" />
+        <None Remove="V2Tests\Samples\minimal.v2.yaml" />
+        <None Remove="V2Tests\Samples\minimal.v3.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiParameter\bodyParameter.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiParameter\formDataParameter.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiParameter\headerParameter.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiParameter\pathParameter.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiParameter\queryParameter.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiSecurityScheme\apiKeySecurityScheme.yaml" />
         <None Remove="V2Tests\Samples\OpenApiSecurityScheme\httpSecurityScheme.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiSecurityScheme\oauth2AccessCodeSecurityScheme.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiSecurityScheme\oauth2ApplicationSecurityScheme.yaml" />
+        <None Remove="V2Tests\Samples\OpenApiSecurityScheme\oauth2PasswordSecurityScheme.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiCallback\basicCallback.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiDiscriminator\basicDiscriminator.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiDocument\basicDocumentWithMultipleServers.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiDocument\brokenMinimalDocument.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiDocument\minimalDocument.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiDocument\petStore.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiDocument\petStoreExpanded.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiDocument\petStoreWithTagAndSecurity.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiEncoding\advancedEncoding.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiEncoding\basicEncoding.yaml" />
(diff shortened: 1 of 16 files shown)


test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
```diff
@@ -13,7 +13,161 @@
...
+        <None Remove="V3Tests\Samples\OpenApiSchema\dictionarySchema.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiSchema\primitiveSchema.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiSchema\simpleSchema.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiSecurityScheme\apiKeySecurityScheme.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiSecurityScheme\bearerSecurityScheme.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiSecurityScheme\httpSecurityScheme.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiSecurityScheme\oauth2SecurityScheme.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiSecurityScheme\openIdConnectSecurityScheme.yaml" />
+        <None Remove="V3Tests\Samples\OpenApiXml\basicXml.yaml" />
+    </ItemGroup>
```
PerthCharern: You can remove these lines

test/Microsoft.OpenApi.Readers.Tests/ReferenceService/TryLoadReferenceV2Tests.cs
```diff
@@ -25,7 +25,7 @@ public void LoadSchemaReference()
             var diagnostic = new OpenApiDiagnostic();
             RootNode rootNode;
 
-            using (var stream = File.OpenRead(Path.Combine(SampleFolderPath, "multipleReferences.v2.yaml")))
+            using (var stream = Resources.GetStream(SampleFolderPath + "multipleReferences.v2.yaml"))
```
PerthCharern: >(SampleFolderPath + "multipleReferences.v2.yaml")) [](start = 51, length = 50)

Path.Combine is usually a preferred way to combine paths so that you don't need to worry about the trailing/beginning slashes. 

Not that big of a deal though.

test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
xuzhg: Oops, they look necessary to let the compiler skip those files. 
Can we double check?

test/Microsoft.OpenApi.Readers.Tests/Resources.cs
```diff
@@ -0,0 +1,58 @@
...
+            const string resourcesFolderName = "Resources";
+            const string pathSeparator = ".";
+
+            if (fileName.Contains("/"))
+            {
+                string temp = fileName.Replace('/', '.');
+                return projectDefaultNamespace + pathSeparator + temp;
+            }
+
+            return projectDefaultNamespace + pathSeparator + resourcesFolderName + pathSeparator + fileName;
```
PerthCharern: >resourcesFolderName + pathSeparator [](start = 61, length = 35)

Do we need this? Why can't we keep this consistent with the case in the "if" above?

test/Microsoft.OpenApi.Readers.Tests/Resources.cs
```diff
@@ -0,0 +1,58 @@
...
+                string message = Error.Format("The embedded resource '{0}' was not found.", path);
+                throw new FileNotFoundException(message, path);
+            }
+
+            return stream;
+        }
+
+        private static string GetPath(string fileName)
+        {
+            const string projectDefaultNamespace = "Microsoft.OpenApi.Readers.Tests";
```
PerthCharern: > const string projectDefaultNamespace = "Microsoft.OpenApi.Readers.Tests"; [](start = 11, length = 74)

Let's get this programmatically from the namespace of this class.

test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
PerthCharern: The compiler can read it. If we use None Remove, won't it disappear from the project list in VS.

---
In reply to: [159328930](https://github.com/Microsoft/OpenAPI.NET/pull/164#discussion_r159328930) [](ancestors = 159328930)

test/Microsoft.OpenApi.Readers.Tests/Resources.cs
xuzhg: Changed.

test/Microsoft.OpenApi.Readers.Tests/Resources.cs
xuzhg: Changed.

test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
xuzhg: Removed.

test/Microsoft.OpenApi.Readers.Tests/ReferenceService/TryLoadReferenceV2Tests.cs
xuzhg: Changed.

test/Microsoft.OpenApi.Readers.Tests/Resources.cs
```diff
@@ -0,0 +1,57 @@
...
+            const string pathSeparator = ".";
+            string projectDefaultNamespace = typeof(Resources).Namespace;
+
+            if (fileName.Contains("/"))
+            {
+                string temp = fileName.Replace('/', '.');
+                return projectDefaultNamespace + pathSeparator + temp;
+            }
+
+            return projectDefaultNamespace + pathSeparator + fileName;
```
PerthCharern: You can replace this with

```
return projectDefaultNamespace + pathSeparator + fileName.Replace('/', '.');
```

test/Microsoft.OpenApi.Readers.Tests/Resources.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.IO;
+
+namespace Microsoft.OpenApi.Readers.Tests
+{
+    internal static class Resources
+    {
+        /// <summary>
+        /// Get the file contents.
+        /// </summary>
+        /// <param name="fileName">The file name with relative path. "/V3Tests/Samples/OpenApiCallback/..</param>
```
PerthCharern: >" [](start = 69, length = 1)

nit, Either remove this quotation mark or add one at the end.

Same thing with the comment below.

Issue #166, opened 2018-01-03
Evaluate whether current implementation for RuntimeExpression is sufficient


PerthCharern:

xuzhg:
https://swagger.io/specification/#runtimeExpression

PerthCharern:
Let's create unit tests for this.

xuzhg:
Merged the PR #168. Thanks.

PR #168, opened 2018-01-03
add the unit test for the runtime expression

xuzhg:
add the unit test for the runtime expression.

See #166.


src/Microsoft.OpenApi/Expressions/RuntimeExpression.cs
```diff
@@ -42,19 +42,19 @@ public static RuntimeExpression Build(string expression)
             // $url
             if (expression == UrlExpression.Url)
             {
-                return new UrlExpression();
+                return UrlExpression.Instance;
```
PerthCharern: >UrlExpression.Instance; [](start = 23, length = 23)

I can see the benefits of the singleton pattern here, but I think it can make things a bit confusing. (Using .Instance is less natural then new Class())

src/Microsoft.OpenApi/Expressions/RuntimeExpression.cs
```diff
@@ -99,5 +99,11 @@ public bool Equals(RuntimeExpression obj)
         {
             return obj != null && obj.Expression == Expression;
         }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            return Expression;
+        }
```
PerthCharern: Not sure when we need this?

test/Microsoft.OpenApi.Tests/Expressions/PathExpressionTests.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.OpenApi.Expressions;
+using Xunit;
+
+namespace Microsoft.OpenApi.Tests.Writers
+{
+    [Collection("DefaultSettings")]
+    public class QueryExpressionTests
```
PerthCharern: >QueryExpressionTests [](start = 17, length = 20)

QueryExpressionTests and PathExpressionTests file names are swapped

src/Microsoft.OpenApi/Expressions/BodyExpression.cs
```diff
@@ -54,5 +54,16 @@ public override string Expression
...
+        /// <summary>
+        /// Gets the fragment string.
+        /// </summary>
+        public string Fragment
+        {
+            get
+            {
+                return Value;
+            }
+        }
```
PerthCharern: :smile:

xuzhg: What does the laugh here mean? :)

test/Microsoft.OpenApi.Tests/Expressions/PathExpressionTests.cs
xuzhg: Oh, yes. Swapped.

src/Microsoft.OpenApi/Expressions/RuntimeExpression.cs
xuzhg: Yes. I think so. It can be used to anywhere if you want the string.

src/Microsoft.OpenApi/Expressions/BodyExpression.cs
PerthCharern: It's a smile. :) I like it that you are adding these properties. They match the runtime expression spec terminology.

src/Microsoft.OpenApi/Expressions/RuntimeExpression.cs
xuzhg: I don't think there's any benefits to use new Class().  I mean for "$method, $url, $statusCode", they are no other information. It's be better to use singleton.

src/Microsoft.OpenApi/Expressions/RuntimeExpression.cs
PerthCharern: I guess it's okay. I was a bit worried about how people interpret what ToString() should output. But I don't think there's other choice on how to output runtime expression as a string.

darrelmiller: Feels natural to me.

src/Microsoft.OpenApi/Expressions/RuntimeExpression.cs
PerthCharern: That's fine. Let's roll with this.

PR #175, opened 2018-01-11
Fix #172: Adding spec version to OpenApiDiagnostic of Reader

### LINKED ISSUE #172: No simple way to convert formats while keeping OpenAPI spec version the same
I'd like to simply convert between `OpenApiFormat` while keeping `OpenApiSpecVersion` the same across the conversion, however, there is no way of doing this that avoids me having to parse the `OpenApiSpecVersion` from the string payload myself.

For example, below is some basic code to demonstrate the scenario and issue. In this example, we'd be converting YAML string -> OpenApirDocument model -> JSON string.

    string documentYaml = "some YAML OpenApi document payload";

    var openApiStringReader = new OpenApiStringReader();
    OpenApiDocument openApiDocument = openApiStringReader.Read( documentYaml, out 
    OpenApiDiagnostic openApiDiagnostic );

    string documentJson = openApiDocument.SerializeAsJson(???);

What is the recommended way of getting to know the `OpenApiSpecVersion` to pass to `SerializeAsJson` (respresented by ??? above) in order to keep the version consistent.

scott-lin:
Introducing a property to track the specification version read, so consumers of the Reader have this metadata and do not have to parse it themselves. See https://github.com/Microsoft/OpenAPI.NET/issues/172


src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -45,18 +45,21 @@ internal OpenApiDocument Parse(YamlDocument yamlDocument, OpenApiDiagnostic diag
...
                 doc = this.VersionService.LoadDocument(this.RootNode);
+                diagnostic.SpecificationVersion = OpenApiSpecVersion.OpenApi2_0;
             }
             else if (inputVersion.StartsWith("3.0."))
             {
                 this.VersionService = new OpenApiV3VersionService();
                 doc = this.VersionService.LoadDocument(this.RootNode);
+                diagnostic.SpecificationVersion = OpenApiSpecVersion.OpenApi3_0;
             }
             else
```
xuzhg: The `else if (...) { }` and `else { ...}` has the same codes? why cannot we merge them together? 

@PerthCharern @darrelmiller

PerthCharern:
I'd rather put these two tests in the same folder:

test/Microsoft.OpenApi.Readers.Tests/OpenApiReaderTests

The ComparisonTests in V2Tests should probably be moved out to this folder as well.

src/Microsoft.OpenApi.Readers/ParsingContext.cs
PerthCharern: The "else" is for a default case. It is, but doesn't have to be, the same as the "else if". This makes sure we list out the cases for 2.0 and 3.0 explicitly.

If you are worried about code duplication, we can refactor the guts out.

---
In reply to: [160842325](https://github.com/Microsoft/OpenAPI.NET/pull/175#discussion_r160842325) [](ancestors = 160842325)

test/Microsoft.OpenApi.Readers.Tests/OpenApiReaderTests/OpenApiDiagnosticTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        public void DetectedSpecificationVersionShouldBeV3_0()
+        {
+            using ( var stream = Resources.GetStream("V3Tests/Samples/OpenApiDocument/minimalDocument.yaml"))
+            {
+                new OpenApiStreamReader().Read( stream, out var diagnostic );
+
+                diagnostic.Should().NotBeNull();
+                diagnostic.SpecificationVersion.Should().Be( OpenApiSpecVersion.OpenApi3_0 );
+            }
+        }
```
PerthCharern: nit, spacing

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiCallbackTests.cs
```diff
@@ -88,7 +88,8 @@ public void ParseAdvancedCallbackWithReferenceShouldSucceed()
 
                 var callback = subscribeOperation.Callbacks["simpleHook"];
 
-                context.ShouldBeEquivalentTo(new OpenApiDiagnostic());
+                context.ShouldBeEquivalentTo(
```
PerthCharern: >context [](start = 16, length = 7)

If you have some time, would you mind fixing these context -> diagnostic? If not, it's okay. We can do a big cleanup later.

src/Microsoft.OpenApi.Readers/ParsingContext.cs
scott-lin: Won't fix with the scope of this change.

---
In reply to: [160843959](https://github.com/Microsoft/OpenAPI.NET/pull/175#discussion_r160843959) [](ancestors = 160843959,160842325)

PR #176, opened 2018-01-13
Validation for custom extensions

darrelmiller:
This PR incorporates the other strongly typed extension work and allows defining validation rules for those extension types.


src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -15,7 +18,17 @@ namespace Microsoft.OpenApi.Readers
     /// </summary>
     public class OpenApiStreamReader : IOpenApiReader<Stream, OpenApiDiagnostic>
     {
+        private OpenApiReaderSettings _settings;
 
+        /// <summary>
+        ///
```
xuzhg: fill?

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -15,7 +18,17 @@ namespace Microsoft.OpenApi.Readers
     /// </summary>
     public class OpenApiStreamReader : IOpenApiReader<Stream, OpenApiDiagnostic>
     {
+        private OpenApiReaderSettings _settings;
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="settings"></param>
+        public OpenApiStreamReader(OpenApiReaderSettings settings = null)
```
xuzhg: it's better to separate two constructors?

src/Microsoft.OpenApi/Any/OpenApiObject.cs
```diff
@@ -14,5 +15,23 @@ public class OpenApiObject : Dictionary<string, IOpenApiAny>, IOpenApiAny
         /// Type of <see cref="IOpenApiAny"/>.
         /// </summary>
         public AnyType AnyType { get; } = AnyType.Object;
+
+        /// <summary>
+        ///
```
xuzhg: fill?

src/Microsoft.OpenApi/Any/OpenApiArray.cs
```diff
@@ -14,5 +15,22 @@ public class OpenApiArray : List<IOpenApiAny>, IOpenApiAny
         /// The type of <see cref="IOpenApiAny"/>
         /// </summary>
         public AnyType AnyType { get; } = AnyType.Array;
+
+        /// <summary>
+        /// Write out contents of OpenApiArray to passed writer
+        /// </summary>
+        /// <param name="writer">Instance of JSON or YAML writer.</param>
+        public void Write(IOpenApiWriter writer)
```
xuzhg: public? not internal?

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
```diff
@@ -1,34 +1,61 @@
...
 {
     /// <summary>
     /// Class containing dispatchers to execute validation rules on for Open API document.
     /// </summary>
-    public class OpenApiValidator : OpenApiVisitorBase
+    public class OpenApiValidator : OpenApiVisitorBase, IValidationContext 
     {
-        readonly ValidationRuleSet _ruleSet;
-        readonly ValidationContext _context;
+
```
xuzhg: remove

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
darrelmiller: I don't think so.  The optional parameter conveys that settings are optional and having one constructor prevents someone from adding code to one ctor that needs to be called from both.

---
In reply to: [161353699](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r161353699) [](ancestors = 161353699)

src/Microsoft.OpenApi/Any/OpenApiArray.cs
darrelmiller: Yes it needs to be public because it is part of an interface that will be implemented by any custom extension implementation.

---
In reply to: [161353844](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r161353844) [](ancestors = 161353844)

PerthCharern:
> using Microsoft.OpenApi.Writers;

nit, add copyright

---
Refers to: src/Microsoft.OpenApi/Interfaces/IOpenApiExtension.cs:1 in 7e1f151. [](commit_id = 7e1f151687792ba9586d2b93ff7f7d926a22c5c1, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
```diff
@@ -12,6 +14,37 @@ namespace Microsoft.OpenApi.Services
     /// </summary>
     public abstract class OpenApiVisitorBase
     {
+        private readonly Stack<string> _path = new Stack<string>();
+       
+        /// <summary>
+        /// Allow Rule to indicate validation error occured a deeper context level.
```
PerthCharern: >  [](start = 59, length = 1)

preposition missing

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -78,11 +87,15 @@ internal void Walk(OpenApiComponents components)
         /// <param name="paths"></param>
         internal void Walk(OpenApiPaths paths)
         {
+            _visitor.Enter(OpenApiConstants.Paths);
             _visitor.Visit(paths);
-            foreach (var pathItem in paths.Values)
+            foreach (var pathItem in paths)
             {
-                Walk(pathItem);
+                _visitor.Enter(pathItem.Key.Replace("/","~1"));
```
PerthCharern: >~1" [](start = 57, length = 3)

nit, could you put comment in the code to explain that "~1" is an escape character for "/" in JSON pointer?

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -101,6 +115,7 @@ internal void Walk(IList<OpenApiServer> servers)
                     Walk(server);
                 }
```
PerthCharern: We should add the array index to get the correct json pointer here.

                for ( int i =0; i<servers.Count(); i++ )
                {
                    _visitor.Enter(i);
                    Walk(servers[i]);
                     _visitor.Exit();
                }

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -46,12 +48,15 @@ public void Walk(OpenApiDocument doc)
         /// <param name="tags"></param>
         internal void Walk(IList<OpenApiTag> tags)
         {
+            _visitor.Enter(OpenApiConstants.Tags);
             _visitor.Visit(tags);
 
             foreach (var tag in tags)
             {
                 Walk(tag);
             }
```
PerthCharern: We should add the array index to get the correct json pointer here.

                for ( int i =0; i<tags.Count(); i++ )
                {
                    _visitor.Enter(i);
                    Walk(tags[i]);
                     _visitor.Exit();
                }

Also, this raises another point that the Enter call should be the responsibility of the caller. In this case, the Walk(tags[i]) should NOT need to call any enter for the Tag object itself (since the array index is used instead).

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -46,12 +48,15 @@ public void Walk(OpenApiDocument doc)
         /// <param name="tags"></param>
         internal void Walk(IList<OpenApiTag> tags)
         {
+            _visitor.Enter(OpenApiConstants.Tags);
```
PerthCharern: >    _visitor.Enter(OpenApiConstants.Tags); [](start = 8, length = 42)

Enter/Exit should be responsibility of the caller (e.g. in this case, it should go in the Walk(doc) ).

This will ensure that we don't do duplicate Enter calls for an object in an array (like the tags array).

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -224,11 +274,13 @@ internal void Walk(IList<OpenApiParameter> parameters)
         {
             if (parameters != null)
             {
+                _visitor.Enter(OpenApiConstants.Parameters);
                 _visitor.Visit(parameters);
                 foreach (var parameter in parameters)
                 {
                     Walk(parameter);
                 }
```
PerthCharern: need indexing - see my other comments.

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
PerthCharern: You already kinda handled this for Tag (but not for Server), but moving the Error/Exits to the caller altogether would make it more consistent. 



---
In reply to: [162477655](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r162477655) [](ancestors = 162477655)

PerthCharern:
> using System;

nit, copyright, and how is this file used?

---
Refers to: src/Microsoft.OpenApi/Services/VisitorContext.cs:1 in 7e1f151. [](commit_id = 7e1f151687792ba9586d2b93ff7f7d926a22c5c1, deletion_comment = False)

src/Microsoft.OpenApi/Validations/ValidationContext.cs
```diff
@@ -3,76 +3,39 @@
...
+        /// Exit from path context elevel.  Enter and Exit calls should be matched.
         /// </summary>
-        public IEnumerable<ValidationError> Errors
-        {
-            get
-            {
-                return _errors;
-            }
-        }
+        void Exit();
```
PerthCharern: Why are these needed in the context?

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
```diff
@@ -1,34 +1,58 @@
...
 using Microsoft.OpenApi.Services;
-using Microsoft.OpenApi.Validations;
 
 namespace Microsoft.OpenApi.Validations
 {
     /// <summary>
     /// Class containing dispatchers to execute validation rules on for Open API document.
     /// </summary>
-    public class OpenApiValidator : OpenApiVisitorBase
+    public class OpenApiValidator : OpenApiVisitorBase, IValidationContext
```
PerthCharern: >, IValidationContext  [](start = 54, length = 21)

It's weird that the Validator is implementing the Context interface. Do we need the context interface at all?

PerthCharern:
> }

Can we add tests with data that is "incorrect" in an array? (something like a tag in tags or server in servers), so that we can check that our Jsonpointer looks correct in that case

---
Refers to: test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs:56 in 7e1f151. [](commit_id = 7e1f151687792ba9586d2b93ff7f7d926a22c5c1, deletion_comment = False)

src/Microsoft.OpenApi/Validations/ValidationContext.cs
darrelmiller: Because often the rules validate additional depth.  Rules add segments for properties in objects being validated.  Also, for custom validation rules may validate several levels of object depth.

---
In reply to: [162479370](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r162479370) [](ancestors = 162479370)

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
darrelmiller: We could pass the entire validator object into rules, however this exposes a bunch of properties that Rules don't need access to.  The way it was implemented before was the state of the validator was refactored out into an aggregated object and the state object was passed into the Rules.  Rather than have the validator be made up of two objects, I just use the IValidationContext as a filter to limit what parts of the Validator the rules are allowed to access.

---
In reply to: [162479509](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r162479509) [](ancestors = 162479509)

darrelmiller:
> using System;

I deleted it.

---
In reply to: [358791618](https://github.com/Microsoft/OpenAPI.NET/pull/176#issuecomment-358791618) [](ancestors = 358791618)

---
Refers to: src/Microsoft.OpenApi/Services/VisitorContext.cs:1 in 7e1f151. [](commit_id = 7e1f151687792ba9586d2b93ff7f7d926a22c5c1, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
```diff
@@ -1,17 +1,50 @@
...
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Models;
 
 namespace Microsoft.OpenApi.Services
 {
     /// <summary>
-    /// Open API visitor base providing base validation logic for each <see cref="IOpenApiElement"/>
+    /// Open API visitor base provides common logic for concrete vistors
```
PerthCharern: >vistors [](start = 65, length = 7)

visitors

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -222,14 +278,19 @@ internal void Walk(OpenApiOperation operation)
         /// <param name="parameters"></param>
         internal void Walk(IList<OpenApiParameter> parameters)
         {
+            _visitor.Enter(OpenApiConstants.Parameters);
```
PerthCharern: >_visitor.Enter(OpenApiConstants.Parameters); [](start = 12, length = 44)

I still think that Enter and Exit should be the responsibility of the caller. Here we are assuming that this property will always be called a "parameters", but this can actually be called anything depending on where it is in the doc, and only the caller has that knowledge. This is similar to the case where Walk(tag) doesn't have the knowledge of what the tag should be called, but the caller knows that each tag can be called by an index.

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
PerthCharern: I opened another comment to discuss this.

---
In reply to: [162478481](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r162478481) [](ancestors = 162478481,162477655)

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
PerthCharern: still missing :)

---
In reply to: [162469699](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r162469699) [](ancestors = 162469699)

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
PerthCharern: Ah that makes sense. Thanks for the explanation. 

I think the design itself is fine, but the nomenclature still throws me off a bit. I would expect the interface to be called something along the line of "IValidationContextHolder".

That name is mouthful though... so I'm also fine with the current state.

---
In reply to: [163074706](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r163074706) [](ancestors = 163074706,162479509)

src/Microsoft.OpenApi/Validations/IValidationContext.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Register an error with the validation context.
+        /// </summary>
+        /// <param name="error">Error to register.</param>
+        void AddError(ValidationError error);
+
+        /// <summary>
+        /// Allow Rule to indicate validation error occured a deeper context level.  
+        /// </summary>
+        /// <param name="segment"></param>
```
PerthCharern: >/// <param name="segment"></param> [](start = 8, length = 34)

nit, Either remove or fill

test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs
```diff
@@ -48,9 +48,42 @@ public void ResponseMustHaveADescription()
...
+            validator.Errors.ShouldBeEquivalentTo(
+                    new List<ValidationError>
+                    {
+                        new ValidationError(ErrorReason.Required, "#/servers/1/url",
+                            String.Format(SRResource.Validation_FieldIsRequired, "url", "server"))
+        });
+        }
+
     }
 }
```
PerthCharern: Any value in having a test with custom extensions?

PerthCharern:
🕐

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
darrelmiller: I am struggling to think of anywhere in the spec where a list of <something> is named more than one thing.  However, it might happen one day.  My gut was telling me that putting the Enter/Exit inside the walk method was more cohesive, but I will look at it again.

test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs
darrelmiller: I put the validation test into the existing custom extension reading test https://github.com/Microsoft/OpenAPI.NET/pull/176/files#diff-f884e65f4ced1eff8c8cfc2496d317d1R33  I'll split out the tests to put stuff in the right place.

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
PerthCharern: I don't think there's anything like that in the spec today. There are places where a non-list is named differently (schema, for example). 

It seems more readable to me if every Walk follows the same pattern, but if you'd rather follow two conventions: one for list and one for non-list, that's also okay to me.

---
In reply to: [163352360](https://github.com/Microsoft/OpenAPI.NET/pull/176#discussion_r163352360) [](ancestors = 163352360)

test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs
darrelmiller: Done.

src/Microsoft.OpenApi/Validations/IValidationContext.cs
darrelmiller: Done

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
darrelmiller: Now it's fixed :-)

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
darrelmiller: @PerthCharern   I can't bring myself to lift all the Enter/Exit calls out of the Walk methods that they are wrapping.  Seeing them at the start and the end just makes me feel safe.  However, I agree the name of the context should be defined by the caller.  Therefore I'm going to add a parameter to all walkers so that the caller can define the name of the context segment, but the callee actually has the Enter/Exit calls.  This should allow ever Walk method to be consistent whether it is a property or array item.  Hopefully that will make both of us happy :-)

darrelmiller: I did the changes and I really don't like where it ended up.  I'm going to try something different.

darrelmiller:
Ok. Not sure where the original thread is where we were talking about the Walker implementation, but I think it is much better now.

darrelmiller:
@PerthCharern  The walker will now handle cycles in the DOM.  Now I'm going to go back to the reader to make sure it handles reading circular $refs.

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
```diff
@@ -39,19 +50,19 @@ public OpenApiValidator(ValidationRuleSet ruleSet = null)
         /// Execute validation rules against an <see cref="OpenApiInfo"/>
         /// </summary>
         /// <param name="item">The object to be validated</param>
-        public override void Visit(OpenApiInfo item) => Validate(item);
+        public override void Visit(OpenApiInfo item) => Validate("info",item);
```
PerthCharern: There are constants for the property names ("info", "contact", etc.) in OpenApiConstants class somewhere.

PerthCharern: Oops. Sorry. I got confused by the GitHub PR UI. This is outdated, so please ignore.

test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs
```diff
@@ -85,5 +87,63 @@ public void ServersShouldBeReferencedByIndex()
...
+            validator.Errors.ShouldBeEquivalentTo(
+                   new List<ValidationError>
+                   {
+                       new ValidationError(ErrorReason.Format, "#/info/x-foo", "Don't say hey")
+                   });
+        }
+
+    }
+
+    public class FooExtension : IOpenApiExtension, IOpenApiElement
```
PerthCharern: nit, suggest making this class private to avoid unintended effects elsewhere.

test/Microsoft.OpenApi.Tests/Walkers/WalkerLocationTests.cs
```diff
@@ -0,0 +1,229 @@
+﻿using System.Collections.Generic;
```
PerthCharern: nit, copyright

test/Microsoft.OpenApi.Tests/Walkers/WalkerLocationTests.cs
```diff
@@ -0,0 +1,229 @@
...
+                "#/components/loopy/properties/name",
+                "#/components/loopy/properties/name/externalDocs",
+                "#/components/loopy/externalDocs",
+                "#/externalDocs",
+                "#/tags"
+            });
+        }
+    }
+
+    public class LocatorVisitor : OpenApiVisitorBase
```
PerthCharern: This is actually a pretty nice feature. What do you think about moving it to the source, instead of leaving it in tests?

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -33,132 +35,186 @@ public void Walk(OpenApiDocument doc)
...
+
+            // Visit Paths
+            if (paths != null)
             {
-                _visitor.Enter(pathItem.Key.Replace("/","~1"));  // JSON Pointer uses ~1 as an escape character for /
-                Walk(pathItem.Value);
-                _visitor.Exit();
+                foreach (var pathItem in paths)
+                {
+                    InContext(pathItem.Key.Replace("/", "~1"), () => Walk(pathItem.Value));// JSON Pointer uses ~1 as an escape character for /
```
PerthCharern: This "Replace" should technically apply to anything in the JSON pointer (not just pathItem names). Can we move it to the InContext method?

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -266,160 +346,132 @@ internal void Walk(OpenApiOperation operation)
...
         {
-            foreach (var item in encoding.Values)
+            _visitor.Visit(encodings);
+
+            if (encodings != null)
             {
-                _visitor.Visit(item);
+                foreach (var item in encodings.Values)
+                {
+                    _visitor.Visit(item);
```
PerthCharern: Should this be InContext?

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -603,12 +674,22 @@ internal void Walk(IOpenApiElement element)
...
                 case IList<OpenApiTag> e: Walk(e); break;
-                case OpenApiXml e: Walk(e); break;
                 case IOpenApiExtensible e: Walk(e); break;
                 case IOpenApiExtension e: Walk(e); break;
-
             }
         }
 
+        /// <summary>
+        /// Adds a segment to the context path to enable pointing to the current locationin the document
```
PerthCharern: nit, space missing between location and in

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -603,12 +674,22 @@ internal void Walk(IOpenApiElement element)
...
+        /// </summary>
+        /// <param name="context">An identifier for the context.</param>
+        /// <param name="walk">An action that walks objects within the context.</param>
+        private void InContext(string context, Action walk)
```
PerthCharern: nit, InContext is a little strange as a method name. 

This is still a Walk method to me (just with a context name). Maybe it should be called "Walk" or "WalkInto"?

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -16,6 +16,8 @@ namespace Microsoft.OpenApi.Services
     public class OpenApiWalker
     {
         private readonly OpenApiVisitorBase _visitor;
+        private readonly Stack<OpenApiSchema> _schemaLoop = new Stack<OpenApiSchema>();
+        private readonly Stack<OpenApiPathItem> _pathItemLoop = new Stack<OpenApiPathItem>();
```
PerthCharern: Wow. I didn't realize before today that pathItem can also cause cycles (via Operation -> Callback)!

test/Microsoft.OpenApi.Tests/Walkers/WalkerLocationTests.cs
PerthCharern: This doesn't have to be done with this PR though. Maybe let's change this to private and we can decide?

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -33,132 +35,186 @@ public void Walk(OpenApiDocument doc)
...
+            {
+                return;
+            }
+
+            if (components.Schemas != null)
+            {
+                foreach (var item in components.Schemas)
+                {
+                    InContext(item.Key, () => Walk(item.Value));
+                }
```
PerthCharern: Each of these loops should be in other InContext:

```
       InContext( OpenApiConstants.Schemas,  () => foreach (var item in components.Schemas)
                {
                    InContext(item.Key, () => Walk(item.Value));
                } );
```

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -266,160 +346,132 @@ internal void Walk(OpenApiOperation operation)
...
             _visitor.Visit(content);
-            foreach (var mediaType in content)
+            if (content != null)
             {
-                _visitor.Enter(mediaType.Key);
-                Walk(mediaType.Value);
-                _visitor.Exit();
+                foreach (var mediaType in content)
+                {
+                    InContext(mediaType.Key.Replace("/", "~1"), () => Walk(mediaType.Value));
```
PerthCharern: Let's move the Replace into InContext since it's used multiple times.

test/Microsoft.OpenApi.Tests/Validations/ValidationRuleSetTests.cs
```diff
@@ -34,7 +34,7 @@ public void DefaultRuleSetPropertyReturnsTheCorrectRules()
             // Assert
             Assert.NotNull(rules);
             Assert.NotEmpty(rules);
-            Assert.Equal(13, rules.ToList().Count); // please update the number if you add new rule.
+            Assert.Equal(14, rules.ToList().Count); // please update the number if you add new rule.
```
PerthCharern: I don't think this test is needed. It doesn't really add much value, and will cause maintenance issue (like the current test failure).

test/Microsoft.OpenApi.Readers.Tests/TestCustomExtension.cs
```diff
@@ -0,0 +1,69 @@
+﻿using FluentAssertions;
```
PerthCharern: nit, copyright

darrelmiller:
I was really hoping to merge this today. But I cannot repro this failing test locally.  Even running the Xunit console runner in release mode does not repro this fail. sigh.

PerthCharern:
Maybe I misunderstood something but based on the AppVeyor TESTS tab, it looks like the failing test is the one counting number of rules. 

I also left a comment earlier suggesting removing that test since it doesn’t really add much value and requires high maintenance.

Does this solve the problem, or are you seeing other tests failing on AppVeyor as well?

darrelmiller:
@PerthCharern I could remove it and that would make the problem go away, but I am concerned that for some reason when reflecting over the available Rules, we are picking up the FooExtension test that is not supposed to be in the set of default rules.  I may remove the test temporarily to get these changes merged in, but I would really like to know why the problem is happening.

PerthCharern:
@darrelmiller My mistake. I wasn't aware that the number increase was not intentional. For some reason, I thought you added a rule and simply did not update the number.

PerthCharern:
@darrelmiller I also can't repro it locally. 

One thing that might be helpful: Make the test output something (using ITestOutputHelper) and we can see the output on AppVeyor.

Given that we can't see the issue locally, I wonder if this has something to do with the lazy initialization with isThreadSafe being true.

darrelmiller:
@PerthCharern Thanks for all the comments. I'll make the changes later this evening.  

I do think the test problem is a threading issue (although is very consistent!).  Previously isThreadSafe was false and I changed it because from the docs it appears that setting it to true adds in a lock.  The parameter isn't asking a question, it is a directive.

I might try pre-reading the ruleset in the test setup to see if that makes the problem go away.

PerthCharern:
@darrelmiller I think I know what's going on. I'll send out a PR today.

Issue #182, opened 2018-01-17
parse will throw stack overflow


partychen:
for example, i am trying to parse https://github.com/Microsoft/OpenAPI.NET.OData/blob/master/docs/oas3_0_0/graph.beta.json, and it will throw stack overflow.

darrelmiller:
Thanks for the report.  We will investigate.

partychen:
I give a detail here , if we have **circular reference** in the definition, it can be easy reproduced. So hope you can resolve this issues.

microsoft.graph.schemaExtension: 
    allOf: 
      - 
        title: "schemaExtension"
        type: "object"
        properties: 
          description: 
            type: "string"
            nullable: true
          targetTypes: 
            type: "array"
            items: 
              type: "string"
          status: 
            type: "string"
          owner: 
            type: "string"
          child:
            $ref: "#/components/schemas/microsoft.graph.schemaExtension"

xuzhg:
@PerthCharern @darrelmiller Would you please take a look?

darrelmiller:
@xuzhg Working on a fix now.

sergey-tihon:
More samples with Stack Overflow:
- https://api.apis.guru/v2/specs/amazonaws.com/ssm/2014-11-06/swagger.json
- https://api.apis.guru/v2/specs/amazonaws.com/ce/2017-10-25/swagger.json
- https://api.apis.guru/v2/specs/amazonaws.com/elasticmapreduce/2009-03-31/swagger.json
- https://api.apis.guru/v2/specs/azure.com/recoveryservicessiterecovery-service/2016-08-10/swagger.json
- https://api.apis.guru/v2/specs/azure.com/resources/2017-05-10/swagger.json
- https://api.apis.guru/v2/specs/azure.com/resources/2016-02-01/swagger.json
- https://api.apis.guru/v2/specs/azure.com/trafficmanager-trafficmanageranalytics/2017-09-01-preview/swagger.json
- https://api.apis.guru/v2/specs/azure.com/timeseriesinsights/2017-11-15/swagger.json
- https://api.apis.guru/v2/specs/azure.com/resources/2016-09-01/swagger.json
- https://api.apis.guru/v2/specs/azure.com/search/2015-08-19/swagger.json
- https://eaccountingapi-sandbox.test.vismaonline.com/swagger/docs/v2
- https://api.apis.guru/v2/specs/cybertaxonomy.eu/1.0/swagger.json
[...]

PR #185, opened 2018-01-19
Prevent overflow for self-referencing schemas 

darrelmiller:
Fix for #182 
**UPDATE**  Major update to this PR.  See below

This fix does address the problem in the parser.  However, it produces a bit of an unusual DOM and I'm not sure this is what we want.  
We detect the cycle whilst constructing the dom object, so when we discover there is a cycle we don't have direct access to the original object we are creating.  What this currently does is create a second instance of the self-referenced object and the second instance contains just an OpenApiReference back to the original object.  This means the object graph itself doesn't have a cycle in it.  This is handy so that the Walker doesn't get stuck in loops.  The reader doesn't have a problem because it only fully explodes objects that live under $/components/schemas so it doesn't even try to render the second instance of the object.

It does mean that someone could construct a graph manually with a cycle in it, that would break the walker.  Again I think the reader would survive.

I will open an issue regarding the walker failing for graphs with cycles and we should consider if the reader should actually try and create a cycle in the DOM.


PerthCharern:
I'll take a look at the code. Here's my comment based on the PR description:

Would it be better if we handle this with the reader/walker/etc. instead of at the DOM level?

The DOM having cycles should remain valid -- an object should be able to refer to itself, just like normal objects in C#. Whoever is operating on the DOM (reader/walker/etc.) should be able to stop the infinite loop once it is detected. 

In this case, the walker can memorize the node it has visited and make sure it does not visit the duplicate node. The reader can do something similar.

I might be missing nuances here, so I'll take a look at the code and I might see the limitations with the method above.

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiSchemaTests.cs
```diff
@@ -430,5 +430,24 @@ public void ParseAdvancedSchemaWithReferenceShouldSucceed()
...
+        [Fact]
+        public void ParseSelfReferencingSchemaShouldNotStackOverflow()
+        {
+            using (var stream = Resources.GetStream(Path.Combine(SampleFolderPath, "selfReferencingSchema.yaml")))
+            {
+                // Act
+                var openApiDoc = new OpenApiStreamReader().Read(stream, out var diagnostic);
+
+                // Assert
+                var components = openApiDoc.Components;
```
PerthCharern: >var components = openApiDoc.Components; [](start = 16, length = 39)

Let's also check that components is parsed correctly.

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiSchemaTests.cs
```diff
@@ -430,5 +430,24 @@ public void ParseAdvancedSchemaWithReferenceShouldSucceed()
...
+
+                diagnostic.ShouldBeEquivalentTo(
+                    new OpenApiDiagnostic() { SpecificationVersion = OpenApiSpecVersion.OpenApi3_0 });
+
+
+            }
+        }
```
PerthCharern: Location-wise I think this should go into OpenApiComponentTests (which doesn't exist yet) - also with the yaml not containing sections other than components.

PerthCharern:
One way I can think of:

Post-process the schema section in the deserialization of Components. After all the schemas are loaded, recursively delve into the schemas (into the fields in Schema object that are also a Schema object) and ensure that they reference the schemas in the Components.Schemas map. 

Essentially, we change a schema we see (possible just a dummy schema with $ref you created) to the real schema. This is possible since all schemas have already been loaded at this point.

darrelmiller:
@PerthCharern Yeah, doing a post process to hook up the cycles will be my fallback approach. I'll see if I can find a cleaner way first.

darrelmiller:
@PerthCharern After failing to find a solution to resolve these $ref on first pass, I'm thinking that using a second pass to resolve the $refs might actually address another concern.  In another branch I have been working on resolving $refs to external documents and have the problem that external $refs need to be loaded asynchronously.  

Perhaps reading can be a three step process:

- Read graph and don't resolve $refs.  
- Iterate over $refs looking for external $refs and load the documents
- Walk the graph and convert $refs into resolved objects.

I'm not sure how to expose the API for this, but at least knowing that it could help with resolving external references gives me more confidence in breaking this down into multiple steps

PerthCharern:
@darrelmiller That sounds good to me. I don't have any concerns with doing multiple passes to complete the reading process. 

Regarding exposing these as APIs, I am afraid I don't have much input for you now. One thing is we should expose the API to resolve external references independently since that sounds pretty useful as a standalone feature.

Given that the logic would be similar regardless of what we expose, I think we can also discuss once you send out a PR.

darrelmiller:
** NEW BEHAVIOUR **

Ok, this PR is finally ready for review. 
- Component objects are new loaded with Reference object populated
- Initial parsing phase does not resolve references.  All references are marked as "unresolved".
- After parsing, the DOM is traversed and references are resolved and replaced with complete objects
- ReaderSettings now have a property to control the reference resolution.  If desired, resolution process can be skipped.
- Cycles due to self $refs are now supported
- The "referenceStore" has been removed from ParsingContext as it is no longer needed

Remote references are not yet supported.  A new branch has been started to support that.  There will need to be a new ReadAsync method to support resolving remote references.

/cc @PerthCharern @xuzhg

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -225,6 +225,11 @@ public class OpenApiSchema : IOpenApiSerializable, IOpenApiReferenceable, IOpenA
         /// </summary>
         public IDictionary<string, IOpenApiExtension> Extensions { get; set; } = new Dictionary<string, IOpenApiExtension>();
 
+        /// <summary>
+        /// Indicates object is a placeholder reference to an actual object and does not contain valid data.
+        /// </summary>
+        public bool UnresolvedReference { get; set; } = false;
```
PerthCharern: >= false; [](start = 54, length = 8)

nit, remove

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -264,5 +265,81 @@ private static void WriteHostInfoV2(IOpenApiWriter writer, IList<OpenApiServer>
...
+
+            if (reference.IsExternal)
+            {
+                // Should not attempt to resolve external references against a single document.
+                throw new ArgumentException(); //TODO Add error message
+            }
+
+            if (!reference.Type.HasValue)
+            {
+                throw new ArgumentException("Local reference must have type specified.");
```
PerthCharern: >"Local reference must have type specified." [](start = 44, length = 43)

nit, This should be a resource for consistency

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
```diff
@@ -144,10 +144,18 @@ public virtual void Visit(OpenApiRequestBody requestBody)
...
+        /// <summary>
+        /// Visits headers.
+        /// </summary>
+        public virtual void Visit(IDictionary<string, OpenApiHeader> headers)
+        {
+        }
+
         /// <summary>
-        /// Visits responses.
+        /// Visits headers.
```
PerthCharern: >headers [](start = 19, length = 7)

callbacks

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -449,11 +573,56 @@ internal void Walk(OpenApiRequestBody requestBody)
...
+            {
+                return;
+            }
+
+            _visitor.Visit(headers);
+            if (headers != null)
+            {
+                foreach (var header in headers)
+                {
+                    Walk(header.Key.Replace("/", "~1"), () => Walk(header.Value));
```
PerthCharern: >.Replace("/", "~1") [](start = 35, length = 19)

nit, I thought we already handled this somewhere else. If not, can we do this in the VisitorBase.Enter, given that this is a JSON pointer requirement.

src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
```diff
@@ -10,11 +10,35 @@
...
+        /// </summary>
+        DoNotResolveReferences,
+        /// <summary>
+        /// Convert local references to references of valid domain objects.
+        /// </summary>
+        ResolveLocalReferences,
+        /// <summary>
+        /// Convert all references to references of valid domain objects.
+        /// </summary>
+        ResolveRemoteReferences
```
PerthCharern: >ResolveRemoteReferences [](start = 8, length = 23)

either ResolveAllReferences or ResolveLocalAndRemoteReferences

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -60,6 +61,21 @@ public OpenApiDocument Read(Stream input, out OpenApiDiagnostic diagnostic)
             // Parse the OpenAPI Document
             var document = context.Parse(yamlDocument, diagnostic);
 
+            // Resolve References if requested
+            switch (_settings.ReferenceResolution) 
+            {
+                case ReferenceResolutionSetting.ResolveRemoteReferences:
+                    throw new ArgumentException(Properties.SRResource.CannotResolveRemoteReferencesSynchronously);
```
PerthCharern: >throw new ArgumentException(Properties.SRResource.CannotResolveRemoteReferencesSynchronously); [](start = 20, length = 94)

This will be supported, right? Can we just go with NotImplementedException()?

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -60,6 +61,21 @@ public OpenApiDocument Read(Stream input, out OpenApiDiagnostic diagnostic)
...
+                case ReferenceResolutionSetting.ResolveLocalReferences:
+                    var resolver = new OpenApiReferenceResolver(document);
+                    var walker = new OpenApiWalker(resolver);
+                    walker.Walk(document);
+                    break;
+                case
+                    ReferenceResolutionSetting.DoNotResolveReferences:
```
PerthCharern: nit, remove this awkward line :)

src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -201,5 +156,52 @@ public void StartObject(string objectName)
...
+            {
+                return false;  // Loop detected
+            }
+        }
+
+        /// <summary>
+        /// Reset loop tracking stack
+        /// </summary>
+        /// <param name="loopid"></param>
+        internal void ClearLoop(string loopid)
```
PerthCharern: >loopid [](start = 39, length = 6)

nit, loopId

src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -201,5 +156,52 @@ public void StartObject(string objectName)
...
+        {
+            _loopStacks[loopid].Clear();
+        }
+
+        /// <summary>
+        /// Exit from the context in cycle detection
+        /// </summary>
+        /// <param name="loopid"></param>
+        public void PopLoop(string loopid)
```
PerthCharern: >loopid [](start = 35, length = 6)

nit, loopId

src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -201,5 +156,52 @@ public void StartObject(string objectName)
         {
             _currentLocation.Push(objectName);
         }
+
+        /// <summary>
+        /// Maintain history of traversals to avoid stack overflows from cycles
+        /// </summary>
+        /// <param name="loopId">Any unique identifier for a stack</param>
+        /// <param name="key">Identifier used to </param>
+        /// <returns></returns>
```
PerthCharern: > /// <returns></returns> [](start = 7, length = 24)

nit, you can just remove these non-filled tags

PerthCharern: Actually, for this one, you might want to fill since it's not that obvious what the returned bool represents.

---
In reply to: [167710868](https://github.com/Microsoft/OpenAPI.NET/pull/185#discussion_r167710868) [](ancestors = 167710868)

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -368,11 +458,18 @@ internal void Walk(IDictionary<OperationType, OpenApiOperation> operations)
...
+                return;
+            }
+
             _visitor.Visit(operation);
 
             Walk(OpenApiConstants.Parameters, () => Walk(operation.Parameters));
             Walk(OpenApiConstants.RequestBody, () => Walk(operation.RequestBody));
             Walk(OpenApiConstants.Responses, () => Walk(operation.Responses));
+            Walk(OpenApiConstants.Callbacks, () => Walk(operation.Callbacks));
+
```
PerthCharern: Walk the Tags array too?

src/Microsoft.OpenApi.Readers/Services/OpenApiReferenceResolver.cs
```diff
@@ -0,0 +1,207 @@
...
+        public override void Visit(OpenApiComponents components)
+        {
+            ResolveMap(components.Parameters);
+            ResolveMap(components.RequestBodies);
+            ResolveMap(components.Responses);
+            ResolveMap(components.Links);
+            ResolveMap(components.Callbacks);
+            ResolveMap(components.Examples);
+            ResolveMap(components.Schemas);
+            ResolveMap(components.SecuritySchemes);
```
PerthCharern: Shouldn't it be the job of the walker to do this "walking"? 

Shouldn't this file just do reference-resolving for the object that's actually a reference and then leave the logic of getting into that object to the Walker?

src/Microsoft.OpenApi.Readers/V3/OpenApiSchemaDeserializer.cs
```diff
@@ -234,15 +234,21 @@ internal static partial class OpenApiV3Deserializer
             {s => s.StartsWith("x-"), (o, p, n) => o.AddExtension(p, n.CreateAny())}
         };
 
+        private const string schemaLoopId = "schema";
```
PerthCharern: >  private const string schemaLoopId = "schema"; [](start = 6, length = 47)

remove

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiSchemaTests.cs
```diff
@@ -424,11 +444,64 @@ public void ParseAdvancedSchemaWithReferenceShouldSucceed()
...
+                //        }
+                //    },
+                //    Reference = new OpenApiReference()
+                //    {
+                //        Type = ReferenceType.Schema,
+                //        Id = "microsoft.graph.schemaExtension"
+                //    }
+                //};
+
+                //schemaExtension.AllOf[0].Properties["child"] = schemaExtension;
```
PerthCharern: Does this fail the unit test due to the loop? If so, you can set xUnit to only delve down until specified level just for this test

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
darrelmiller: I don't think this will be handled.  If you want to resolve remote references then you should call the ReadAsync method instead.  The ReadAsync will also load related documents into a workspace.

src/Microsoft.OpenApi.Readers/Services/OpenApiReferenceResolver.cs
darrelmiller: I wish I could do that.  However, we replace the "unresolved" object that is empty, except for the Reference, with the completely filled object.  Therefore, I need to detect the reference from the parent object that holds the pointer to the unresolved object so that we can change it.

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiSchemaTests.cs
darrelmiller: I couldn't figure out why it was failing.  I kept getting an error that made no sense.  Anyway, I just tried running it again and it is working now.

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
PerthCharern: Makes sense. 

---
In reply to: [167757446](https://github.com/Microsoft/OpenAPI.NET/pull/185#discussion_r167757446) [](ancestors = 167757446)

src/Microsoft.OpenApi.Readers/Services/OpenApiReferenceResolver.cs
PerthCharern: That is true. It's kinda unfortunate since that means we can't fully leverage the Walking capability. 

If we revise the Walk method design to take in another variable (in this case the parent object), and then it can make changes to that variable, then that would work. But that's a big change, and it doesn't seem clean.

darrelmiller: @PerthCharern Yes, the walker is not ideal in this scenario.  However, it is expected to use the parent object to do certain types of work.  e.g. Checking if a value is not null, needs to be done from above.  Checking if there are no duplicate parameter names.  Checking if two peer values are not contradicting each other.

Passing in the parent object is hard because you don't know its type.  It would need to be passed in as an object and then cast to each of the possible parent types.  In the end I think it is just as easy to do certain contextually sensitive work from the parent visit method.

PR #186, opened 2018-01-19
Fix formData/body parameter and consumes handling

### LINKED ISSUE #178: Produces/consumes in V2
Currently consumes label is not affecting the body content in the reader.
for example:

/uploadImage": {
			"post": {
				"tags": ["image"],
				"summary": "uploads an image",
				"description": "",
				"operationId": "uploadFile",
				"consumes": ["multipart/form-data"],
				"produces": ["application/json"],
				"parameters": [{
					"name": "Id",
					"in": "path",
					"description": "ID to update",
					"required": true,
					"type": "integer",
					"format": "int64"
				}],
				"responses": {
					"200": {
						"description": "successful operation",
						"schema": {
							"$ref": "#/definitions/ApiResponse"
						}
					}
				}
			}
		}

gives a "application/json" as body content and not "multipart/form-data" as should be.
I tried to change the consumes value to "application/xml" but still doesn't work.

### LINKED ISSUE #179: formData parameter in V2 is not supported
Parameters in formData are ignored by the reader.

fot example:
```json
"/uploadImage": {
			"post": {
				"tags": ["upload"],
				"summary": "uploads an image",
				"description": "",
				"operationId": "uploadFile",
				"consumes": ["application/xml"],
				"produces": ["application/json"],
				"parameters": [
				{
					"name": "additionalMetadata",
					"in": "formData",
					"description": "Additional data to pass to server",
					"required": true,
					"type": "string"
				}],
				"responses": {
					"200": {
						"description": "successful operation",
						"schema": {
							"$ref": "#/definitions/ApiResponse"
						}
					}
				}
			}
		}
```

is ignores the parameter

PerthCharern:
- Fix the serialization and deserialization logic for V2 formData and body parameter.

- V2 operation deserialization now respects the `consumes` property.

- Unit tests verify that serializing, deserializing, and re-deserializing all work as expected.

Fix #178 
Fix #179

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
@@ -92,13 +92,15 @@ namespace Microsoft.OpenApi.Readers.V2
                 {s => s.StartsWith("x-"), (o, p, n) => o.AddExtension(p, n.CreateAny())}
             };
 
-        private static FixedFieldMap<OpenApiResponses> _responsesFixedFields = new FixedFieldMap<OpenApiResponses>();
+        private static readonly FixedFieldMap<OpenApiResponses> _responsesFixedFields =
+            new FixedFieldMap<OpenApiResponses>();
 
-        private static PatternFieldMap<OpenApiResponses> _responsesPatternFields = new PatternFieldMap<OpenApiResponses>
-        {
-            {s => !s.StartsWith("x-"), (o, p, n) => o.Add(p, LoadResponse(n))},
-            {s => s.StartsWith("x-"), (o, p, n) => o.AddExtension(p, n.CreateAny())}
-        };
+        private static readonly PatternFieldMap<OpenApiResponses> _responsesPatternFields =
+            new PatternFieldMap<OpenApiResponses>
+            {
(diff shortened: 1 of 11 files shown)


PerthCharern:
cc @talamram

src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
```diff
@@ -137,22 +139,33 @@ public static OpenApiResponses LoadResponses(ParseNode node)
...
+                        {
+                            var schema = v.Schema;
+                            schema.Description = v.Description;
+                            return schema;
+                        }),
+                    Required = formParameters.Where(p => p.Required).Select(p => p.Name).ToList()
                 }
             };
 
+            var consumes = context.GetFromTempStorage<List<string>>("operationconsumes") ??
```
darrelmiller: It is worth noting that this will import a valid V2 document correctly, but it will also create an invalid DOM if the media type is not a valid one.  Validating this is going to be tricky, so I'm ok if we punt on checking for that.

PerthCharern: Agree. It’s a bit complicated to add “validation logic” into the reader at this point. I’d rather wait for your PR to be completed and we can retroactively clean-up these validation grey areas later.

Issue #192, opened 2018-01-25
Handle empty URI when reading


PerthCharern:

sergey-tihon:
Schemas
- https://api.apis.guru/v2/specs/azure.com/storageimportexport/2016-11-01/swagger.json

Stack Trace
```
System.UriFormatException: Invalid URI: The URI is empty.
  at System.Uri.CreateThis (System.String uri, System.Boolean dontEscape, System.UriKind uriKind) [0x0007b] in <a24e193d5cf14a42ac00ebaea26ec7bc>:0 
  at System.Uri..ctor (System.String uriString) [0x00014] in <a24e193d5cf14a42ac00ebaea26ec7bc>:0 
  at Microsoft.OpenApi.Readers.V2.OpenApiV2Deserializer+<>c.<.cctor>b__65_26 (Microsoft.OpenApi.Models.OpenApiExternalDocs o, Microsoft.OpenApi.Readers.ParseNodes.ParseNode n) [0x00007] in <ae665b1781534df6a22c871b7a998f76>:0 
  at Microsoft.OpenApi.Readers.ParseNodes.PropertyNode.ParseField[T] (T parentInstance, System.Collections.Generic.IDictionary`2[TKey,TValue] fixedFields, System.Collections.Generic.IDictionary`2[TKey,TValue] patternFields) [0x0002b] in <ae665b1781534df6a22c871b7a998f76>:0 
  at Microsoft.OpenApi.Readers.V2.OpenApiV2Deserializer.ParseMap[T] (Microsoft.OpenApi.Readers.ParseNodes.MapNode mapNode, T domainObject, Microsoft.OpenApi.Readers.ParseNodes.FixedFieldMap`1[T] fixedFieldMap, Microsoft.OpenApi.Readers.ParseNodes.PatternFieldMap`1[T] patternFieldMap, System.Collections.Generic.List`1[T] requiredFields) [0x00014] in <ae665b1781534df6a22c871b7a998f76>:0 
  at Microsoft.OpenApi.Readers.V2.OpenApiV2Deserializer.LoadExternalDocs (Microsoft.OpenApi.Readers.ParseNodes.ParseNode node) [0x00010] in <ae665b1781534df6a22c871b7a998f76>:0 
[...]

PR #197, opened 2018-01-30
Fix temp parameters duplicate issue + Allow relative URIs where appropriate

### LINKED ISSUE #191: "formparameters" temp variable must be cleared after each operation is processed

PerthCharern:
- Handle URI parsing. Allows relative URI where allowed by the spec. Partially addressed #192.

- Temp variables in Operation deserializer (formParameters, bodyParameter, operationConsumes, operationProduces) are set to null at the beginning of each operation processing. Fix #191


PerthCharern:
cc @sergey-tihon

sergey-tihon:
Just a question, should this PR also consider URLs in `$ref`s?

Sample 

```json
"enum": {
  "type": "array",
  "items": {
    "$ref": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/apiVersion"
  },
  "minItems": 1,
  "uniqueItems": true
}
```
from https://github.com/Azure/azure-resource-manager-schemas/blob/master/tools/ResourceMetaSchema.json#L160

PerthCharern:
@darrelmiller is working on the support for external references. That feature will likely come as a separate PR.

src/Microsoft.OpenApi.Readers/V2/OpenApiSecurityRequirementDeserializer.cs
```diff
@@ -33,8 +33,8 @@ public static OpenApiSecurityRequirement LoadSecurityRequirement(ParseNode node)
                 }
                 else
                 {
-                    node.Diagnostic.Errors.Add(
-                        new OpenApiError(node.Context.GetLocation(), $"Scheme {property.Name} is not found"));
+                    //node.Diagnostic.Errors.Add(
```
sergey-tihon: What should happen if `scheme` is `null`?

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiXmlTests.cs
```diff
@@ -46,5 +46,35 @@ public void ParseBasicXmlShouldSucceed()
...
+
+                // Act
+                var xml = OpenApiV3Deserializer.LoadXml(node);
+
+                // Assert
+                xml.ShouldBeEquivalentTo(
+                    new OpenApiXml
+                    {
+                        Name = "name1",
+                        // TODO: Verify that the URL caused a parsing error.
```
sergey-tihon: Looks like a candidate for one more assertion for `diagnostic `.

sergey-tihon:
I do not familiar with codebase yet, but there are 2 comments from me on the diff 😊

src/Microsoft.OpenApi.Readers/V2/OpenApiSecurityRequirementDeserializer.cs
darrelmiller: The spec says 

> the name used for each property MUST correspond to a security scheme declared in the Security Schemes under the Components Object.
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#security-requirement-object

So I believe we should mark this as an error.  Not sure why this has been commented out /cc @PerthCharern

src/Microsoft.OpenApi.Readers/V2/OpenApiExternalDocsDeserializer.cs
```diff
@@ -26,7 +26,11 @@ internal static partial class OpenApiV2Deserializer
                 {
                     "url", (o, n) =>
                     {
-                        o.Url = new Uri(n.GetScalarValue());
+                        Uri uri;
+                        if (Uri.TryCreate(n.GetScalarValue(), UriKind.RelativeOrAbsolute, out uri))
+                        {
+                            o.Url = uri;
+                        }
```
darrelmiller: I'm not sure I'm a fan of this approach.  We are going to swallow the error and when reporting the problem all we can say is "you are missing a URL". This is going to be confusing to the person who believes they put a URL and it is invalid.  I think it is valid to have two distinct sources of errors.  One from parsing content, and one from validating the DOM.  We can't avoid it anyway because sometimes the YAML/JSON is going to be badly formatted.

src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
```diff
@@ -104,6 +105,12 @@ internal static partial class OpenApiV2Deserializer
 
         internal static OpenApiOperation LoadOperation(ParseNode node)
         {
+            // Reset these temp storage parameters for each operation.
+            node.Context.SetTempStorage("bodyParameter", null);
+            node.Context.SetTempStorage("formParameters", null);
+            node.Context.SetTempStorage("operationProduces", null);
+            node.Context.SetTempStorage("operationConsumes", null);
```
darrelmiller: I guess we should probably create constants for these.

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
```diff
@@ -37,7 +39,7 @@ internal static partial class OpenApiV2Deserializer
             {
                 "schema", (o, n) =>
                 {
-                    n.Context.SetTempStorage("operationschema", LoadSchema(n));
+                    _responseSchema = LoadSchema(n);
```
darrelmiller: Why this change to a static?  This is going to make the deserializer not thread safe.  That is most likely going to be a problem.

src/Microsoft.OpenApi.Readers/V2/OpenApiV2VersionService.cs
```diff
@@ -66,6 +66,14 @@ public bool TryLoadReference(ParsingContext context, OpenApiReference reference,
...
             var node = context.RootNode.Find(jsonPointer);
 
+            if (node == null)
+            {
+                throw new OpenApiException(
+                    string.Format(
+                        SRResource.JsonPointerCannotBeResolved,
+                        jsonPointer));
+            }
+
```
darrelmiller: Maybe once we consolidate the error classes we can just throw this exception from inside the Find.  Is there somewhere where we need the response to be null on fail?

PerthCharern:
Apologies for the delay on this, and thanks for all your comments (@sergey-tihon, @darrelmiller)

I'll try to address all the comments by end of this week.

src/Microsoft.OpenApi.Readers/V2/OpenApiExternalDocsDeserializer.cs
PerthCharern: We should still produce the document though. If I throw an error here, everything breaks.

We need to have a way to still allow reading to go through and add the errors along the way.

---
In reply to: [165092404](https://github.com/Microsoft/OpenAPI.NET/pull/197#discussion_r165092404) [](ancestors = 165092404)

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
PerthCharern: You are right. I wasn't thinking about that issue. I'll revert this back. There are a few instances like this in the code, so I'll fix those separately.

---
In reply to: [165093793](https://github.com/Microsoft/OpenAPI.NET/pull/197#discussion_r165093793) [](ancestors = 165093793)

src/Microsoft.OpenApi.Readers/V2/OpenApiSecurityRequirementDeserializer.cs
PerthCharern: I commented this out so that I could test without running into NRE and forgot to revert back. 

---
In reply to: [165090553](https://github.com/Microsoft/OpenAPI.NET/pull/197#discussion_r165090553) [](ancestors = 165090553)

PerthCharern:
Update: I got booked up with other commitments this week, but I'm almost done. I'll update the PR next week. Sorry again for the delay.

src/Microsoft.OpenApi.Readers/V2/OpenApiExternalDocsDeserializer.cs
PerthCharern: Without that mechanism, what we can do is making the reporting message a bit more generic:

"The string passed in cannot be parsed into a URL."

This should be a very rare case. Almost any string can be a url. Even "abcdef" is allowed to be parsed in as a URL. Cases I suspect will be caught would be something like an empty string or a white space string.

---
In reply to: [167382549](https://github.com/Microsoft/OpenAPI.NET/pull/197#discussion_r167382549) [](ancestors = 167382549,165092404)

src/Microsoft.OpenApi.Readers/V2/OpenApiV2VersionService.cs
PerthCharern: Throwing the error here allows us to control what to do with json pointers that we cannot resolve. At least, we can leave them as a reference object (and reporting it) rather than always throwing an error and stopping the entire parsing process.

This probably no longer applies with the new reference resolving method.

---
In reply to: [165094630](https://github.com/Microsoft/OpenAPI.NET/pull/197#discussion_r165094630) [](ancestors = 165094630)

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiXmlTests.cs
PerthCharern: I changed this to a success case since that's what the spec allows (relative url should also work).

---
In reply to: [164986283](https://github.com/Microsoft/OpenAPI.NET/pull/197#discussion_r164986283) [](ancestors = 164986283)

PerthCharern: I created a new test case with empty URL. The diagnostic should contain an error for empty URL.

---
In reply to: [168329298](https://github.com/Microsoft/OpenAPI.NET/pull/197#discussion_r168329298) [](ancestors = 168329298,164986283)

PerthCharern:
I've been thinking a bit more about this. Here's what I'll do:

1) I'll revert URL-related changes from this PR. This is a bigger problem than I thought, so I'll tackle it separately. Here's what I propose: We change the models to use "string" instead of "Uri" type. This way, we can handle all the format validation in the validator without failing the parsing. This falls in line with our current paradigm (reader just reads, validator verifies the format).

2) I'll address all other comments and send out a new commit for this PR.

PerthCharern:
@darrelmiller This PR has become much smaller and is ready now. I will handle #192 in a separate PR since it turned out to be a bit more involved than I expected. What do you think about my proposal in 1. below?

src/Microsoft.OpenApi.Readers/ParseNodes/JsonPointerExtensions.cs
```diff
@@ -47,9 +47,9 @@ public static YamlNode Find(this JsonPointer currentpointer, YamlNode baseYamlNo
 
                 return pointer;
             }
-            catch (Exception ex)
+            catch (Exception)
             {
-                throw new ArgumentException("Failed to dereference pointer", ex);
+                return null;
```
darrelmiller: I'm still a little nervous about this change, but let's try it and see if it causes any issues.

darrelmiller:
:shipit:

PR #200, opened 2018-01-31
Re-enable the default rule count test case

PerthCharern:
Re-enable the default rule count test case (disabled in #176)

The problem was this:

ValidateCustomExtension test calls ValidationRuleSet.DefaultRuleSet and makes changes to it directly. Since the DefaultRuleSet is cached, the underlying private `_defaultRuleSet` is actually modified.

The reason we did not see the issue locally might be that the order of local test run and the order on AppVeyor are different. We see the issue only if the "Count" test is run after the custom extension test.

### DIFF
--- a/src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
+++ b/src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
@@ -24,17 +24,22 @@ namespace Microsoft.OpenApi.Validations
         /// <summary>
         /// Gets the default validation rule sets.
         /// </summary>
-        public static ValidationRuleSet DefaultRuleSet
+        /// <remarks>
+        /// This is a method instead of a property to signal that a new default ruleset object is created
+        /// per call. Making this a property may be misleading callers to think the returned rulesets from multiple calls
+        /// are the same objects.
+        /// </remarks>
+        public static ValidationRuleSet GetDefaultRuleSet()
         {
-            get
+            // Reflection can be an expensive operation, so we cache the default rule set that has already been built.
+            if (_defaultRuleSet == null)
             {
-                if (_defaultRuleSet == null)
-                {
-                    _defaultRuleSet = BuildDefaultRuleSet();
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -24,17 +24,22 @@ public sealed class ValidationRuleSet : IEnumerable<ValidationRule>
...
-                    _defaultRuleSet = BuildDefaultRuleSet();
-                }
-
-                return _defaultRuleSet;
+                _defaultRuleSet = BuildDefaultRuleSet();
             }
+
+            // We create a new instance of ValidationRuleSet per call as a safeguard
+            // against unintentional modification of the private _defaultRuleSet.
+            return new ValidationRuleSet(_defaultRuleSet);
```
PerthCharern: >        return new ValidationRuleSet(_defaultRuleSet); [](start = 4, length = 54)

We could alternatively prescribe that the caller never modifies the DefaultRuleSet and always uses it like this if they need to use default rules + more rules:

new ValidationRuleSet(ValidationRuleSet.DefaultRuleSet)

But then that's kinda risky and people will not always follow the best practice.

Creating the new instance here safeguards against that scenario as well.

darrelmiller: Creating the new collection on each call seems like the best approach.  The rules don't need to get re-instantiated so the cost should be minimal.  The other option would be to return a ReadOnlyCollection.

BTW, thanks for finding this.  I was looking in the completely wrong direction.

PerthCharern: No problem. 

A ReadOnlyCollection is something I wanted to do, but it will require more changes since we return the default rule set now as a class object, not as a list/dictionary of rules. I'll go with the current implementation for now, and we can revisit if this turns out not working well.

Issue #149, opened 2017-12-15
NuGet package needs a logo.


PerthCharernZero:

PerthCharern:
We have the logo now. I'll upload it when I can.

PerthCharern:
@darrelmiller We still don't have a logo, do we? We can use Microsoft logo for now, if that's preferred.

PR #206, opened 2018-02-05
Add Nuget logo file

PerthCharern:
PR to update nuspec to point to the logo will come separately. 

Part 1 of #149


sergey-tihon:
Why does it look like Swagger but not like OpenAPI?
logo

PerthCharern:
@sergey-tihon Good question. We might have to think this through a bit more. I'll close this PR for now and will send out a new PR when ready.

Issue #199, opened 2018-01-31
Float and Double values cannot be used where an IOpenAny is expected (JSON V3)


joelwkall:
### Summary
For example in the Example property, we need to convert any float or double to OpenApiFloat or OpenApiDouble, respectively. 

However, when serializing, only "decimal" is accepted as a floating point type, but there is no OpenApiDecimal. 

### Reproduction
Create a OpenApiSchema, set the Example property to new OpenApiDouble(5.5) or equivalent. Serialize the schema with V3 and JSON format.

### Expected results
The example property be serialized to a JSON number ("Example":5.5)

### Actual results
The serialization (version V3) fails with "The type 'System.Double' is not supported in Open API document."

### Workaround
I don't know how to work around this, since the writer only accepts decimal, but there is no OpenApiDecimal type.

### Suggested solutions
Either
1. Introduce a OpenApiDecimal type, or
2. Allow the V3 JSON writer to accept double and/or float values

PerthCharern:
Thanks @joelwkall for catching this, and for the thorough analysis of the issue!

This is indeed a problem, and I agree either of your suggested solution should work.

- I think 2 should be done regardless, so that these C# primitive types can be serialized. People might bring these primitive types into their custom extensions.

- I am also thinking we should 
   - only have one OpenApiDecimal type to represent OpenApiAny that's a floating point and
   - get rid of OpenApiFloat and OpenApiDouble. 

  The OpenAPI spec itself doesn't differentiate between these types, so there's no point having multiple OpenApiAny's to support them. (Supporting the primitive C# decimal, float, double should still be done since users can still bring in those types into Example/Extensions.)

Would you be able to help experiment and send out a PR?

joelwkall:
I can take a look. 

There is also OpenApiLong which is probably superfluous since there is only one integer type in the OpenApiSpec. Are there more unnecessary types?

PerthCharern:
I would agree with you. I think the original idea was that we tried to mimic this data table:

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#data-types

I don't think that makes much sense given how we are using the OpenApiAny types. I'd say we should more closely align with the C# types.

In that sense, I think we can also get rid of: Long, Password, Binary, (and possibly Null since we can simply use the C# null anyway).

joelwkall:
Allright, seems fair. 

[...]

PR #209, opened 2018-02-06
Support all numeric IOpenApiAny types in OpenApiWriterBase

joelwkall:
The purpose of this PR is make sure that OpenApiFloat, OpenApiDouble, and OpenApiLong are supported in the default serializers (which inherits from OpenApiWriterBase).

Fixes issue https://github.com/Microsoft/OpenAPI.NET/issues/199

Some questsions:

1. I first added tests for long, float and double. Float and double tests failed expectedly, but the long tests somehow passed, even though no overloads or cases for long had yet been added. I dug around but could not find the reason why this is happening. Is the runtime implicitly casting the longs as ints? The tests still pass using long.MaxValue, which should rule out that possibility. I still added an overload and a case for long, for consistency. Please comment.

2. Is the tests I wrote for double and float satisfactory? Do we need to take culture or formatting in account?

### DIFF
--- a/src/Microsoft.OpenApi/Writers/OpenApiWriterBase.cs
+++ b/src/Microsoft.OpenApi/Writers/OpenApiWriterBase.cs
@@ -111,6 +111,26 @@ namespace Microsoft.OpenApi.Writers
         /// </summary>
         /// <param name="value">The string value.</param>
         public abstract void WriteValue(string value);
+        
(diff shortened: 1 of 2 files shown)


PerthCharern:
I think this is okay. If this issue shows up again in another capacity, we can tackle it then.

Thanks again for the PR!

PR #211, opened 2018-02-08
Do not mark require property as read only

Shwetap05:
V2 serialization code changes to not allow a child property of a schema to be readonly if its inlcuded in the list of required properties of parent schema, to comply with V2 spec.

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -375,7 +375,7 @@ namespace Microsoft.OpenApi.Models
         /// </summary>
         public void SerializeAsV2(IOpenApiWriter writer)
         {
-            SerializeAsV2(writer, new List<string>(), null);
+            SerializeAsV2(writer: writer, parentRequiredProperties: new List<string>(), propertyName: null);
         }
 
         /// <summary>
@@ -383,7 +383,10 @@ namespace Microsoft.OpenApi.Models
         /// </summary>
         public void SerializeAsV2WithoutReference(IOpenApiWriter writer)
         {
-            SerializeAsV2WithoutReference(writer, new List<string>(), null);
+            SerializeAsV2WithoutReference(
+                writer: writer,
+                parentRequiredProperties: new List<string>(),
+                propertyName: null);
         }
 
         /// <summary>
@@ -393,7 +396,9 @@ namespace Microsoft.OpenApi.Models
         /// <param name="writer">The open api writer.</param>
         /// <param name="parentRequiredProperties">The list of required properties in parent schema.</param>
         /// <param name="propertyName">The property name that will be serialized.</param>
-        internal void SerializeAsV2(IOpenApiWriter writer, IList<string> parentRequiredProperties,
+        internal void SerializeAsV2(
+            IOpenApiWriter writer,
+            IList<string> parentRequiredProperties,
             string propertyName)
         {
             if (writer == null)
@@ -422,7 +427,9 @@ namespace Microsoft.OpenApi.Models
         /// <param name="writer">The open api writer.</param>
         /// <param name="parentRequiredProperties">The list of required properties in parent schema.</param>
         /// <param name="propertyName">The property name that will be serialized.</param>
-        internal void SerializeAsV2WithoutReference(IOpenApiWriter writer, IList<string> parentRequiredProperties,
+        internal void SerializeAsV2WithoutReference(
+            IOpenApiWriter writer,
+            IList<string> parentRequiredProperties,
             string propertyName)
         {
             writer.WriteStartObject();
@@ -494,7 +501,9 @@ namespace Microsoft.OpenApi.Models
             writer.WriteExtensions(Extensions);
         }
 
-        internal void WriteAsSchemaProperties(IOpenApiWriter writer, IList<string> parentRequiredProperties,
+        internal void WriteAsSchemaProperties(
+            IOpenApiWriter writer,
+            IList<string> parentRequiredProperties,
             string propertyName)
         {
             if (writer == null)
@@ -582,14 +591,11 @@ namespace Microsoft.OpenApi.Models
             writer.WriteProperty(OpenApiConstants.Discriminator, Discriminator?.PropertyName);
 
             // readOnly
-            // In V2 schema if a property is part of required properties of schema, it cannot be marked as readonly.
-            if (parentRequiredProperties.Contains(propertyName))
+            // In V2 schema if a property is part of required properties of parent schema,
+            // it cannot be marked as readonly.
+            if (!parentRequiredProperties.Contains(propertyName))
             {
-                writer.WriteProperty(OpenApiConstants.ReadOnly, false, false);
-            }
-            else
-            {
-                writer.WriteProperty(OpenApiConstants.ReadOnly, ReadOnly, false);
+                writer.WriteProperty(name: OpenApiConstants.ReadOnly, value: ReadOnly, defaultValue: false);
             }
 
             // xml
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -549,7 +582,15 @@ internal void WriteAsSchemaProperties(IOpenApiWriter writer)
             writer.WriteProperty(OpenApiConstants.Discriminator, Discriminator?.PropertyName);
 
             // readOnly
-            writer.WriteProperty(OpenApiConstants.ReadOnly, ReadOnly, false);
+            // In V2 schema if a property is part of required properties of schema, it cannot be marked as readonly.
```
Shwetap05: >schem [](start = 76, length = 5)

Nit. parent schema

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -374,6 +374,27 @@ public void SerializeAsV3WithoutReference(IOpenApiWriter writer)
         /// Serialize <see cref="OpenApiSchema"/> to Open Api v2.0
         /// </summary>
         public void SerializeAsV2(IOpenApiWriter writer)
+        {
+            SerializeAsV2(writer, new List<string>(), null);
```
PerthCharern: > new List<string>(), null [](start = 33, length = 25)

nit, suggest using named param for clarify

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -374,6 +374,27 @@ public void SerializeAsV3WithoutReference(IOpenApiWriter writer)
...
+        }
+
+        /// <summary>
+        /// Serialize to OpenAPI V2 document without using reference.
+        /// </summary>
+        public void SerializeAsV2WithoutReference(IOpenApiWriter writer)
+        {
+            SerializeAsV2WithoutReference(writer, new List<string>(), null);
```
PerthCharern: >new List<string>(), null) [](start = 50, length = 25)

nit, suggest using named param for clarify

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -374,6 +374,27 @@ public void SerializeAsV3WithoutReference(IOpenApiWriter writer)
...
+        }
+
+        /// <summary>
+        /// Serialize <see cref="OpenApiSchema"/> to Open Api v2.0 and handles not marking the provided property 
+        /// as readonly if its included in the provided list of required properties of parent schema.
+        /// </summary>
+        /// <param name="writer">The open api writer.</param>
+        /// <param name="parentRequiredProperties">The list of required properties in parent schema.</param>
+        /// <param name="propertyName">The property name that will be serialized.</param>
+        internal void SerializeAsV2(IOpenApiWriter writer, IList<string> parentRequiredProperties,
```
PerthCharern: >IOpenApiWriter writer, IList<string> parentRequiredProperties, [](start = 36, length = 62)

nit, suggest each parameter on its own line for readability

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -386,16 +407,26 @@ public void SerializeAsV2(IOpenApiWriter writer)
...
         /// <summary>
-        /// Serialize to OpenAPI V2 document without using reference.
+        /// Serialize to OpenAPI V2 document without using reference and handles not marking the provided property 
+        /// as readonly if its included in the provided list of required properties of parent schema.
         /// </summary>
-        public void SerializeAsV2WithoutReference(IOpenApiWriter writer)
+        /// <param name="writer">The open api writer.</param>
+        /// <param name="parentRequiredProperties">The list of required properties in parent schema.</param>
+        /// <param name="propertyName">The property name that will be serialized.</param>
+        internal void SerializeAsV2WithoutReference(IOpenApiWriter writer, IList<string> parentRequiredProperties,
```
PerthCharern: >IOpenApiWriter writer, IList<string> parentRequiredProperties, [](start = 52, length = 62)

nit, suggest each parameter on its own line for readability

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -463,7 +494,8 @@ internal void WriteAsItemsProperties(IOpenApiWriter writer)
             writer.WriteExtensions(Extensions);
         }
 
-        internal void WriteAsSchemaProperties(IOpenApiWriter writer)
+        internal void WriteAsSchemaProperties(IOpenApiWriter writer, IList<string> parentRequiredProperties,
+            string propertyName)
```
PerthCharern: nit, suggest each parameter on its own line for readability

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -549,7 +582,15 @@ internal void WriteAsSchemaProperties(IOpenApiWriter writer)
             writer.WriteProperty(OpenApiConstants.Discriminator, Discriminator?.PropertyName);
 
             // readOnly
-            writer.WriteProperty(OpenApiConstants.ReadOnly, ReadOnly, false);
+            // In V2 schema if a property is part of required properties of schema, it cannot be marked as readonly.
+            if (parentRequiredProperties.Contains(propertyName))
+            {
+                writer.WriteProperty(OpenApiConstants.ReadOnly, false, false);
```
PerthCharern: >         writer.WriteProperty(OpenApiConstants.ReadOnly, false, false); [](start = 7, length = 71)

Let's just not write anything since the default value according to the spec is already false.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -549,7 +582,15 @@ internal void WriteAsSchemaProperties(IOpenApiWriter writer)
...
+            }
+            else
+            {
+                writer.WriteProperty(OpenApiConstants.ReadOnly, ReadOnly, false);
```
PerthCharern: > false [](start = 73, length = 6)

nit, suggest named param for this last parameter

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -363,5 +417,65 @@ public void SerializeReferencedSchemaAsV3JsonWorks()
...
             actual.Should().Be(expected);
         }
+
+        [Fact]
+        public void SerializeSchemaWRequiredPropertiesAsV2JsonWorks()
+        {
+            // Arrange
+            var outputStringWriter = new StringWriter();
+            var writer = new OpenApiJsonWriter(outputStringWriter);
+            // Arrange
```
PerthCharern: >     // Arrange [](start = 7, length = 15)

nit, not needed twice

darrelmiller:
Sorry I didn't get chance to review this last week, but I have a question.  Why would we not do this as a validation rule?   Putting validation constraints in the writer seems like an odd place to put them.  A user may intentionally set a property as read-only and then be surprised when it doesn't get serialized.

It does bring up the interesting question that maybe we should run the validator before writing.

/cc @Shwetap05 @PerthCharern

PerthCharern:
This comes back to my original question of V2 vs V3 rule. In V3, the document with required and readonly is valid; one will simply take precedence over the other. In V2, it is invalid.

The library should still write the best document it can that complies with the spec for V2 serialization. I don’t think we should write something that we know is wrong.

The correct workflow for the consumer should still be call the validator first, but then this is a V2 rule only. We can create the rule for our customer but we can’t add it to the Default set.

darrelmiller:
@PerthCharern My apologies.  I didn't ready the V3 rules closely enough.  I was not involved in discussion around that rule change.  Preventing the V2 document from being invalid when writing seems like the best way of dealing with this.

PR #215, opened 2018-02-15
Fix Responses rule

### LINKED ISSUE #57: More specific check for the reading content
For example:

When we read the response, Can we make sure the response status codes is HTTP status code (3-digit-number starting with 1, 2, 3, 4, 5 or the range definition allowed: 1XX, 2XX, 3XX, 4XX, 5XX)

### LINKED ISSUE #214: Rule in OpenApiResponsesRules is too restrictive.
Spec (both V2 and V3):

The Responses Object MUST contain at least one response code, and it SHOULD be the response for a successful operation call.

Our rule:

Responses must contain success response

What I think we should do:

Just check that there is one response. We don't have the capability for the SHOULD type of rule right now.

PerthCharern:
Fix #214, Fix #57

### DIFF
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiResponsesRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiResponsesRules.cs
@@ -41,13 +41,15 @@ namespace Microsoft.OpenApi.Validations.Rules
                     {
                         context.Enter(key);
 
-                        if (key != "default" && !Regex.IsMatch(key, "^[1-5|X][0-9|X][0-9|X]$"))
+                        if (key != "default" && !Regex.IsMatch(key, "^[1-5]([0-9][0-9]|XX)$"))
                         {
                             context.AddError(
                                 new ValidationError(
                                     ErrorReason.Format,
                                     context.PathString,
-                                    "Responses key must be 'default' or an HTTP status code"));
+                                    "Responses key must be 'default', an HTTP status code, " +
+                                    "or one of the following strings representing a range of HTTP status codes: " +
+                                    "'1XX', '2XX', '3XX', '4XX', '5XX'"));
                         }
 
                         context.Exit();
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi/Validations/Rules/OpenApiResponsesRules.cs
```diff
@@ -14,16 +14,44 @@ namespace Microsoft.OpenApi.Validations.Rules
...
+        /// </summary>
+        public static ValidationRule<OpenApiResponses> ResponsesMustBeIdentifiedByDefaultOrStatusCode =>
+            new ValidationRule<OpenApiResponses>(
+                (context, responses) =>
+                {
+                    foreach (var key in responses.Keys)
+                    {
+                        context.Enter(key);
+
+                        if (key != "default" && !Regex.IsMatch(key, "^[1-5|X][0-9|X][0-9|X]$"))
```
PerthCharern: The spec is kinda ambiguous here on whether XXX is allowed. I can see that the intention is probably "not allowed" - since "default" is created for that purpose. However, the exact wording of the spec itself does not make it crystal clear which digits the wildcard X can be used in.

To define a range of response codes, this field MAY contain the uppercase wildcard character X. For example, 2XX represents all response codes between [200-299]. The following range definitions are allowed: 1XX, 2XX, 3XX, 4XX, and 5XX. If a response range is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code.

darrelmiller: I submitted a PR to the spec to clarify the wording.  Does [this](https://github.com/OAI/OpenAPI-Specification/pull/1489/files?short_path=18985dd#diff-18985dd9fd564e6d6dbe0c6cb887484d) help?

PerthCharern: Yup. That helps. I'll fix my code

PR #226, opened 2018-03-16
Fixed security requirement resolution 

### LINKED ISSUE #222: openapi, Security parse issue: openApiSecurityScheme.Flows will always be null.
for example, parse below content of security,  the flows will be always empty. And in the v1.0.0-beta012 it works, but in v1.0.0-beta014 it was broken.

```yaml
components:
  securitySchemes:
    azure_auth: 
      type: oauth2
      description: Azure Active Directory OAuth2 Flow.
      in: header
      flows: 
        implicit:
          authorizationUrl: https://example.com/api/oauth/dialog
          tokenUrl: https://example.com/token
          scopes:
            user_impersonation: modify pets in your account
            read: read your pets
        authorizationCode:
          authorizationUrl: https://example.com/api/oauth/dialogcode
          tokenUrl: https://example.com/api/oauth/tokencode
          scopes:
            user_impersonation: modify pets in your account code.
            read: read your pets code.
```

darrelmiller:
Fixes #222

### DIFF
--- a/src/Microsoft.OpenApi.Readers/Services/OpenApiReferenceResolver.cs
+++ b/src/Microsoft.OpenApi.Readers/Services/OpenApiReferenceResolver.cs
@@ -90,6 +90,24 @@ namespace Microsoft.OpenApi.Readers.Services
         }
 
         /// <summary>
+        /// Resolve all references to SecuritySchemes
+        /// </summary>
+        /// <param name="securityRequirement"></param>
+        public override void Visit(OpenApiSecurityRequirement securityRequirement)
+        {
+            foreach (var scheme in securityRequirement.Keys.ToList())
+            {
+                ResolveObject(scheme, (resolvedScheme) => {
+                    // If scheme was unresolved
+                    // copy Scopes and remove old unresolved scheme
+                    var scopes = securityRequirement[scheme];
+                    securityRequirement.Remove(scheme);
+                    securityRequirement.Add(resolvedScheme, scopes);
+                });
+            }
+        }
+
+        /// <summary>
         /// Resolve all references to links
         /// </summary>
         public override void Visit(IDictionary<string, OpenApiLink> links)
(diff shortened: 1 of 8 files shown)


src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -470,10 +470,32 @@ internal void Walk(OpenApiOperation operation)
             Walk(OpenApiConstants.Responses, () => Walk(operation.Responses));
             Walk(OpenApiConstants.Callbacks, () => Walk(operation.Callbacks));
             Walk(OpenApiConstants.Tags, () => Walk(operation.Tags));
-
+            Walk(OpenApiConstants.Security, () => Walk(operation.Security));
```
xuzhg: If we miss add the process code, it should not be failed. So, I can't understand why? Is there any default setting that we can fix it?

xuzhg:
Also, I'd like we should have some test cases to cover it.

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
darrelmiller: I'm not sure I follow.  If we are missing Walk() methods then there may be references that remain unresolved or objects not validated.  Technically that's not an error as Unresolved references is a valid state.  They could be resolved manually later.

Yes, there has been an unfortunate number of cases of missing resolved references.  In theory these should have been picked up by regular test cases as unresolved references have null/default property values..  Not quite sure why more were not.

src/Microsoft.OpenApi.Readers/Services/OpenApiReferenceResolver.cs
```diff
@@ -89,6 +89,23 @@ public override void Visit(OpenApiResponses responses)
             ResolveMap(responses);
         }
 
+        /// <summary>
+        /// Resolve all references to SecuritySchemes
```
xuzhg: the comments should be "SecurityRequirement" not "SecuritySchemes" ?

darrelmiller: Actually no.  SecurityRequirement is a dictionary whose keys are SecuritySchemes.  It is those keys that are being replaced if they are unresolved references.

Issue #225, opened 2018-03-14
V2 Seriazlizer: Map custom type for headers/path/query to string


Shwetap05:
In V3, for parameters headers/path/query a custom object can be used to define the schema but in V2 it must be one of "string", "number", "integer", "boolean", "array" or "file".

So when serializing a V3 document with custom type to V2, custom type should be mapped to string.

PR #231, opened 2018-03-19
V2 Seriazlizer: Map custom type for headers/path/query to string #225

Shwetap05:


src/Microsoft.OpenApi/Models/OpenApiParameter.cs
```diff
@@ -257,6 +257,12 @@ public void SerializeAsV2WithoutReference(IOpenApiWriter writer)
             {
                 writer.WriteOptionalObject(OpenApiConstants.Schema, Schema, (w, s) => s.SerializeAsV2(w));
             }
+            // In V2 parameter's type can't be a reference to a custom object schema or can't be of type object
+            // So in that case map the type as string.
+            else if (Schema?.Reference?.Type == ReferenceType.Schema || Schema?.Type == "object")
```
darrelmiller: The referenced schema might actually be a primitive schema, like string or integer.  I would change the condition to  `else if (Schema?.UnresolvedReference == false && Schema?.Type == "object") ` If someone wants to output the referenced schema as `type: string` they will have to do the resolution.

Shwetap05: ok will fix it.

---
In reply to: [175582906](https://github.com/Microsoft/OpenAPI.NET/pull/231#discussion_r175582906) [](ancestors = 175582906)

Shwetap05:
Thanks for fixing the tests

Issue #223, opened 2018-03-12
String reader diagnostic error when description is empty for a response


Shwetap05:
OpenApiStringReader.Read throw below diagnostic error when description is empty for a response:
"The field 'description' in 'response' object is REQUIRED."

Per Specification description is required, does it mean it should not be empty too? b/c the document loads fine in swagger editor without any schema failures.

Sample Json:

{
  "openapi": "3.0.1",
  "info": {
    "title": "test",
    "version": "1.0.0"
  },
  "servers": [
    {
      "url": "https://test.net"
    }
  ],
  "paths": {
    "/v1/applicationSettings": {
      "get": {
        "tags": [
          "ApplicationSettings"
        ],
        "summary": "Gets the settings of this application.",
        "operationId": "getV1ApplicationSettings",
        "responses": {
          "200": {
            "description": "",
            "content": {
              "application/json": {
                "schema": {
                  "type": "string"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
  }
}

darrelmiller:
If empty was allowed the requirement for it to be required would be pointless.

darrelmiller:
Currently the spec is interpreted as empty satisfies the requirement.  This is a topic for next week's meeting to decide if this needs to change for properties like `name`.  https://github.com/OAI/OpenAPI-Specification/issues/1516

scott-lin:
@darrelmiller, curious what the reasoning was to allow empty string for these fields when they say "Required" in the spec. As your original comment suggested, it's seemingly "pointless".

If this stands, the documentation is arguably ambiguous ([Response description for example](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#responseObject)) in my opinion.

darrelmiller:
@scott-lin  I have had a few conversations about this with the OpenAPI working group.   The spec isn't clear about what is considered necessary to satisfy the required condition.  However, considering JSON Schema considers the mere presence of a property as satisfying the required constraint, it can be safely assumed that some folks are relying on the fact that an empty description meets the condition.

[...]

Issue #201, opened 2018-01-31
Rule "name" should be saved in Validation Error


PerthCharern:
Currently validation error is just a group of strings that are loosely tied to the rule name itself. We should save the rule name itself into the error.

darrelmiller:
I'll address this when I take care of #170

Issue #170, opened 2018-01-04
Consolidate Exception and Error classes


darrelmiller:
We have an OpenApiException class in the Model project.  We have a OpenApiError class in the reader project for capturing syntax errors while reading the document and doing some validation that currently overlaps with the validator.  And finally we have ValidationError that has a type that distinguishes between Format errors and Required errors.

I'd like to suggest we clean this up a bit.  Perhaps by just creating exception classes derived from OpenApiException.

     - OpenApiException
       - OpenApiParsingException
       - OpenApiRequiredException
       - OpenApiFormatException

/cc @PerthCharern @xuzhg

PR #235, opened 2018-04-02
Consolidating error classes and adding more metadata to errors

darrelmiller:
Addresses #201 #170 #223

Also considering adding "Syntax" and "Structure" to the ErrorReason enum.  Syntax would be for invalid YAML/JSON.  Structure would be for things like when we find a map where there should be a list.


src/Microsoft.OpenApi/Models/OpenApiError.cs
```diff
@@ -28,6 +51,20 @@ public OpenApiError(string pointer, string message)
             Message = message;
         }
 
+        /// <summary>
+        /// Initializes the <see cref="OpenApiError"/> class.
+        /// </summary>
+        public OpenApiError(ErrorReason reason, string pointer, string message)
```
Shwetap05: >ErrorReason reaso [](start = 28, length = 17)

this is not being used.

Shwetap05:
>         public OpenApiError(OpenApiException exception)

this can be OpenApiError(OpenApiException exception) : this(exception.Message,exception.Pointer )

---
Refers to: src/Microsoft.OpenApi/Models/OpenApiError.cs:39 in 7c7a32a. [](commit_id = 7c7a32a3d6a182593ea70384a61ebc1f624f5374, deletion_comment = False)

src/Microsoft.OpenApi/Validations/Rules/OpenApiInfoRules.cs
```diff
@@ -16,28 +16,26 @@ public static class OpenApiInfoRules
...
-        public static ValidationRule<OpenApiInfo> FieldIsRequired =>
+        public static ValidationRule<OpenApiInfo> InfoRequiredFields =>
             new ValidationRule<OpenApiInfo>(
                 (context, item) =>
                 {
 
                     // title
                     context.Enter("title");
-                    if (String.IsNullOrEmpty(item.Title))
+                    if (item.Title == null)
```
Shwetap05: >m.Title == [](start = 27, length = 10)

so now required fields can be empty, right?

src/Microsoft.OpenApi/Models/OpenApiError.cs
```diff
@@ -3,13 +3,36 @@
...
 using Microsoft.OpenApi.Exceptions;
 
-namespace Microsoft.OpenApi.Readers
+namespace Microsoft.OpenApi.Models
 {
     /// <summary>
-    /// Error related to reading Open API YAML/JSON.
+    /// Error reason.
+    /// </summary>
+    public enum ErrorReason
```
Shwetap05: >ErrorReas [](start = 16, length = 9)

Is this needed asking b/c from a consumer perspective i am not able to figure out what do these mean? 

I looked at the exceptions themselves and they are descriptive enough to convey the problem e.g. EmailMustBeEmailFormat

darrelmiller: It is possible that now we are including the RuleName in the error, the reason is less useful.  However, the reason does allow some amount of statistical analysis on documents to understand the types of errors that people are making.  Arguably this information could be derived from well known rule names.  I'm on the fence.  Perhaps Sam has thoughts.  /cc @xuzhg

Shwetap05: Oh ok, i do not have strong opinion on removing it but feel it should be little bit more descriptive like IncorrectFormat and RequiredFieldMissing or something along those lines?

---
In reply to: [178636972](https://github.com/Microsoft/OpenAPI.NET/pull/235#discussion_r178636972) [](ancestors = 178636972)

src/Microsoft.OpenApi/Models/OpenApiError.cs
```diff
@@ -28,6 +51,20 @@ public OpenApiError(string pointer, string message)
...
         }
 
+        /// <summary>
+        /// Initializes the <see cref="OpenApiError"/> class.
+        /// </summary>
+        public OpenApiError(ErrorReason reason, string pointer, string message)
+        {
+            Pointer = pointer;
+            Message = message;
+        }
```
PerthCharern: Perhaps adding a constructor that takes rule name as well.

src/Microsoft.OpenApi/Models/OpenApiError.cs
PerthCharern: I agree with this part: "Arguably this information could be derived from well known rule names."

Having to specify this ErrorReason together with the rule name seems redundant, and as Shweta mentioned, it's a little difficult to understand what these reasons mean. I'd say we keep the statistical analysis concern separate --that can be done by postprocessing the rule name.

---
In reply to: [178638357](https://github.com/Microsoft/OpenAPI.NET/pull/235#discussion_r178638357) [](ancestors = 178638357,178636972)

PerthCharern: If we want to facilitate any analysis later, maybe we can build a map containing information about a rule  (rule -> ruleInformation) with ruleInformation containing things like these enums? Users can add info to that map for their custom rules.

---
In reply to: [178909044](https://github.com/Microsoft/OpenAPI.NET/pull/235#discussion_r178909044) [](ancestors = 178909044,178638357,178636972)

PerthCharern:
>                         }

use context.CreateError with name

---
Refers to: src/Microsoft.OpenApi/Validations/Rules/OpenApiResponsesRules.cs:50 in 7c7a32a. [](commit_id = 7c7a32a3d6a182593ea70384a61ebc1f624f5374, deletion_comment = False)

src/Microsoft.OpenApi/Validations/ValidationExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Validations
+{
+    /// <summary>
+    /// Helper methods to simplify creating validation rules
+    /// </summary>
+    public static class ValidationExtensions
```
PerthCharern: >ValidationExtensions [](start = 24, length = 20)

ValidationContextExtensions ?

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiDocumentTests.cs
```diff
@@ -108,6 +108,9 @@ public void ParseBrokenMinimalDocumentShouldYieldExpectedDiagnostic()
                         Errors =
                         {
                             new OpenApiError("#/info/title", "The field 'title' in 'info' object is REQUIRED.")
+                            {
+                                RuleName = "InfoRequiredFields"
```
PerthCharern: >InfoRequiredFields [](start = 44, length = 18)

nit, use nameof

test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs
```diff
@@ -55,10 +55,13 @@ public void ResponseMustHaveADescription()
...
 
             validator.Errors.ShouldBeEquivalentTo(
-                    new List<ValidationError>
+                    new List<OpenApiError>
                     {
-                        new ValidationError(ErrorReason.Required, "#/paths/~1test/get/responses/200/description",
+                        new OpenApiError(ErrorReason.Required, "#/paths/~1test/get/responses/200/description",
                             String.Format(SRResource.Validation_FieldIsRequired, "description", "response"))
+                        {
+                            RuleName = "ResponseRequiredFields"
```
PerthCharern: >ResponseRequiredFields [](start = 40, length = 22)

nit, use nameof

test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs
```diff
@@ -87,10 +90,13 @@ public void ServersShouldBeReferencedByIndex()
...
 
             validator.Errors.ShouldBeEquivalentTo(
-                    new List<ValidationError>
+                    new List<OpenApiError>
                     {
-                        new ValidationError(ErrorReason.Required, "#/servers/1/url",
+                        new OpenApiError(ErrorReason.Required, "#/servers/1/url",
                             String.Format(SRResource.Validation_FieldIsRequired, "url", "server"))
+                        {
+                            RuleName = "ServerRequiredFields"
```
PerthCharern: >ServerRequiredFields [](start = 40, length = 20)

nit, use nameof

src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
```diff
@@ -16,17 +16,16 @@ public static class OpenApiResponseRules
...
         /// Validate the field is required.
         /// </summary>
-        public static ValidationRule<OpenApiResponse> FieldIsRequired =>
+        public static ValidationRule<OpenApiResponse> ResponseRequiredFields =>
             new ValidationRule<OpenApiResponse>(
                 (context, response) =>
                 {
                     // description
                     context.Enter("description");
                     if (String.IsNullOrEmpty(response.Description))
```
PerthCharern: >String.IsNullOrEmpty(response.Description)) [](start = 24, length = 43)

Based on your response in [https://github.com/Microsoft/OpenAPI.NET/issues/223](https://github.com/Microsoft/OpenAPI.NET/issues/223), this should only check for null at the moment, right?

PerthCharern:
I think there's still some value in designating whether an error is from a validation or from general reading process. I think your "ErrorReason" field might fit this purpose more - with the values like "Validation", "Reading", etc. -- Currently, it might be confusing for consumers of the OpenApiDiagnostic object since they don't know where the OpenApiError actually comes from.

src/Microsoft.OpenApi/Validations/Rules/OpenApiResponseRules.cs
darrelmiller: Yeah, I hadn't fixed all the cases.  I should have before pushing the PR, but I ran out of time and wanted to get feedback on the overall set of changes.

darrelmiller:
@PerthCharern   Where the problem was detected is different than the "class" of problem, although they may be related.  E.g. The "structure" and "syntax" errors would be detected at reading time.
I'd really like to know of a good use case for collecting this metadata before debating it for too long.

src/Microsoft.OpenApi/Models/OpenApiError.cs
darrelmiller: I'll remove it.

src/Microsoft.OpenApi/Validations/Rules/OpenApiInfoRules.cs
darrelmiller: Yes, but I have a few more places where it needs to be fixed.

PerthCharern:
Thinking out loud here: 

Maybe it's fine not to consolidate them. Validation and Reading can be thought of as separate processes with their own error objects. Validation error objects need RuleName, but that concept doesn't really apply to Reading error. Adding RuleName to a generic error doesn't seem right given that it will not be populated in a lot of cases (in case we have more processor error types in the future)

What do you think about this?

OpenApiError class as abstract base class. This contains Message and Pointer.

OpenApiReaderError class derives from the base class. Doesn't contain extra field.

OpenApiValidatorError class derives from the base class. Contains RuleName as extra field.

The OpenApiDiagnostic object contains more than one Errors list. One per each class of OpenApiError.

---
In reply to: [378343191](https://github.com/Microsoft/OpenAPI.NET/pull/235#issuecomment-378343191) [](ancestors = 378343191)

PerthCharern:
It seems to me that ReadingErrors seem more "severe" in general - if we get a ReadingError, we can't really be sure if the object that was read in correctly represents what the user wants. ValidationErrors say something's wrong with the doc, but at least the doc object is assumed to be the correct representation.

I think it's nice to know which processor the error originates from. Later on, we might have more than these two processors.

And also, see my other comment regarding each error having different metadata.

---
In reply to: [378344518](https://github.com/Microsoft/OpenAPI.NET/pull/235#issuecomment-378344518) [](ancestors = 378344518)

darrelmiller:
That makes a lot of sense.  I like that.

---
In reply to: [378344812](https://github.com/Microsoft/OpenAPI.NET/pull/235#issuecomment-378344812) [](ancestors = 378344812,378343191)

darrelmiller:
I created the OpenApiReaderError and OpenApiValidatorError, but didn't split the error collections out on Diagnostic.  If folks want to filter they can do that by type.

---
In reply to: [378346984](https://github.com/Microsoft/OpenAPI.NET/pull/235#issuecomment-378346984) [](ancestors = 378346984,378344518)

src/Microsoft.OpenApi/Models/OpenApiError.cs
```diff
@@ -3,17 +3,17 @@
...
     public class OpenApiError
     {
         /// <summary>
         /// Initializes the <see cref="OpenApiError"/> class using the message and pointer from the given exception.
         /// </summary>
-        public OpenApiError(OpenApiException exception)
+        public OpenApiError(OpenApiException exception) : this(exception.Pointer, exception.Message)
         {
             Message = exception.Message;
             Pointer = exception.Pointer;
```
Shwetap05: this is not required anymore

src/Microsoft.OpenApi/Validations/OpenApiValidatorError.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public class OpenApiValidatorError : OpenApiError
+    {
+        /// <summary>
+        /// Initializes the <see cref="OpenApiError"/> class.
+        /// </summary>
+        public OpenApiValidatorError(string ruleName, string pointer, string message) : base(pointer, message)
+        {
+            Pointer = pointer;
+            Message = message;
```
Shwetap05: this will get assigned in base class, so remove it.

src/Microsoft.OpenApi/Validations/OpenApiValidatorError.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Validations
+{
+    /// <summary>
+    /// Errors detected when validating a OpenAPI Element
```
Shwetap05: >a [](start = 40, length = 1)

nit. an

src/Microsoft.OpenApi/Extensions/OpenApiElementExtensions.cs
```diff
@@ -19,7 +20,7 @@ public static class OpenApiElementExtensions
         /// <param name="element">Element to validate</param>
         /// <param name="ruleSet">Optional set of rules to use for validation</param>
         /// <returns>An IEnumerable of errors.  This function will never return null.</returns>
-        public static IEnumerable<ValidationError> Validate(this IOpenApiElement element, ValidationRuleSet ruleSet = null)         {
+        public static IEnumerable<OpenApiError> Validate(this IOpenApiElement element, ValidationRuleSet ruleSet = null)         {
```
Shwetap05: >         { [](start = 120, length = 10)

nit, put it on new line, will help with readability.

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.OpenApi.Validations
     public class OpenApiValidator : OpenApiVisitorBase, IValidationContext 
     {
         private readonly ValidationRuleSet _ruleSet;
-        private readonly IList<ValidationError> _errors = new List<ValidationError>();
+        private readonly IList<OpenApiError> _errors = new List<OpenApiError>();
```
Shwetap05: >OpenApiError [](start = 31, length = 12)

this should be OpenApiValidatorError right?

src/Microsoft.OpenApi/Validations/Rules/OpenApiLicenseRules.cs
```diff
@@ -16,16 +16,15 @@ public static class OpenApiLicenseRules
         /// <summary>
         /// REQUIRED.
         /// </summary>
-        public static ValidationRule<OpenApiLicense> FieldIsRequired =>
+        public static ValidationRule<OpenApiLicense> LicenseRequiredFields =>
```
Shwetap05: >LicenseRequiredF [](start = 53, length = 16)

feel this should be LicenseIsRequired? or LicenseFieldIsRequired

src/Microsoft.OpenApi/Validations/Rules/OpenApiOAuthFlowRules.cs
```diff
@@ -16,37 +16,34 @@ public static class OpenApiOAuthFlowRules
         /// <summary>
         /// Validate the field is required.
         /// </summary>
-        public static ValidationRule<OpenApiOAuthFlow> FieldIsRequired =>
+        public static ValidationRule<OpenApiOAuthFlow> OAuthFlowRequiredFields =>
```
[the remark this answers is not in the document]
darrelmiller: The way the rules were setup is that we validate all the "RequiredFields" in an object in a single rule.  I'm open to renaming/reorganizing these, but I don't want to spend to much time on naming these things.

---
In reply to: [179300300](https://github.com/Microsoft/OpenAPI.NET/pull/235#discussion_r179300300) [](ancestors = 179300300)

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
```diff
@@ -30,7 +30,7 @@ public OpenApiValidator(ValidationRuleSet ruleSet = null)
         /// <summary>
         /// Gets the validation errors.
         /// </summary>
-        public IEnumerable<ValidationError> Errors
+        public IEnumerable<OpenApiError> Errors
```
Shwetap05: OpenApiValidatorError

src/Microsoft.OpenApi/Models/OpenApiError.cs
```diff
@@ -3,20 +3,18 @@
...
 using Microsoft.OpenApi.Exceptions;
 
-namespace Microsoft.OpenApi.Readers
+namespace Microsoft.OpenApi.Models
 {
     /// <summary>
-    /// Error related to reading Open API YAML/JSON.
+    /// Error related to the Open API Document.
     /// </summary>
     public class OpenApiError
```
PerthCharern: >public [](start = 4, length = 6)

I think this class should be abstract, so that it can't be instantiated.

PerthCharern: OpenApiError should be abstract, so that it can't be instantiated.

PerthCharern:
Sounds good to me!

---
In reply to: [378612779](https://github.com/Microsoft/OpenAPI.NET/pull/235#issuecomment-378612779) [](ancestors = 378612779,378346984,378344518)

test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs
```diff
@@ -106,7 +107,7 @@ public void ValidateCustomExtension()
                  {
                      if (item.Bar == "hey")
                      {
-                         context.AddError(new ValidationError(ErrorReason.Format, context.PathString, "Don't say hey"));
+                         context.AddError(new OpenApiValidatorError("FooExtensionRule", context.PathString, "Don't say hey"));
```
PerthCharern: >"FooExtensionRule" [](start = 68, length = 18)

nit, use nameof

src/Microsoft.OpenApi/Models/OpenApiError.cs
darrelmiller: @PerthCharern  I have to move OpenApiReferenceResolver in an upcoming PR to the OpenAPI Model project and it raises errors.  I was considering just creating a OpenApiError for that, but maybe I will create an OpenApiReferenceError.  However, if we make OpenApiError abstract I have to consider whether it should just be an interface.... decisions, decisions.

PerthCharern: Making this abstract signals that each processor should have their own error classes, and it's easy to filter out errors of the same type from the OpenApiDiagnostic. 

I would argue against making this an interface. The common constructor that you have now makes sense - it can be leveraged by children classes.

---
In reply to: [179584501](https://github.com/Microsoft/OpenAPI.NET/pull/235#discussion_r179584501) [](ancestors = 179584501)

PR #236, opened 2018-04-04
Fixes for issue 233

darrelmiller:
See #233 

- Schema default is now parses as OpenApiAny instead of OpenApiString and so preserves typing information
- V2 Header objects now render schema properties correctly
- Casing of enum values has been corrected
- collectionFormat is now serialized out in V2
- Response bodies that have no schema are now rendered out and therefore correctly generate produces array in V2
- externalDocs in Tags were not being parsed in either V2 or V3. This has been fixed.

The diff for the round-tripped Petstore V2 JSON document now has 12 differences.  They include not rendering `required: false` because that is the default value and we don't render out default values for optional properties.   Also, the input document has some empty parameter arrays.  These are not rendered in the output.  The final difference is that there are some request bodies that have a schema but no consumes.  This is not possible to represent in OpenAPI V3.  To work around this, we create a content object of type `application/json`.  This ends up generating a consumes property in the output that isn't present in the input.  This is the only semantic difference between the input and output versions and one that I believe is tolerable.

/cc @scott-lin


src/Microsoft.OpenApi/Models/SecuritySchemeType.cs
```diff
@@ -29,5 +29,6 @@ public enum SecuritySchemeType
         /// Use OAuth2 with OpenId Connect URL to discover OAuth2 configuration value.
         /// </summary>
         [Display("openIdConnect")] OpenIdConnect
+
     }
```
Shwetap05: nit remove empty line

src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
```diff
@@ -154,9 +154,15 @@ private static void LoadStyle(OpenApiParameter p, string v)
         {
             switch (v)
             {
-                // TODO: Handle "csv" for query / form parameter. The style should be Form, not Simple.
                 case "csv":
-                    p.Style = ParameterStyle.Simple;
+                    if (p.In == ParameterLocation.Query)
```
PerthCharern: >Query [](start = 50, length = 5)

Cookie as well

src/Microsoft.OpenApi/Models/OpenApiParameter.cs
```diff
@@ -280,8 +280,25 @@ public void SerializeAsV2WithoutReference(IOpenApiWriter writer)
...
+                {
+                    if (this.Style == ParameterStyle.Form && this.Explode == true)
+                    {
+                        writer.WriteProperty("collectionFormat", "multi");
+                    }
+                    else if (this.Style == ParameterStyle.PipeDelimited)
+                    {
+                        writer.WriteProperty("collectionFormat", "pipes");
+                    }
+                    else if (this.Style == ParameterStyle.PipeDelimited)
```
PerthCharern: >PipeDelimited [](start = 58, length = 13)

SpaceDelimited

src/Microsoft.OpenApi.Readers/V3/OpenApiTagDeserializer.cs
```diff
@@ -12,27 +13,45 @@ namespace Microsoft.OpenApi.Readers.V3
...
+        {
+            {s => s.StartsWith("x-"), (o, p, n) => o.AddExtension(p, n.CreateAny())}
+        };
+
         public static OpenApiTag LoadTag(ParseNode n)
         {
             var mapNode = n.CheckMapNode("tag");
 
-            var obj = new OpenApiTag();
+            var domainObject = new OpenApiTag();
```
PerthCharern: >domainObject [](start = 16, length = 12)

Why is this called domainObject?

PerthCharern:
@darrelmiller Darrel. I see that you already merged the commit but there are two issues above (1. Cookie 2. SpaceDelimited) which should be addressed for accuracy. Did you have a chance to take a look?

src/Microsoft.OpenApi/Models/OpenApiParameter.cs
darrelmiller: Fixed.

src/Microsoft.OpenApi.Readers/V3/OpenApiTagDeserializer.cs
darrelmiller: So that when I cut and paste the code from one load to the next, I don't have to rename it :-)

src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
PerthCharern: Makes sense. Totally forgot :)

PR #238, opened 2018-04-06
Updated package version to 1.0.0-beta016

darrelmiller:


PerthCharern:
We haven't released beta015 in the Releases page yet. Not sure if we still would like to maintain that page. I think it's a good practice to have a place to capture what has changed in each version. 

https://github.com/Microsoft/OpenAPI.NET/releases

darrelmiller:
oops.  i'll add notes.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Issue #243, opened 2018-04-25
Type dateTime not supported


TheEres:
Hi,

I found a problem with dateTime type.

Steps:

* Try to validate/serialize an OpenApi definition containing `dateTime` schema (e.g. [Docker Engine API](https://docs.docker.com/engine/api/v1.37/swagger.yaml)).

Result:
* `OpenApiWriterException: The type 'System.DateTimeOffset' is not supported in Open API document.
`

Proposed solution:
* Add support to DateTimeOffset, ISO 8601 style : `Writer.Write(value.ToString(@"O"))`.

darrelmiller:
Thanks for the feedback.  I will investigate.

darrelmiller:
I was able to reproduce and resolve.  I need to add some tests around this, but I think we should be able to include this our 1.0 release.

PR #246, opened 2018-04-27
Fix for writing DateTimeOffset values

darrelmiller:
Addresses #243


darrelmiller:
@PerthCharern Any idea why we have this license/cla check that is holding up the PR?

PerthCharern:
@darrelmiller 

It's required per Microsoft policy. The check evaluates whether a CLA needs to be signed (based on various factors like contributor(s), size of the change, etc.)

I have seen cases where it hangs up indefinitely for some reason. In that case, the fastest way to resolve it is to commit a new whitespace change.

darrelmiller:
@PerthCharern I knew what it was, I just couldn't understand why it was an issue.  I'll try the whitespace trick.

PerthCharern:
That I don't know. It started happening recently. I'll get in touch with the OSS team.

PR #254, opened 2018-05-10
Regex changes to allow square brackets in key

### LINKED ISSUE #253: OpenApiDiagnostic error when key contains square brackets for components
If a V2 json generated using swashbuckle contains square brackets in the key of the #definitions, OpenApiDiagnostics throw below error. Is there a reason why square brackets are not allowed? This validation is blocking one of the internal tool that consumes OpenApi.Net

Message:"The key 'Core.Collections.ResourceCollection[Services.Products.V2.Models.Availability]' in 'schemas' of components MUST match the regular expression '^[a-zA-Z0-9\\.\\-_]+$'."
Pointer: "#/components"
RuleName: "KeyMustBeRegularExpression"

Sample Json which is failing OpenApi.NET validation but renders fine in swagger editor:

{
  "swagger": "2.0",
  "info": {
    "version": "v1",
    "title": "TestService",
    "description": "Test service"
  },
  "host": "test.net",
  "basePath": "/test/WebApiV2",
  "schemes": [ "https" ],
  "paths": {
    "/v1/products/{productId}/skus/{skuId}/availabilities": {
      "get": {
        "tags": [ "Availability" ],
        "operationId": "Get /v1/products/{productId}/skus/{skuId}/availabilities?country={country_value}&targetSegment={targetSegment_value}",
        "consumes": [],
        "produces": [ "application/json", "text/json", "application/xml", "text/xml" ],
        "parameters": [
          {
            "name": "productId",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "skuId",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "country",
            "in": "query",
            "required": false,
            "type": "string"
          },
          {
            "name": "targetSegment",
            "in": "query",
            "required": false,
            "type": "string"
          },
          {
            "name": "Authorization",
            "in": "header",
            "description": "Authorization Bearer token",
            "required": true,
            "type": "string",
            "default": "Bearer "
          }
        ],
        "responses": {
          "200": {
            "description": "The request was successful. The availabilities are returned in the response body.",
            "schema": { "$ref": "#/definitions/Core.Collections.ResourceCollection[Services.Products.V2.Models.Availability]" }
          }
        }
      }
    }
  },
  "definitions": {
    "Core.Collections.ResourceCollection[Services.Products.V2.Models.Availability]": {
      "type": "object",
      "properties": {
        "totalCount": {
          "format": "int32",
          "type": "integer",
          "readOnly": true
        }
      }
    }
  },
  "securityDefinitions": {
    "Authorization": {
      "type": "apiKey",
      "description": "Authorization Bearer token",
      "name": "Authorization",
      "in": "header"
    }
  }
}

Shwetap05:
This is to fix #253

### DIFF
--- a/src/Microsoft.OpenApi/Validations/Rules/OpenApiComponentsRules.cs
+++ b/src/Microsoft.OpenApi/Validations/Rules/OpenApiComponentsRules.cs
@@ -17,11 +17,11 @@ namespace Microsoft.OpenApi.Validations.Rules
         /// <summary>
         /// The key regex.
         /// </summary>
-        public static Regex KeyRegex = new Regex(@"^[a-zA-Z0-9\.\-_]+$");
+        public static Regex KeyRegex = new Regex(@"^[a-zA-Z0-9\.\-_\[\]]+$");
 
         /// <summary>
         /// All the fixed fields declared above are objects
-        /// that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.
+        /// that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_\[\]]+$.
         /// </summary>
         public static ValidationRule<OpenApiComponents> KeyMustBeRegularExpression =>
             new ValidationRule<OpenApiComponents>(
--- a/src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
+++ b/src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
@@ -10,7 +10,7 @@
         <Company>Microsoft</Company>
         <Title>Microsoft.OpenApi</Title>
         <PackageId>Microsoft.OpenApi</PackageId>
-        <Version>1.0.0</Version>
+        <Version>1.0.1</Version>
         <Description>.NET models with JSON and YAML writers for OpenAPI specification</Description>
         <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
         <PackageTags>OpenAPI .NET</PackageTags>
--- a/test/Microsoft.OpenApi.Tests/Validations/OpenApiComponentsValidationTests.cs
+++ b/test/Microsoft.OpenApi.Tests/Validations/OpenApiComponentsValidationTests.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using Microsoft.OpenApi.Extensions;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Properties;
-using Microsoft.OpenApi.Services;
 using Microsoft.OpenApi.Validations.Rules;
 using Xunit;
 
@@ -41,6 +40,30 @@ namespace Microsoft.OpenApi.Validations.Tests
             OpenApiError error = Assert.Single(errors);
             Assert.Equal(String.Format(SRResource.Validation_ComponentsKeyMustMatchRegularExpr, key, "responses", OpenApiComponentsRules.KeyRegex.ToString()),
                 error.Message);
+        }
+
+        [Fact]
+        public void ValidateKeyShouldAllowSquareBracketsInComponents()
+        {
+            // Arrange
+            const string key = "[abc]";
+
+            OpenApiComponents components = new OpenApiComponents()
+            {
+                Responses = new Dictionary<string, OpenApiResponse>
+                {
+                    { key, new OpenApiResponse { Description = "any" } }
+                }
+            };
+
+            var errors = components.Validate( ValidationRuleSet.GetDefaultRuleSet() );
+
+            // Act
+            bool result = errors.Any();
+
+
+            // Assert
+            Assert.False( result );
         }
     }
 }


src/Microsoft.OpenApi/Validations/Rules/OpenApiComponentsRules.cs
```diff
@@ -17,11 +17,11 @@ public static class OpenApiComponentsRules
...
-        public static Regex KeyRegex = new Regex(@"^[a-zA-Z0-9\.\-_]+$");
+        public static Regex KeyRegex = new Regex(@"^[a-zA-Z0-9\.\-_\[\]]+$");
 
         /// <summary>
         /// All the fixed fields declared above are objects
-        /// that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.
+        /// that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_\[\]]+$.
         /// </summary>
         public static ValidationRule<OpenApiComponents> KeyMustBeRegularExpression =>
             new ValidationRule<OpenApiComponents>(
```
PerthCharern: The existing RegEx is a valid rule in V3.

For now, we should fix it in a different way. Create a new rule in your project - and when you call the reader, specify a ruleset with this new rule instead of the default V3 rule.

There was a discussion earlier about creating rulesets for V2 and V3, but I think we decided against it at that point. It seems like this is creating more work for our customers though @darrelmiller ?

darrelmiller: I think we should translate invalid characters into some valid character.  e.g. into a '-'.  Having square brackets in a component name would break JSON pointer references as square brackets have a specific meaning in a JSON pointer.

PerthCharern: @darrelmiller I thought the only special characters in JSON pointer are '/' and '~'. Regardless, I think the rule here should stay the same since that's what's required in V3. I would be a bit cautious replacing characters directly since that might cause unexpected issues for our customers - throwing validation error seems more predictable to me.

Regardless, we need to allow people who's using the DOM for their V2 document to not get blocked by this rule. Currently they can manually remove the rule, but that seems like an unnecessary and a bit obscure extra work. What do you think about creating a pre-defined rulesets for V2 / V3 / Vx that the customers can choose from?

---
In reply to: [187430025](https://github.com/Microsoft/OpenAPI.NET/pull/254#discussion_r187430025) [](ancestors = 187430025)

darrelmiller: @PerthCharern You are right about the special characters.  I always get mixed up about how you reference array elements with JSON Pointer.  
I understand the concern about changing characters, but I'm concerned about allowing V2 documents to be imported and then exported as V3 but with invalid content.  The DOM is a V3 object model.  It is going to make changes to V2 imported documents.  It changes how produces/consumes works.  Round tripping V2 to V2 wasn't a primary goal.  Upgrading V2 to V3 and downgrading V3 to V2 was.  Roundtripping V2 was a bit of an accidental benefit.

Customers understand that when translating from one version to the other there is going to be possible changes to the doc.  If we create V2 ruleset, we are implying that we "support V2 documents" and I think that is a bit of a stretch.

Shwetap05: Darrel, then should openapi.net replace all unsupported chars? or its going to be selective of replacing the one's that are supported in V2 but not in V3?

---
In reply to: [187445907](https://github.com/Microsoft/OpenAPI.NET/pull/254#discussion_r187445907) [](ancestors = 187445907)

darrelmiller: @Shwetap05 That's an interesting question.  I think we should only replace chars during the deserialization of V2 documents.  V2 had no limit on the allowed characters, so all unsupported characters should be changed.
However, if someone is importing a V3 document, we should not replace unsupported characters, we should create an error.
This is because we want to be explicit that we are "converting" a V2 to a V3 when we import so some transformation is expected.  If someone is importing a V3, we shouldn't change anything.

PerthCharern: @darrelmiller @Shwetap05  I agree with Darrel's approach above.

PerthCharern: To be clear, I think this is how that approach translates to the implementation:

- keep this rule the same (This will ensure that the final DOM object still has the correct set of characters.)

- convert unsupported characters in V2 deserializer.

src/Microsoft.OpenApi/Validations/Rules/OpenApiComponentsRules.cs
```diff
@@ -17,11 +17,11 @@ public static class OpenApiComponentsRules
...
         /// </summary>
-        public static Regex KeyRegex = new Regex(@"^[a-zA-Z0-9\.\-_]+$");
+        public static Regex KeyRegex = new Regex(@"^[a-zA-Z0-9\.\-_\[\]]+$");
 
         /// <summary>
         /// All the fixed fields declared above are objects
-        /// that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.
+        /// that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_\[\]]+$.
         /// </summary>
         public static ValidationRule<OpenApiComponents> KeyMustBeRegularExpression =>
```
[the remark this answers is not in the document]
PerthCharern: I think we do now.

Shwetap05:
Closing this pull request as we decided not to do this.

Issue #256, opened 2018-05-14
Add the ability to read OpenAPI description fragments


darrelmiller:
In the work that I am doing to support Swashbuckle and Azure API Management I have run into scenarios where I need to parse a string of OAS compatible text into a some IOpenAPIElement.  For Examples I need to create an IOpenApiAny instance.  For schema fragments I need to create an OpenApiSchema from a string.
I am working on PR and will link to this issue when I push it.

Currently I am attempting to create a new method on OpenApiStreamReader that looks like:

```
       public T ReadFragment<T>(Stream input, OpenApiSpecVersion version, out OpenApiDiagnostic diagnostic) 
                   where T : IOpenApiElement, new()
        { … }
      
```

PR #258, opened 2018-05-15
Read fragments of OpenAPI Specifications

darrelmiller:
See issue #256

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -90,11 +90,14 @@ namespace Microsoft.OpenApi.Readers
             }
 
             // Validate the document
-            var errors = document.Validate(_settings.RuleSet);
-            foreach (var item in errors)
+            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count > 0)
             {
-                diagnostic.Errors.Add(item);
-            } 
+                var errors = document.Validate(_settings.RuleSet);
+                foreach (var item in errors)
+                {
+                    diagnostic.Errors.Add(item);
+                }
+            }
 
             return document;
         }
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -144,6 +144,16 @@ namespace Microsoft.OpenApi.Readers
                 diagnostic.Errors.Add(new OpenApiError(ex));
             }
 
+            // Validate the element
+            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count > 0)
+            {
+                var errors = element.Validate(_settings.RuleSet);
+                foreach (var item in errors)
+                {
+                    diagnostic.Errors.Add(item);
+                }
+            }
+
             return (T)element;
         }
 
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using Microsoft.OpenApi.Exceptions;
 using Microsoft.OpenApi.Extensions;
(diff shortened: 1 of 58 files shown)


src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -102,6 +103,51 @@ public OpenApiDocument Read(Stream input, out OpenApiDiagnostic diagnostic)
             return document;
         }
 
+        /// <summary>
+        /// Reads the stream input and parses the fragment of an OpenAPI descinto an Open API Element.
```
PerthCharern: >descinto [](start = 73, length = 8)

Typo. - I'm actually not sure which word this is supposed to be a typo of :)

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -102,6 +103,51 @@ public OpenApiDocument Read(Stream input, out OpenApiDiagnostic diagnostic)
...
+            try
+            {
+                // Parse the OpenAPI element
+                element = context.ParseFragment<T>(yamlDocument, version, diagnostic);
+            }
+            catch (OpenApiException ex)
+            {
+                diagnostic.Errors.Add(new OpenApiError(ex));
+            }
+
```
PerthCharern: Can we also validate the fragment after parsing?

src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -64,6 +64,34 @@ internal OpenApiDocument Parse(YamlDocument yamlDocument, OpenApiDiagnostic diag
...
+
+            switch (version)
+            {
+                case OpenApiSpecVersion.OpenApi2_0:
+                    VersionService = new OpenApiV2VersionService();
+                    element = this.VersionService.LoadElement<T>(node);
+                    break;
+
+                case OpenApiSpecVersion.OpenApi3_0:
+                    this.VersionService = new OpenApiV3VersionService();
```
PerthCharern: >this. [](start = 20, length = 5)

nit, remove these "this." for consistency.

src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -64,6 +64,34 @@ internal OpenApiDocument Parse(YamlDocument yamlDocument, OpenApiDiagnostic diag
             return doc;
         }
 
+        /// <summary>
+        /// Initiates the parsing process of a fragment.  Not thread safe and should only be called once on a parsing context
+        /// </summary>
+        /// <param name="yamlDocument"></param>
+        /// <param name="version">OpenAPI version of the fragment</param>
+        /// <param name="diagnostic"></param>
```
PerthCharern: nit, either remove or populate

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
darrelmiller: @PerthCharern I'm not sure.  I didn't think we could because of references, but as they are unresolved that shouldn't be a problem.  Now I'm not sure.  Will think about it some more.

test/Microsoft.OpenApi.SmokeTests/ApiGurus.cs
```diff
@@ -27,6 +27,7 @@ public ApisGuruTests(ITestOutputHelper output)
 
         static ApisGuruTests()
         {
+            System.Net.ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
             _httpClient = new HttpClient(new HttpClientHandler()
```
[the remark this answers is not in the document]
darrelmiller: Apparently HttpClient defaults to an earlier version of TLS that apis.guru no longer supports.  This static option is just a way of telling HttpClient to use a new version of TLS.

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
darrelmiller: Added it.  Seems to be fine.

PR #261, opened 2018-05-17
Fix header deserialization in V2 documents

darrelmiller:
Several fields were not supported and some were supported that don't exist in V2.  I also re-ordered the parsing code to match the spec order.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiHeaderDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiHeaderDeserializer.cs
@@ -166,10 +166,6 @@ namespace Microsoft.OpenApi.Readers.V2
                     return;
                 case "tsv":
                     throw new NotSupportedException();
-                case "multi":
-                    h.Style = ParameterStyle.Form;
-                    h.Explode = true;
-                    return;
             }
         }
     }
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiHeaderDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiHeaderDeserializer.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
+using System;
+using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Extensions;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Readers.ParseNodes;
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi.Readers/V2/OpenApiHeaderDeserializer.cs
```diff
@@ -82,5 +150,27 @@ public static OpenApiHeader LoadHeader(ParseNode node)
...
+                    return;
+                case "pipes":
+                    h.Style = ParameterStyle.PipeDelimited;
+                    return;
+                case "tsv":
+                    throw new NotSupportedException();
+                case "multi":
+                    h.Style = ParameterStyle.Form;
+                    h.Explode = true;
+                    return;
```
PerthCharern: "multi" is now allowed in the header in V2 spec. Looks like it's only allowed in Param.

PerthCharern:
> collectionFormat: csv

Can we add enum here as well? It's a bit different than other fields and is worth testing.

---
Refers to: test/Microsoft.OpenApi.Readers.Tests/V2Tests/Samples/OpenApiParameter/headerParameter.yaml:11 in bed9b57. [](commit_id = bed9b5721909c4e4a8a50ee88caefe7c81b6cd07, deletion_comment = False)

src/Microsoft.OpenApi.Readers/V2/OpenApiHeaderDeserializer.cs
```diff
@@ -82,5 +150,23 @@ public static OpenApiHeader LoadHeader(ParseNode node)
 
             return header;
         }
+
+        private static void LoadStyle(OpenApiHeader h, string v)
```
PerthCharern: can we use full word here? header and format instead of h and v.

src/Microsoft.OpenApi.Readers/V2/OpenApiHeaderDeserializer.cs
```diff
@@ -82,5 +150,23 @@ public static OpenApiHeader LoadHeader(ParseNode node)
...
+                    return;
+                case "ssv":
+                    h.Style = ParameterStyle.SpaceDelimited;
+                    return;
+                case "pipes":
+                    h.Style = ParameterStyle.PipeDelimited;
+                    return;
+                case "tsv":
+                    throw new NotSupportedException();
+            }
```
PerthCharern: >   } [](start = 9, length = 4)

Looks like we should throw an error for case default? That would indicate that the document is broken in V2.

PR #264, opened 2018-05-24
Added parsing externalValue to Examples Object

darrelmiller:
Added missing property in Examples Object.  This is a quick fix for the Docs team.  In 1.1 we will add support for reading external files.


AP-G:
Hi. Do you have any updates on this? I still get an error `Data and type mismatch found. at .../examples/.../value` when using `externalValue` instead of `value`.

PR #267, opened 2018-06-05
A variety of bug fixes for 1.1 preview

darrelmiller:
- Merged fixes from 1.0 release into 1.1
- Added validation to 
- Made Paths null by default to fix error reporting
- Do not include default basePath in output. 
- Handle V2 import when produces comes after responses 
- Fix MakeServers for V2 
- Add settings parameter for BaseUrl
- Add support for null to GetJsonCompatibleString 
- Output multiple examples in V2 
- Exposed Settings in IOpenApiWriter interface to allow access to selected version when writing extensions. (Potentially a breaking change?)

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
@@ -49,5 +49,9 @@ namespace Microsoft.OpenApi.Readers
         /// </summary>
         public ValidationRuleSet RuleSet { get; set; } = ValidationRuleSet.GetDefaultRuleSet();
 
+        /// <summary>
+        /// URL where relative references should be resolved from if the description does not contain Server definitions
+        /// </summary>
+        public Uri BaseUrl { get; internal set; } = new Uri("https://example.org/");
     }
 }
--- a/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
@@ -52,6 +52,6 @@ namespace Microsoft.OpenApi.Readers
         /// <summary>
         /// URL where relative references should be resolved from if the description does not contain Server definitions
         /// </summary>
-        public Uri BaseUrl { get; internal set; } = new Uri("https://example.org/");
+        public Uri BaseUrl { get; set; } = new Uri("https://example.org/");
     }
 }
--- a/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
@@ -52,6 +52,6 @@ namespace Microsoft.OpenApi.Readers
         /// <summary>
         /// URL where relative references should be resolved from if the description does not contain Server definitions
         /// </summary>
-        public Uri BaseUrl { get; set; } = new Uri("https://example.org/");
+        public Uri BaseUrl { get; set; } 
     }
 }
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
@@ -136,15 +136,28 @@ namespace Microsoft.OpenApi.Readers.V2
             basePath = basePath ?? defaultUrl.GetComponents(UriComponents.Path, UriFormat.SafeUnescaped);
             schemes = schemes ?? new List<string> { defaultUrl.GetComponents(UriComponents.Scheme, UriFormat.SafeUnescaped) };
 
+            
             // Create the Server objects
             if (schemes != null)
             {
                 foreach (var scheme in schemes)
                 {
+                    var ub = new UriBuilder(scheme, host)
+                    {
+                        Path = basePath
+                    };
+
                     var server = new OpenApiServer
                     {
-                        Url = $"{scheme}://{host}{basePath}"
+                        Url = ub.ToString()
                     };
+
+                    // Server Urls are always appended to Paths and Paths must start with /
+                    // so removing the slash prevents a double slash.
+                    if (server.Url.EndsWith("/"))
+                    {
+                        server.Url = server.Url.Substring(0, server.Url.Length - 1);
+                    }
                     servers.Add(server);
                 }
             } 
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
@@ -137,7 +137,8 @@ namespace Microsoft.OpenApi.Readers.V2
                 host = host ?? defaultUrl.GetComponents(UriComponents.NormalizedHost, UriFormat.SafeUnescaped);
                 basePath = basePath ?? defaultUrl.GetComponents(UriComponents.Path, UriFormat.SafeUnescaped);
                 schemes = schemes ?? new List<string> { defaultUrl.GetComponents(UriComponents.Scheme, UriFormat.SafeUnescaped) };
-            } else if (String.IsNullOrEmpty(host) && String.IsNullOrEmpty(basePath))
+            }
+            else if (String.IsNullOrEmpty(host) && String.IsNullOrEmpty(basePath))
             {
                 return;  // Can't make a server object out of just a Scheme
             }
@@ -147,6 +148,11 @@ namespace Microsoft.OpenApi.Readers.V2
             {
                 foreach (var scheme in schemes)
                 {
+                    if (String.IsNullOrEmpty(scheme))
+                    {
+                        host = "//" + host;  // The double slash prefix creates a relative url where the scheme is defined by the BaseUrl
+                    }
+
                     var uriBuilder = new UriBuilder(scheme, host)
                     {
                         Path = basePath
@@ -164,7 +170,7 @@ namespace Microsoft.OpenApi.Readers.V2
             {
                 if (!String.IsNullOrEmpty(host))
                 {
-                    host = "//" + host;
+                    host = "//" + host;  // The double slash prefix creates a relative url where the scheme is defined by the BaseUrl
                 }
                 var uriBuilder = new UriBuilder()
                 {
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
@@ -132,35 +132,63 @@ namespace Microsoft.OpenApi.Readers.V2
             }
 
             // Fill in missing information based on the defaultUrl
-            host = host ?? defaultUrl.GetComponents(UriComponents.NormalizedHost, UriFormat.SafeUnescaped);
-            basePath = basePath ?? defaultUrl.GetComponents(UriComponents.Path, UriFormat.SafeUnescaped);
-            schemes = schemes ?? new List<string> { defaultUrl.GetComponents(UriComponents.Scheme, UriFormat.SafeUnescaped) };
-
+            if (defaultUrl != null)
+            {
+                host = host ?? defaultUrl.GetComponents(UriComponents.NormalizedHost, UriFormat.SafeUnescaped);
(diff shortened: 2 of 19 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -29,7 +29,7 @@ public class OpenApiDocument : IOpenApiSerializable, IOpenApiExtensible
         /// <summary>
         /// REQUIRED. The available paths and operations for the API.
         /// </summary>
-        public OpenApiPaths Paths { get; set; } = new OpenApiPaths();
```
PerthCharern: > = new OpenApiPaths(); [](start = 47, length = 22)

The reason we had this was because OpenApiPaths is a collection (a dictionary). Is it causing issue?

src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
```diff
@@ -49,5 +49,9 @@ public class OpenApiReaderSettings
         /// </summary>
         public ValidationRuleSet RuleSet { get; set; } = ValidationRuleSet.GetDefaultRuleSet();
 
+        /// <summary>
+        /// URL where relative references should be resolved from if the description does not contain Server definitions
+        /// </summary>
+        public Uri BaseUrl { get; set; } = new Uri("https://example.org/");
```
PerthCharern: > Uri("https://example.org/"); [](start = 46, length = 29)

The spec says this: 

[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields)

 An array of Server Objects, which provide connectivity information to a target server. If the servers property is not provided, or is an empty array, the default value would be a Server Object with a url value of /.

PerthCharern: Actually, that might be unrelated. Let me read through more, and I'll readdress this.

---
In reply to: [193214821](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193214821) [](ancestors = 193214821)

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -119,21 +119,48 @@ internal static partial class OpenApiV2Deserializer
...
+
+            
+            // Create the Server objects
             if (schemes != null)
             {
                 foreach (var scheme in schemes)
                 {
-                    var server = new OpenApiServer();
-                    server.Url = scheme + "://" + (host ?? "example.org/") + (basePath ?? "/");
+                    var ub = new UriBuilder(scheme, host)
```
PerthCharern: >ub [](start = 24, length = 2)

nit, uriBuilder

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -119,21 +119,48 @@ internal static partial class OpenApiV2Deserializer
...
             var schemes = context.GetFromTempStorage<List<string>>("schemes");
 
+            // If nothing is provided, don't create a server
+            if (host == null && basePath == null && schemes == null)
+            {
+                return;
+            }
+
+            // Fill in missing information based on the defaultUrl
+            host = host ?? defaultUrl.GetComponents(UriComponents.NormalizedHost, UriFormat.SafeUnescaped);
```
PerthCharern: >defaultUrl.GetComponents(UriComponents.NormalizedHost, UriFormat.SafeUnescaped); [](start = 27, length = 80)

V2 spec:

If the host is not included, the host serving the documentation is to be used (including the port). 

This is in line with V3 spec which defaults server URL to "/" with the condition that "This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the OpenAPI document is being served."


We have a few special cases to handle:

=> Host empty + schemes not empty =>  This is weird. I don't know how to represent this in V3. Maybe we can simply ignore the schemes since we don't really have a way to handle it.

=> Host empty + schemes empty + basePath not empty => Essentially, what this means is we use the Host serving the doc but with appended path. To represent this in V3, we simply need the basePath in Server.url

=> Host empty + schemes empty + basePath empty => You handled this. This means no need for Server object in V3.

=> Host not empty + schemes empty => We can just use the host and basePath without schemes in Server.url in V3? That still seems like a valid URL to me.

src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
```diff
@@ -131,7 +131,12 @@ internal static OpenApiOperation LoadOperation(ParseNode node)
                     operation.RequestBody = CreateFormBody(node.Context, formParameters);
                 }
             }
-            
+
+            foreach (var response in operation.Responses.Values)
+            {
+                ProcessProduces(response, node.Context);
```
PerthCharern: >    ProcessProduces(response, node.Context); [](start = 12, length = 44)

Isn't this handled in OpenApiResponseDeserializer.cs line 123?

src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
PerthCharern: OK. I finished looking through the whole PR. I don't think we should use an arbitrary URL here. 

I'd handle this inside the deserializer itself (see my other comment there about special cases). It seems a bit confusing to have a baseUrl in the reader/parsingContext, especially the one that's not given by the caller.

---
In reply to: [193215807](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193215807) [](ancestors = 193215807,193214821)

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
darrelmiller: Yes.  We don't raise provide a validation error when a user provides a document that doesn't have a paths property.  We used to validate this in the reader, but we since moved it into our Validator.

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
darrelmiller: Just so we are the same page, the BaseUrl in the Settings is where you put "the Host serving the doc".  I think what you are saying, is you don't want an arbitrary default for "host serving the doc" when one isn't provided.  Is that correct?

src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
darrelmiller: I had to put it in operation for the case when produces came after responses.  But I needed it also in responses for when responses are defined at the root level as a reusable thing.  So I made it work when you call it in both places.

src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
darrelmiller: I can leave this null by default and fix the MakeServer stuff, but we still need a way for the user to specify the URL where the doc came from.

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
PerthCharern: ah okay. Correct - it's a little odd to default to an arbitrary example.org. Also, the conversion is possible (except case 1&2) above even without the info on "host serving the doc", so I'm not sure if we really need it.

---
In reply to: [193228718](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193228718) [](ancestors = 193228718)

PerthCharern: Ah. I see that the baseUrl has to be there for reference resolving. In that case, I think we should leave it blank like you suggested, handle the MakeServers here, and use the property for Reference Resolving as planned.

---
In reply to: [193231179](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193231179) [](ancestors = 193231179,193228718)

src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
PerthCharern: Sounds good.

---
In reply to: [193230553](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193230553) [](ancestors = 193230553)

src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
PerthCharern: Tricky! Sounds good to me.

---
In reply to: [193229384](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193229384) [](ancestors = 193229384)

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -114,6 +114,42 @@ public static void Serialize<T>(this T element, IOpenApiWriter writer, OpenApiSp
...
+
+                case OpenApiSpecVersion.OpenApi2_0:
+                    element.SerializeAsV2(writer);
+                    break;
+
+                default:
+                    throw new OpenApiException(string.Format(SRResource.OpenApiSpecVersionNotSupported, writer.Settings.SpecVersion));
+            }
+
+            writer.Flush();
```
PerthCharern: Can we leverage the method above?


return element.Serialize(writer, writer.Settings.SpecVersion);

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -119,21 +119,76 @@ internal static partial class OpenApiV2Deserializer
...
+                    };
+
                     servers.Add(server);
                 }
             }
+            else
+            {
+                if (!String.IsNullOrEmpty(host))
+                {
+                    host = "//" + host;
```
PerthCharern: >"//" [](start = 27, length = 4)

Is this needed?

PerthCharern: What is the benefit of adding these two slashes?

---
In reply to: [193593020](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193593020) [](ancestors = 193593020)

PerthCharern: >= "//" + host; [](start = 25, length = 14)

nit, it wasn't obvious to me in the beginning that we need this "//" to denote "the current protocol". 

Could you add some comments to explain this?

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -119,21 +119,76 @@ internal static partial class OpenApiV2Deserializer
...
+                }
+                var uriBuilder = new UriBuilder()
+                {
+                    Scheme = null,
+                    Host = host,
```
PerthCharern: The way this works is kinda strange - I'd hope the UriBuilder can add that "//" for us when the Scheme is null. However, it seems like the default behavior is that it would add http instead - which is not what we want.

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
darrelmiller: Fixed. Thanks!

---
In reply to: [193587359](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193587359) [](ancestors = 193587359)

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
darrelmiller: Yeah, I was hoping UriBuilder would add the // but sadly it doesn't.

---
In reply to: [193594097](https://github.com/Microsoft/OpenAPI.NET/pull/267#discussion_r193594097) [](ancestors = 193594097)

PR #275, opened 2018-06-11
Fixed thrown exceptions 

darrelmiller:
Resolves issue #270

### DIFF
--- a/test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
+++ b/test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
@@ -1,8 +1,7 @@
-﻿using System;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.OpenApi.Exceptions;
 using Microsoft.OpenApi.Models;
@@ -28,18 +27,14 @@ paths:
         '200':
           description: ok
           schema:
-            $ref: '#/definition/does/notexist'
+            $ref: '#/defi888nition/does/notexist'
 ";
 
-            var reader = new OpenApiStringReader(new OpenApiReaderSettings() {
+            var reader = new OpenApiStringReader();
+
+            Assert.Throws<OpenApiReaderException>(() => { 
+                var doc = reader.Read(input, out var diagnostic);
             });
-
-            OpenApiDocument doc = null;
-
-            Assert.Throws<OpenApiReaderException>(() =>
-            doc = reader.Read(input, out var diagnostic));
-
-            Assert.Null(doc);
         }
 
 
@@ -62,16 +57,13 @@ paths:
             $ref: '#/definitions/doesnotexist'
 ";
 
-            var reader = new OpenApiStringReader(new OpenApiReaderSettings()
-            {
-            });
+            var reader = new OpenApiStringReader();
 
-            OpenApiDocument doc = null;
-
-            doc = reader.Read(input, out var diagnostic);
+            var doc = reader.Read(input, out var diagnostic);
 
             diagnostic.Errors.ShouldBeEquivalentTo(new List<OpenApiError> {
                 new OpenApiError( new OpenApiException("Invalid Reference identifier 'doesnotexist'.")) });
+            doc.Should().NotBeNull();
         }
     }
 }
--- a/test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
+++ b/test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.OpenApi.Exceptions;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers.Exceptions;
+using Xunit;
+
+namespace Microsoft.OpenApi.Readers.Tests.V2Tests
+{
+    public class OpenApiDocumentTests
+    {
+        [Fact]
+        public void ShouldThrowWhenReferenceTypeIsInvalid()
+        {
+            var input = @"
+swagger: 2.0
+info: 
+  title: test
+  version: 1.0.0
+paths: 
+  '/':
+    get:
+      responses:
+        '200':
+          description: ok
+          schema:
+            $ref: '#/definition/does/notexist'
+";
+
+            var reader = new OpenApiStringReader(new OpenApiReaderSettings() {
+            });
+
+            OpenApiDocument doc = null;
+
+            Assert.Throws<OpenApiReaderException>(() =>
+            doc = reader.Read(input, out var diagnostic));
+
+            Assert.Null(doc);
+        }
+
+
+        [Fact]
+        public void ShouldThrowWhenReferenceDoesNotExist()
+        {
+            var input = @"
+swagger: 2.0
+info: 
+  title: test
+  version: 1.0.0
+paths: 
+  '/':
+    get:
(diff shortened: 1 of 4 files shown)


PerthCharern:
>                         return this.Components.Schemas[reference.Id];

If the reference.Id doesn't exist in Schemas.Keys, should we also throw the same error?

---
Refers to: src/Microsoft.OpenApi/Models/OpenApiDocument.cs:317 in d1b8b4d. [](commit_id = d1b8b4d4eeda113c2de8565b9dddcd8f27fb8ae5, deletion_comment = False)

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            $ref: '#/definition/does/notexist'
+";
+
+            var reader = new OpenApiStringReader(new OpenApiReaderSettings() {
+            });
+
+            OpenApiDocument doc = null;
+
+            Assert.Throws<OpenApiReaderException>(() =>
+            doc = reader.Read(input, out var diagnostic));
```
PerthCharern: >        doc = reader.Read(input, out var diagnostic)); [](start = 4, length = 54)

nit, declare the OpenApiDocument doc here instead of above

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
```diff
@@ -0,0 +1,77 @@
...
+          description: ok
+          schema:
+            $ref: '#/definitions/doesnotexist'
+";
+
+            var reader = new OpenApiStringReader(new OpenApiReaderSettings()
+            {
+            });
+
+            OpenApiDocument doc = null;
```
PerthCharern: nit, just declare the doc below at the same time as assignment

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
```diff
@@ -0,0 +1,77 @@
...
+  title: test
+  version: 1.0.0
+paths: 
+  '/':
+    get:
+      responses:
+        '200':
+          description: ok
+          schema:
+            $ref: '#/definition/does/notexist'
```
PerthCharern: >definition [](start = 21, length = 10)

nit, can we change this to something totally different, so that it's obvious it doesn't exist. At first glance, I was confused because I mistakenly read this as "definitions"?

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
PerthCharern: >   [](start = 10, length = 2)

nit, could you put one indent here?

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
```
PerthCharern: nit, copyright missing

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            });
+
+            OpenApiDocument doc = null;
+
+            doc = reader.Read(input, out var diagnostic);
+
+            diagnostic.Errors.ShouldBeEquivalentTo(new List<OpenApiError> {
+                new OpenApiError( new OpenApiException("Invalid Reference identifier 'doesnotexist'.")) });
+        }
+    }
```
PerthCharern: Why should we expect one case to throw an exception and the other case to simply add the error to the diagnostic? Shouldn't we standardize the behavior?

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+            });
+
+            OpenApiDocument doc = null;
+
+            doc = reader.Read(input, out var diagnostic);
+
+            diagnostic.Errors.ShouldBeEquivalentTo(new List<OpenApiError> {
+                new OpenApiError( new OpenApiException("Invalid Reference identifier 'doesnotexist'.")) });
+        }
```
PerthCharern: >      } [](start = 2, length = 7)

add check for the expected state of doc.

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
darrelmiller: I accidentally found this scenario by mistyping :-)

---
In reply to: [194477506](https://github.com/Microsoft/OpenAPI.NET/pull/275#discussion_r194477506) [](ancestors = 194477506)

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
darrelmiller: Yes.  This is a problem because one error is raised from OpenAPI.Net and so is a OpenApiException which is caught.  However, the other raises OpenApiReaderException which isn't caught.  This is fixed in the next PR so both types of exceptions are caught and put into the diagnostics error collection.  I will fix the test when I merge.

---
In reply to: [194478408](https://github.com/Microsoft/OpenAPI.NET/pull/275#discussion_r194478408) [](ancestors = 194478408)

darrelmiller:
>                         return this.Components.Schemas[reference.Id];

We do.  Are you suggesting we should return a different error?

---
In reply to: [396314166](https://github.com/Microsoft/OpenAPI.NET/pull/275#issuecomment-396314166) [](ancestors = 396314166)

---
Refers to: src/Microsoft.OpenApi/Models/OpenApiDocument.cs:317 in d1b8b4d. [](commit_id = d1b8b4d4eeda113c2de8565b9dddcd8f27fb8ae5, deletion_comment = False)

PerthCharern:
>                         return this.Components.Schemas[reference.Id];

Won't it be throwing KeyNotFoundException? I would rather it throw OpenApiException.

---
In reply to: [396439436](https://github.com/Microsoft/OpenAPI.NET/pull/275#issuecomment-396439436) [](ancestors = 396439436,396314166)

---
Refers to: src/Microsoft.OpenApi/Models/OpenApiDocument.cs:317 in d1b8b4d. [](commit_id = d1b8b4d4eeda113c2de8565b9dddcd8f27fb8ae5, deletion_comment = False)

darrelmiller:
>                         return this.Components.Schemas[reference.Id];

There is a try catch around the switch that does exactly that.

---
In reply to: [396673889](https://github.com/Microsoft/OpenAPI.NET/pull/275#issuecomment-396673889) [](ancestors = 396673889,396439436,396314166)

---
Refers to: src/Microsoft.OpenApi/Models/OpenApiDocument.cs:317 in d1b8b4d. [](commit_id = d1b8b4d4eeda113c2de8565b9dddcd8f27fb8ae5, deletion_comment = False)

PerthCharern:
>                         return this.Components.Schemas[reference.Id];

You are right. My bad :|

---
In reply to: [396676426](https://github.com/Microsoft/OpenAPI.NET/pull/275#issuecomment-396676426) [](ancestors = 396676426,396673889,396439436,396314166)

---
Refers to: src/Microsoft.OpenApi/Models/OpenApiDocument.cs:317 in d1b8b4d. [](commit_id = d1b8b4d4eeda113c2de8565b9dddcd8f27fb8ae5, deletion_comment = False)

PerthCharern:
>             } catch(KeyNotFoundException)

nit, could you move catch down one line (and add one space after the word "catch" )?

---
Refers to: src/Microsoft.OpenApi/Models/OpenApiDocument.cs:346 in a915e5b. [](commit_id = a915e5be0fce4cbe65ffb8bc3343d37b7b376d98, deletion_comment = False)

PR #278, opened 2018-06-12
Updated all deserializers to allow loading custom extensions

darrelmiller:


PerthCharern:
>         }

This should be in [https://github.com/Microsoft/OpenAPI.NET/blob/master/src/Microsoft.OpenApi.Readers/V3/OpenApiV3Deserializer.cs](https://github.com/Microsoft/OpenAPI.NET/blob/master/src/Microsoft.OpenApi.Readers/V3/OpenApiV3Deserializer.cs)

---
Refers to: src/Microsoft.OpenApi.Readers/V3/OpenApiDocumentDeserializer.cs:71 in 60da5b7. [](commit_id = 60da5b7cbe65b5ed9fc37423146fdfddca8938c0, deletion_comment = False)

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -228,7 +229,18 @@ private static void FixRequestBodyReferences(OpenApiDocument doc)
...
+        {
+            if (node.Context.ExtensionParsers.TryGetValue(name, out var parser))
+            {
+                return parser(node.CreateAny());
+            }
+            else
+            {
+                return node.CreateAny();
+            }
         }
```
PerthCharern: This should be in [https://github.com/Microsoft/OpenAPI.NET/blob/master/src/Microsoft.OpenApi.Readers/V2/OpenApiV2Deserializer.cs](https://github.com/Microsoft/OpenAPI.NET/blob/master/src/Microsoft.OpenApi.Readers/V2/OpenApiV2Deserializer.cs)

PR #280, opened 2018-06-12
Added ContextKeys object to VisitorBase

darrelmiller:
This allows visitors to process the OpenAPIDocument based on context of where an object exists within the document.

### DIFF
--- a/src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
@@ -15,20 +15,11 @@ namespace Microsoft.OpenApi.Services
     public abstract class OpenApiVisitorBase
     {
         private readonly Stack<string> _path = new Stack<string>();
-        private CurrentKeys _currentKeys = new CurrentKeys();
-
-
 
         /// <summary>
         /// Properties available to identify context of where an object is within OpenAPI Document
         /// </summary>
-        public CurrentKeys CurrentKeys
-        {
-            get
-            {
-                return _currentKeys;
-            }
-        }
+        public CurrentKeys CurrentKeys { get; } = new CurrentKeys();
 
         /// <summary>
         /// Allow Rule to indicate validation error occured at a deeper context level.  
@@ -57,17 +48,6 @@ namespace Microsoft.OpenApi.Services
                 return "#/" + String.Join("/", _path.Reverse());
             }
         }
-
-        internal void AttachCurrentKeys(CurrentKeys currentKeys)
-        {
-
-        }
-
-        internal void ClearCurrentKeys()
-        {
-
-        }
-    
 
         /// <summary>
         /// Visits <see cref="OpenApiDocument"/>
--- a/src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
@@ -15,7 +15,21 @@ namespace Microsoft.OpenApi.Services
     public abstract class OpenApiVisitorBase
     {
         private readonly Stack<string> _path = new Stack<string>();
-       
+        private CurrentKeys _currentKeys = new CurrentKeys();
+
+
+
+        /// <summary>
+        /// Properties available to identify context of where an object is within OpenAPI Document
+        /// </summary>
+        public CurrentKeys CurrentKeys
+        {
+            get
+            {
+                return _currentKeys;
+            }
+        }
+
         /// <summary>
         /// Allow Rule to indicate validation error occured at a deeper context level.  
         /// </summary>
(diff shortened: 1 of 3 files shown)


PerthCharern:
What's an example use case for this?

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
```diff
@@ -15,7 +15,21 @@ namespace Microsoft.OpenApi.Services
...
+        /// <summary>
+        /// Properties available to identify context of where an object is within OpenAPI Document
+        /// </summary>
+        public CurrentKeys CurrentKeys
+        {
+            get
+            {
+                return _currentKeys;
+            }
+        }
```
PerthCharern: Does this syntactic sugar work?

public CurrentKeys CurrentKeys { get; } = new CurrentKeys();

No need to list out the private field directly.

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
```diff
@@ -44,6 +58,15 @@ public string PathString
...
 
+        internal void AttachCurrentKeys(CurrentKeys currentKeys)
+        {
+
+        }
+
+        internal void ClearCurrentKeys()
+        {
+
+        }
```
PerthCharern: This is not used, and not marked virtual. Not sure what it is for?

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -787,6 +808,7 @@ internal void Walk(IDictionary<string,OpenApiExample> examples)
             {
                 foreach (var example in examples)
                 {
+                    _visitor.CurrentKeys.Example = example.Key;
                     Walk(example.Key, () => Walk(example.Value));
                 }
             }
```
PerthCharern: set the CurrentKeys.Example to null after the loop

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
darrelmiller: Yeah, that will work.

My particular usecase is fairly obscure.  When we import an OpenAPI document into API Management, we take any inline schemas and move them into the components/schemas section so that we can store all of the schemas as a single JSON Schema document.  To do that we need to generate a name for those schemas.  In order to generate a unique name it is important to know the path/operation/response code/mediatype that the schema lives under.

More generically, I could imagine tooling that walks the an OpenAPIDocument and checks schemas for certain characteristics.  Knowing if it is a schema under a Request or a Response is very useful.  Knowing if it is a schema for a 500 or a 200 might be useful.  Knowing the media type of an example if you want to verify examples would be useful.

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
darrelmiller: I forgot to delete it.  I did have CurrentKeys in the Walker and then I moved it into the visitor base.

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
PerthCharern: What do you think about putting it inside the loop instead?

                foreach (var example in examples)
                {
                    _visitor.CurrentKeys.Example = example.Key;
                    Walk(example.Key, () => Walk(example.Value));
                    _visitor.CurrentKeys.Example = null;
                }

I know it's less efficient, but it's more semantically correct in my opinion. Once we finish processing that example object, we should set the key back to null, instead of relying on the value being overridden by the next loop item.

---
In reply to: [194930193](https://github.com/Microsoft/OpenAPI.NET/pull/280#discussion_r194930193) [](ancestors = 194930193)

src/Microsoft.OpenApi/Services/OpenApiVisitorBase.cs
PerthCharern: Interesting... Technically, you can use the path stack to get that unique key for your use case I think. But having the keys in strong-typed object like this may prove useful in the future. I'll sign off.

---
In reply to: [194930721](https://github.com/Microsoft/OpenAPI.NET/pull/280#discussion_r194930721) [](ancestors = 194930721)

darrelmiller: Yeah, that's what I have done up to now.  However, it doesn't allow me to mimic the previous behaviour of our import for V2 OpenAPI and I do thing there is value in being able to get at these discrete pieces of context information.,

---
In reply to: [194931324](https://github.com/Microsoft/OpenAPI.NET/pull/280#discussion_r194931324) [](ancestors = 194931324,194930721)

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
darrelmiller: I'm ok with that.

---
In reply to: [194930909](https://github.com/Microsoft/OpenAPI.NET/pull/280#discussion_r194930909) [](ancestors = 194930909,194930193)

PR #289, opened 2018-06-25
Handle parsing port in host when reading V2

darrelmiller:


src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -153,12 +153,23 @@ private static void MakeServers(IList<OpenApiServer> servers, ParsingContext con
                     {
                         host = "//" + host;  // The double slash prefix creates a relative url where the scheme is defined by the BaseUrl
                     }
-
+                    int? port = null;
```
PerthCharern: >int? [](start = 20, length = 4)

nullable not needed --  int.Parse(pieces.Last()) throws an error if pieces.Last() is not parsable.

PerthCharern:
> 

Same changes as above needed here too?

---
Refers to: src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs:197 in dff6ff7. [](commit_id = dff6ff7aa0e66a1aba4355bdbab601eb38c80b06, deletion_comment = False)

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
darrelmiller: @PerthCharern I think my logic is correct here.  If the host does not contain ":" then we won't try to call int.Parse(pieces.Last()) and therefore port will be null.  I don't want to update uriBuilder.Port if port isn't specified because the URL should omit port.
If the host does contain ":" and the part after the colon is not a number, I'm ok with it throwing a parsing exception.

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -149,39 +149,19 @@ private static void MakeServers(IList<OpenApiServer> servers, ParsingContext con
             {
```
PerthCharern: If schemes is set to an empty list, I think we should treat that in the same way as schemes being null, i.e. line 148 should be `if (schemes != null && schemes.Count != 0)`

Spec says:
If the schemes is not included, the default scheme to be used is the one used to access the Swagger definition itself.

The term "not included" is not really precisely defined, but I think it should also apply to an empty schemes list. If we don't specifically handle it here, we won't have any `server` when `schemes` is an empty list.

src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
PerthCharern: That makes sense.

PerthCharern:
I approved, but please see the above comment before checking in.

Issue #298, opened 2018-07-19
Add a reader for YamlDocument object


talamram:
Hi,

I would like to add a new reader (except OpenApiStreamReader and OpenApiStringReader) that called OpenApiDocumentReader, and gives the ability to parse YamlDocument SharpYaml object directly.
The class signature will be:

public class OpenApiDocumentReader : IOpenApiReader<YamlDocument, OpenApiDiagnostic>

The OpenApiStreamReader will use it after it generates YamlDocument.
This class is useful for developers who want to generate the YamlDocument object independently (with SharpYaml package) and want to be able to parse it with OpenApi parsers.

Tanks, Tal.

darrelmiller:
@talamram That sounds great.  If you could create a PR against the vnext branch we should be able to get that into our 1.1 release.

darrelmiller:
Unfortunately, we don't want to expose our dependency on YamlDocument.

PR #300, opened 2018-07-22
Talamr/add document reader

talamram:
#298

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiDocumentReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiDocumentReader.cs
@@ -1,0 +1,131 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System;
+using Microsoft.OpenApi.Exceptions;
+using Microsoft.OpenApi.Extensions;
+using Microsoft.OpenApi.Interfaces;
+using Microsoft.OpenApi.Models;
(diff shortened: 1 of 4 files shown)


darrelmiller:
@talamram This looks good.  The only thing we have to be a bit concerned about is that I think this is the only place that we expose our usage of the SharpYaml package in our public API.  If we haven't already done that we need to decide if we want to tie our public interface to that library. I think currently, SharpYaml is an implementation detail.

talamram:
Hi,
Unfortunetly, I can't deliver it since it turn out that I can't sign the CLA contract :( (my copmpany doesn't allow it).
If it look useful in your opinion, please deliver it.

Thanks!

Jetski5822:
@talamram you explained to them that it’s in their interest to sign it?? Will benefit them...

talamram:
Yes I did... Out of my control unfortunately

darrelmiller:
@talamram Let me investigate if there is a way we can refactor some of the code in the StreamReader class so that you can create your own YamlDocumentReader without having to re-create all the initialization code.  Maybe if we moved some of the code into an abstract base class?

darrelmiller:
There is now a YamlDocumentReader 
https://github.com/microsoft/OpenAPI.NET/blob/vnext/src/Microsoft.OpenApi.Readers/OpenApiYamlDocumentReader.cs

However, it is marked as internal.  We currently do not want to expose the dependency on this library publicly.  However, we may reconsider that decision.

PR #316, opened 2018-08-16
Open api schema diff

Shwetap05:
Added  more comparers.

### DIFF
--- a/src/Microsoft.OpenApi/Services/ComparisonContext.cs
+++ b/src/Microsoft.OpenApi/Services/ComparisonContext.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Services
+{
+    /// <summary>
+    /// A class encapsulating the comarision context.
+    /// </summary>
+    public class ComparisonContext
+    {
+        private readonly IList<OpenApiDifference> _openApiDifferences = new List<OpenApiDifference>();
+        private readonly Stack<string> _path = new Stack<string>();
+        internal OpenApiComparerFactory OpenApiComparerFactory;
+
+        /// <summary>
+        /// Creates instance of <see cref="ComparisonContext"/>.
+        /// </summary>
+        /// <param name="openApiComparerFactory"></param>
+        public ComparisonContext(OpenApiComparerFactory openApiComparerFactory)
+        {
+            OpenApiComparerFactory = openApiComparerFactory;
+        }
+
+        /// <summary>
+        /// Gets the list of open api differences.
+        /// </summary>
+        public IEnumerable<OpenApiDifference> OpenApiDifferences => _openApiDifferences;
+
+        /// <summary>
+        /// Pointer to the source of difference in the document.
+        /// </summary>
+        public string PathString => "#/" + string.Join("/", _path.Reverse());
+
+        /// <summary>
+        /// Adds an open api difference.
+        /// </summary>
+        /// <param name="openApiDifference">The open api difference to add.</param>
+        public void AddOpenApiDifference(OpenApiDifference openApiDifference)
+        {
+            if (openApiDifference == null)
+            {
+                throw Error.ArgumentNull(nameof(openApiDifference));
+            }
+
+            _openApiDifferences.Add(openApiDifference);
+        }
+
+        /// <summary>
+        /// Allow Rule to indicate difference occured at a deeper context level.
+        /// </summary>
+        /// <param name="segment">Identifier for the context.</param>
+        public void Enter(string segment)
+        {
+            _path.Push(segment);
+        }
+
+        /// <summary>
+        /// Exit from path context level.  Enter and Exit calls should be matched.
+        /// </summary>
+        public void Exit()
+        {
+            _path.Pop();
+        }
+
+        /// <summary>
+        /// Gets the comparer instance for the requested type.
+        /// </summary>
+        /// <typeparam name="T">Type of requested comparer.</typeparam>
+        /// <returns>Comparer instance to use when comparing requested type.</returns>
+        internal OpenApiComparerBase<T> GetComparer<T>()
+        {
+            return OpenApiComparerFactory.GetComparer<T>();
+        }
+    }
+}
--- a/src/Microsoft.OpenApi/Services/ComparisonContext.cs
+++ b/src/Microsoft.OpenApi/Services/ComparisonContext.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.OpenApi.Models;
 
 namespace Microsoft.OpenApi.Services
 {
@@ -13,15 +14,23 @@ namespace Microsoft.OpenApi.Services
     {
         private readonly IList<OpenApiDifference> _openApiDifferences = new List<OpenApiDifference>();
         private readonly Stack<string> _path = new Stack<string>();
+        internal readonly OpenApiDocument SourceDocument;
+        internal readonly Stack<OpenApiSchema> SourceSchemaLoop = new Stack<OpenApiSchema>();
+        internal readonly OpenApiDocument TargetDocument;
+        internal readonly Stack<OpenApiSchema> TargetSchemaLoop = new Stack<OpenApiSchema>();
         internal OpenApiComparerFactory OpenApiComparerFactory;
 
         /// <summary>
         /// Creates instance of <see cref="ComparisonContext"/>.
         /// </summary>
-        /// <param name="openApiComparerFactory"></param>
-        public ComparisonContext(OpenApiComparerFactory openApiComparerFactory)
+        public ComparisonContext(
+            OpenApiComparerFactory openApiComparerFactory,
+            OpenApiDocument sourceDocument,
+            OpenApiDocument targetDocument)
         {
             OpenApiComparerFactory = openApiComparerFactory;
+            SourceDocument = sourceDocument;
+            TargetDocument = targetDocument;
         }
 
         /// <summary>
--- a/src/Microsoft.OpenApi/Services/ComparisonContext.cs
+++ b/src/Microsoft.OpenApi/Services/ComparisonContext.cs
@@ -3,12 +3,11 @@
 
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.OpenApi.Models;
 
 namespace Microsoft.OpenApi.Services
 {
     /// <summary>
-    /// A class encapsulating the comarision context.
+    /// A class encapsulating the comparision context.
     /// </summary>
     public class ComparisonContext
     {
--- a/src/Microsoft.OpenApi/Services/OpenApiComparer.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiComparer.cs
@@ -26,7 +26,7 @@ namespace Microsoft.OpenApi.Services
                 throw Error.ArgumentNull(nameof(target));
             }
 
-            var comparisionContext = new ComparisonContext(new OpenApiComparerFactory());
+            var comparisionContext = new ComparisonContext(new OpenApiComparerFactory(), source, target);
 
             new OpenApiDocumentComparer().Compare(source, target, comparisionContext);
 
--- a/src/Microsoft.OpenApi/Services/OpenApiComparer.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiComparer.cs
@@ -26,11 +26,11 @@ namespace Microsoft.OpenApi.Services
                 throw Error.ArgumentNull(nameof(target));
             }
 
-            var comparisionContext = new ComparisonContext(new OpenApiComparerFactory(), source, target);
+            var comparisonContext = new ComparisonContext(new OpenApiComparerFactory(), source, target);
 
-            new OpenApiDocumentComparer().Compare(source, target, comparisionContext);
+            new OpenApiDocumentComparer().Compare(source, target, comparisonContext);
 
-            return comparisionContext.OpenApiDifferences;
+            return comparisonContext.OpenApiDifferences;
         }
     }
 }
--- a/src/Microsoft.OpenApi/Services/OpenApiComparer.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiComparer.cs
@@ -14,10 +14,23 @@ namespace Microsoft.OpenApi.Services
         /// <summary>
         /// Compares two <see cref="OpenApiDocument"/>s and returns a list of differences.
         /// </summary>
-        public static List<OpenApiDifference> Compare(OpenApiDocument source, OpenApiDocument target)
+        public static IEnumerable<OpenApiDifference> Compare(OpenApiDocument source, OpenApiDocument target)
         {
-            var diffs = new List<OpenApiDifference>();
-            return diffs;
+            if (source == null)
+            {
+                throw Error.ArgumentNull(nameof(source));
+            }
+
+            if (target == null)
+            {
+                throw Error.ArgumentNull(nameof(target));
+            }
+
+            var comparisionContext = new ComparisonContext(new OpenApiComparerFactory());
+
+            new OpenApiDocumentComparer().Compare(source, target, comparisionContext);
+
+            return comparisionContext.OpenApiDifferences;
         }
     }
 }
--- a/src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
@@ -2,11 +2,12 @@
 // Licensed under the MIT license.
 
 using System;
+using Microsoft.OpenApi.Models;
 
 namespace Microsoft.OpenApi.Services
 {
     /// <summary>
-    /// Defines behavior for comparing parts of <see cref="OpenAPiDocument"/> class.
+    /// Defines behavior for comparing parts of <see cref="OpenApiDocument"/> class.
     /// </summary>
     /// <typeparam name="T">Type of class to compare.</typeparam>
     public abstract class OpenApiComparerBase<T>
@@ -64,6 +65,32 @@ namespace Microsoft.OpenApi.Services
                 {
                     OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
                     OpenApiComparedElementType = typeof(bool),
+                    SourceValue = source,
+                    TargetValue = target,
+                    Pointer = comparisonContext.PathString
+                });
+            }
+        }
+
+        /// <summary>
+        /// Compares two decimal object.
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <param name="target">The target.</param>
+        /// <param name="comparisonContext">The context under which to compare the objects.</param>
+        internal void Compare(decimal? source, decimal? target, ComparisonContext comparisonContext)
+        {
+            if (source == null && target == null)
+            {
+                return;
+            }
+
+            if (source != target)
+            {
+                comparisonContext.AddOpenApiDifference(new OpenApiDifference
+                {
+                    OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
+                    OpenApiComparedElementType = typeof(decimal),
                     SourceValue = source,
                     TargetValue = target,
                     Pointer = comparisonContext.PathString
--- a/src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
@@ -64,7 +64,7 @@ namespace Microsoft.OpenApi.Services
                 comparisonContext.AddOpenApiDifference(new OpenApiDifference
                 {
                     OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
-                    OpenApiComparedElementType = typeof(bool),
+                    OpenApiComparedElementType = typeof(bool?),
                     SourceValue = source,
                     TargetValue = target,
                     Pointer = comparisonContext.PathString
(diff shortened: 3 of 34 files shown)


src/Microsoft.OpenApi/Services/ComparisonContext.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.OpenApi.Models;
```
Shwetap05: >ft.OpenAp [](start = 13, length = 9)

Still adding more tests for comparers

src/Microsoft.OpenApi/Services/OpenApiComparer.cs
```diff
@@ -26,7 +26,7 @@ public static IEnumerable<OpenApiDifference> Compare(OpenApiDocument source, Ope
                 throw Error.ArgumentNull(nameof(target));
             }
 
-            var comparisionContext = new ComparisonContext(new OpenApiComparerFactory());
+            var comparisionContext = new ComparisonContext(new OpenApiComparerFactory(), source, target);
```
PerthCharern: >comparisionContext [](start = 16, length = 18)

nit, typo

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
```diff
@@ -63,7 +64,73 @@ internal void Compare(bool? source, bool? target, ComparisonContext comparisonCo
...
+            }
+        }
+
+        /// <summary>
+        /// Compares Enum.
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <param name="target">The target.</param>
+        /// <param name="comparisonContext">The context under which to compare the objects.</param>
+        internal void Compare<TE>(Enum source, Enum target, ComparisonContext comparisonContext)
```
PerthCharern: >TE [](start = 30, length = 2)

TEnum

src/Microsoft.OpenApi/Services/OpenApiDocumentComparer.cs
```diff
@@ -14,21 +15,35 @@ public class OpenApiDocumentComparer : OpenApiComparerBase<OpenApiDocument>
...
+            WalkAndCompare(
+                comparisonContext,
+                OpenApiConstants.Components,
+                () => comparisonContext
+                    .GetComparer<OpenApiComponents>()
+                    .Compare(sourceDocument.Components, targetDocument.Components, comparisonContext));
+
+            WalkAndCompare(
+                comparisonContext,
+                OpenApiConstants.Components,
```
PerthCharern: >Components [](start = 33, length = 10)

Servers

Shwetap05: good catch :-)

---
In reply to: [220656072](https://github.com/Microsoft/OpenAPI.NET/pull/316#discussion_r220656072) [](ancestors = 220656072)

PerthCharern:
>             comparisonContext.Enter(segment.Replace("/", "~1"));

need to escape ~ to ~0 too.

comparisonContext.Enter(segment.Replace("~", "~0").Replace("/", "~1"));

Same thing with the one below.

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs:152 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiEncodingComparer.cs
```diff
@@ -0,0 +1,68 @@
...
+            WalkAndCompare(comparisonContext, OpenApiConstants.ContentType,
+                () => Compare(sourceEncoding.ContentType, targetEncoding.ContentType, comparisonContext));
+
+            WalkAndCompare(comparisonContext, OpenApiConstants.Explode,
+                () => Compare(sourceEncoding.Explode, targetEncoding.Explode, comparisonContext));
+
+            WalkAndCompare(comparisonContext, OpenApiConstants.AllowReserved,
+                () => Compare(sourceEncoding.AllowReserved, targetEncoding.AllowReserved, comparisonContext));
+
+
```
PerthCharern: nit, remove this extra line

PerthCharern:
> }

Add comparer for reference object?

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParameterComparer.cs:61 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiHeaderComparer.cs
```diff
@@ -0,0 +1,112 @@
...
+                return;
+            }
+
+            if (sourceHeader.Reference != null
+                && targetHeader.Reference != null
+                && sourceHeader.Reference.Id != targetHeader.Reference.Id)
+            {
+                WalkAndAddOpenApiDifference(
+                    comparisonContext,
+                    "$ref",
```
PerthCharern: >"$ref" [](start = 20, length = 6)

Use DollarRef constant

src/Microsoft.OpenApi/Services/OpenApiHeaderComparer.cs
```diff
@@ -0,0 +1,112 @@
...
+                        OpenApiComparedElementType = typeof(OpenApiReference)
+                    });
+
+                return;
+            }
+
+            if (sourceHeader.Reference != null)
+            {
+                sourceHeader = (OpenApiHeader) comparisonContext.SourceDocument.ResolveReference(
+                    targetHeader.Reference);
```
PerthCharern: >targetHeader [](start = 20, length = 12)

sourceHeader

src/Microsoft.OpenApi/Services/OpenApiMediaTypeComparer.cs
```diff
@@ -0,0 +1,61 @@
...
+                return;
+            }
+
+            comparisonContext
+                .GetComparer<OpenApiSchema>()
+                .Compare(sourceMediaType.Schema, targetMediaType.Schema, comparisonContext);
+
+            WalkAndCompare(
+                comparisonContext,
+                OpenApiConstants.Content,
```
PerthCharern: >Content [](start = 33, length = 7)

Encoding

PerthCharern:
>     public static class OpenApiComparer

Is it not possible to model this based on the VisitorBase, i.e. in a similar way to [https://github.com/Microsoft/OpenAPI.NET/blob/vnext/src/Microsoft.OpenApi/Validations/OpenApiValidator.cs](https://github.com/Microsoft/OpenAPI.NET/blob/vnext/src/Microsoft.OpenApi/Validations/OpenApiValidator.cs)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiComparer.cs:12 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiResponseComparer.cs
```diff
@@ -0,0 +1,96 @@
...
+                return;
+            }
+
+            if (sourceResponse.Reference != null
+                && targetResponse.Reference != null
+                && sourceResponse.Reference.Id != targetResponse.Reference.Id)
+            {
+                WalkAndAddOpenApiDifference(
+                    comparisonContext,
+                    "$ref",
```
PerthCharern: >"$ref" [](start = 20, length = 6)

DollarRef constant

src/Microsoft.OpenApi/Services/OpenApiRequestBodyComparer.cs
```diff
@@ -0,0 +1,91 @@
...
+                return;
+            }
+
+            if (sourceRequestBody.Reference != null
+                && targetRequestBody.Reference != null
+                && sourceRequestBody.Reference.Id != targetRequestBody.Reference.Id)
+            {
+                WalkAndAddOpenApiDifference(
+                    comparisonContext,
+                    "$ref",
```
PerthCharern: >"$ref" [](start = 20, length = 6)

DollarRef constant

src/Microsoft.OpenApi/Services/OpenApiPathItemComparer.cs
```diff
@@ -40,8 +40,20 @@ public class OpenApiPathItemComparer : OpenApiComparerBase<OpenApiPathItem>
...
+            WalkAndCompare(
+                comparisonContext,
+                OpenApiConstants.Parameters,
+                () => comparisonContext
+                    .GetComparer<IList<OpenApiParameter>>()
+                    .Compare(sourcePathItem?.Parameters, targetPathItem?.Parameters, comparisonContext));
+
+            WalkAndCompare(
+                comparisonContext,
+                OpenApiConstants.Parameters,
```
PerthCharern: >Parameters [](start = 33, length = 10)

Servers

PerthCharern:
>     public static class OpenApiComparer

That way, we don't need to handle the walking separately. We can use it as:

var comparingWalker = new OpenApiWalker(openApiComparer)

directly.

Something like how we use the validators in this test:

[https://github.com/Microsoft/OpenAPI.NET/blob/e28c438a0046c4da4eb6a22e7b450f1546376dba/test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs](https://github.com/Microsoft/OpenAPI.NET/blob/e28c438a0046c4da4eb6a22e7b450f1546376dba/test/Microsoft.OpenApi.Tests/Services/OpenApiValidatorTests.cs)

---
In reply to: [424881544](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-424881544) [](ancestors = 424881544)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiComparer.cs:12 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

test/Microsoft.OpenApi.Tests/Services/OpenApiComponentsTests.cs
```diff
@@ -0,0 +1,610 @@
...
+            differences.Count().ShouldBeEquivalentTo(expectedDifferences.Count);
+
+            var notExpectedDifferences = differences.Where(
+                actualDiff => !expectedDifferences.Any(
+                    expectedDiff => expectedDiff.Pointer == actualDiff.Pointer
+                        && expectedDiff.OpenApiComparedElementType == actualDiff.OpenApiComparedElementType
+                        && expectedDiff.OpenApiDifferenceOperation ==
+                            actualDiff.OpenApiDifferenceOperation)).ToList();
+
+            notExpectedDifferences.Count.ShouldBeEquivalentTo(0);
```
PerthCharern: Can't we just do

differences.ShouldBeEquivalentTo(expectedDifferences) ?

PerthCharern: Same comment with other test files.

---
In reply to: [220743939](https://github.com/Microsoft/OpenAPI.NET/pull/316#discussion_r220743939) [](ancestors = 220743939)

Shwetap05: So i can't do that, as in the difference result we have SourceValue and TargetValue of type object and equivalent fails when i use the actual type e.g. Server in expected result.

---
In reply to: [220743996](https://github.com/Microsoft/OpenAPI.NET/pull/316#discussion_r220743996) [](ancestors = 220743996,220743939)

Shwetap05:
>                     i.ToString(),

Agree the diff that is being performed above is not index to index so if 

source = [1,2,3]
target =[3,1,4,5]

diff is 4,5 new but the index of then in path would be 0.1 which is not correct, it should be 2,3 respectively.

so i will fix code to represent the location from the source/target list. so targetParameters.IndexOf( newParametersInTarget[i] ).ToString()


---
In reply to: [424889731](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-424889731) [](ancestors = 424889731)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:46 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

Shwetap05:
> }

added a to do for it. Will take care of it in next PR.

---
In reply to: [424879209](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-424879209) [](ancestors = 424879209)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParameterComparer.cs:61 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

Shwetap05:
>     public static class OpenApiComparer

While designing this feature i did go through these files implementing visitor pattern, and couldn't use them b/c VisitorBase is the class that manages path stack and Validator derives from it and leverages visiting logic from it. But Comparer cannot derive it from VisitorBase as it just visits one document. Make sense?

---
In reply to: [424890444](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-424890444) [](ancestors = 424890444,424881544)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiComparer.cs:12 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

PerthCharern:
>     public static class OpenApiComparer

That's fair. There might be a way to unify these two, but that will be much more work.

---
In reply to: [425254973](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425254973) [](ancestors = 425254973,424890444,424881544)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiComparer.cs:12 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

PerthCharern:
>                     i.ToString(),

But order should matter. 

If

source [3,1] 
target [1,3]

This should report 2 updates. If we are not going with the index-to-index comparison, we'll need to first find a baseline on how to determine which objects remain the same in the list.

---
In reply to: [425251278](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425251278) [](ancestors = 425251278,424889731)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:46 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

PerthCharern:
>             comparisonContext.Enter(segment.Replace("/", "~1"));

It must be in the order I described above. Otherwise the ~ in ~1 will get replaced twice

i.e. / will be come ~01 (which is not what we want)

---
In reply to: [424875930](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-424875930) [](ancestors = 424875930)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs:152 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

Shwetap05:
>                     i.ToString(),

But if i am consumer of diff and i moved around my parameters in terms of index so per me that should generate a diff

---
In reply to: [425256973](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425256973) [](ancestors = 425256973,425251278,424889731)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:46 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

Shwetap05:
>             comparisonContext.Enter(segment.Replace("/", "~1"));

ah yes. long day for me :-).

---
In reply to: [425257539](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425257539) [](ancestors = 425257539,424875930)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs:152 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

PerthCharern:
>                     i.ToString(),

Not sure if I understand. Based on the current code, if you put in something like

source [3,1] 
target [1,3]

it will not generate any diff. This doesn't sound right to me.

---
In reply to: [425257623](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425257623) [](ancestors = 425257623,425256973,425251278,424889731)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:46 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

Shwetap05:
>                     i.ToString(),

so if you perform suppose a string list compare options in C# it also does index variant compare and it will not result in diff. And if in a OpenApiDocument i have parameters shuffled around e.g. like below it should not result in diff

SOURCE
"parameters": [
          {
            "name": "sampleHeaderParam1",
            "in": "header"
          },
          {
            "name": "sampleHeaderParam2",
            "in": "header"
            }
          }
}

TARGET:

"parameters": [
          {
            "name": "sampleHeaderParam2",
            "in": "header"
          },
          {
            "name": "sampleHeaderParam1",
            "in": "header"
            }
          }
}

---
In reply to: [425258563](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425258563) [](ancestors = 425258563,425257623,425256973,425251278,424889731)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:46 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

test/Microsoft.OpenApi.Tests/Services/OpenApiComponentsTests.cs
PerthCharern: ShouldBeEquivalentTo should be able to to do a deep comparison into any object type. In your expectedDifferences, you can assign the correct sourceValue and targetValue and FluentAssertion should be able to compare it with the differences you got.

In the case there's a bug in FluentAssertion library and the above doesn't work, you can simply exclude certain path from comparison in fluent assertion. [https://stackoverflow.com/questions/22142576/how-to-use-exclude-in-fluentassertions-for-property-in-collection](https://stackoverflow.com/questions/22142576/how-to-use-exclude-in-fluentassertions-for-property-in-collection)

This will ensure this test has good coverage even when we update OpenApiDifference in the future.

---
In reply to: [221082998](https://github.com/Microsoft/OpenAPI.NET/pull/316#discussion_r221082998) [](ancestors = 221082998,220743996,220743939)

PerthCharern:
>                     i.ToString(),

Discussed with @shwetap05. We will use sourceIndex for removal, and targetIndex for addition. We will treat parameter list as unordered given its semantic meaning.

---
In reply to: [425273672](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425273672) [](ancestors = 425273672,425262254,425260367,425258563,425257623,425256973,425251278,424889731)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:46 in 4a66112. [](commit_id = 4a66112152f3f4c3948c0e8c147bf2c2b5a72906, deletion_comment = False)

PerthCharern:
>             }

What if only one is null? 

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:31 in 91a905d. [](commit_id = 91a905d0630869e24cc02045896bbaf2334edd19, deletion_comment = False)

PerthCharern:
>             }

We should simply add one OpenApiDifference (add or remove) the parameter itself, not the item, is that right?

---
In reply to: [425517341](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425517341) [](ancestors = 425517341)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:31 in 91a905d. [](commit_id = 91a905d0630869e24cc02045896bbaf2334edd19, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs
```diff
@@ -43,7 +43,7 @@ public class OpenApiParametersComparer : OpenApiComparerBase<IList<OpenApiParame
             {
                 WalkAndAddOpenApiDifference(
                     comparisonContext,
-                    i.ToString(),
+                    targetParameters.IndexOf(newParametersInTarget[i]).ToString(),
```
PerthCharern: >.IndexOf(newParametersInTarget[i]) [](start = 36, length = 34)

Interesting way to use IndexOf. :)

Shwetap05:
>             }

oh looks like i missed that, yes i think i should add one diff with parameters

---
In reply to: [425517623](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425517623) [](ancestors = 425517623,425517341)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:31 in 91a905d. [](commit_id = 91a905d0630869e24cc02045896bbaf2334edd19, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs
```diff
@@ -75,6 +75,11 @@ public class OpenApiParametersComparer : OpenApiComparerBase<IList<OpenApiParame
...
                     .FirstOrDefault(param => param.Name == sourceParameter.Name && param.In == sourceParameter.In);
 
+                if (targetParameter == null)
+                {
+                    continue;
+                }
+
                 WalkAndCompare(
                     comparisonContext,
                     i.ToString(),
```
PerthCharern: >                   i.ToString(), [](start = 1, length = 32)

This one is kinda tricky but I think it's already alright. It makes sense to use the source index here.

PerthCharern:
>             }

As discussed, switch these two loops and reverse the order in the source loop. This will make our difference list compliant to JSON PATCH.

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:70 in 91a905d. [](commit_id = 91a905d0630869e24cc02045896bbaf2334edd19, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs
PerthCharern: Update: With the changes as discussed, we will need to use the target index here.

---
In reply to: [221344148](https://github.com/Microsoft/OpenAPI.NET/pull/316#discussion_r221344148) [](ancestors = 221344148)

PerthCharern:
>             }

Yup. Based on your other comparers, that would be an Update. Not add/remove. My bad.

---
In reply to: [425523780](https://github.com/Microsoft/OpenAPI.NET/pull/316#issuecomment-425523780) [](ancestors = 425523780,425517623,425517341)

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiParametersComparer.cs:31 in 91a905d. [](commit_id = 91a905d0630869e24cc02045896bbaf2334edd19, deletion_comment = False)

test/Microsoft.OpenApi.Tests/Services/OpenApiComponentsTests.cs
Shwetap05: interesting it worked, i remember it not working somewhere will fix all test.

---
In reply to: [221094506](https://github.com/Microsoft/OpenAPI.NET/pull/316#discussion_r221094506) [](ancestors = 221094506,221082998,220743996,220743939)

PerthCharern: It actually does compare, right? Did you try testing it by intentionally making it fail? Just want to make sure it's not just giving out an automatic pass without comparing.

---
In reply to: [221702898](https://github.com/Microsoft/OpenAPI.NET/pull/316#discussion_r221702898) [](ancestors = 221702898,221094506,221082998,220743996,220743939)

Shwetap05: yes tested it, it does compare it and not false positive.

---
In reply to: [221749561](https://github.com/Microsoft/OpenAPI.NET/pull/316#discussion_r221749561) [](ancestors = 221749561,221702898,221094506,221082998,220743996,220743939)

PR #317, opened 2018-08-30
Update SharpYaml package to 1.6.4 to enable support for .net standard…

ArchanaKasi:
Update SharpYaml package to 1.6.4 to enable support for .net standard 2.0


darrelmiller:
Hi @ArchanaKasi Thanks for the PR.  Could you help me understand why we would want to do this?  We currently support .Net Standard 1.6 and therefore any library that wants to target .net Standard 2.0 can consume this library.  Unless we actually use features in .net Standard that are not in 1.6, I don't see a reason for limiting the platforms that can use us.

ArchanaKasi:
Thanks Darrel. Makes sense.

Issue #313, opened 2018-08-12
Development process changes


darrelmiller:
Moving forward, I don't think we should continue with the vnext branch.  We ended up accumulating too many changes in the vnext branch and for every reported bug fix we had to deal with the question of should we merge the fix into master or into the vnext branch.  

There are bunch of recently opened bugs that I would like to get fixed and I wanted to get them into the 1.1 release, but it was continuing to delay the release and we already have people using the -preview Nuget in prod.

If we have a feature that merits a minor release, then we should just release it.  Accumulating these changes to keep the version numbers low is of dubious merit.  

On a related note, I would like to change the way we create release notes to use a markdown file in the repo.  The file should be updated as fixes and features are completed, rather than having to update the GitHub page.  We can still create GitHub releases and tags, but do it automatically from the VSTS release process.

Thoughts?  @PerthCharern @Shwetap05 @xuzhg @scott-lin @EricAtMSFT

Shwetap05:
@darrelmiller make sense, then there is no more preview of any features that are added to the library? e.g. i wanted Diff to be released as preview and if there is no VNext how do you suggest we release it? As a major version of library?

PerthCharern:
- Based on #260, shouldn't every bug fix go into both the master and the vNext branch?  If we make the bug fix go into vNext branch every time it's merged into master, then the vNext will always have the most updated patch as well. You should not have to deal with "the question of should we merge the fix into master or into the vnext branch."

- This would still allow speedy release of the minor version. The reason why we had to split it out was to allow an even speedier release of the patch when the minor version was under going "preview" release.

- Could you elaborate on the release note a bit? Are you suggesting we maintain a new markdown file? I am okay with any process as long as it's easy to figure out what has been released in each version.

darrelmiller:
@Shwetap05 We manually update the nuget package for each release.  There's no reason why we couldn't add a -preview tag in the master branch for a preview release.
There is an interesting variant described [here](https://oren.codes/2018/05/15/oss-build-and-release-with-vsts/)  where master always deployes a preview and you branch to a rel/1.0 rel/1.1 when you choose to cut a release.  One nice thing about that is that you don't need to create a dummy PR that needs to be approved just to remove the nuget -preview.

[...]

PR #320, opened 2018-09-26
Allow composite expression

### LINKED ISSUE #287: Callback key can be a string with embedded RuntimeExpression
@darrelmiller 

Issue from @epet, @Shwetap05 

Currently we only allow RuntimeExpression in a key in a Callback object.

```csharp
        /// <summary>
        /// A Path Item Object used to define a callback request and expected responses.
        /// </summary>
        public Dictionary<RuntimeExpression, OpenApiPathItem> PathItems { get; set; }
            = new Dictionary<RuntimeExpression, OpenApiPathItem>();
```

This is not correct given an example in the spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#callback-object-example

```
myWebhook:
  'http://notificationServer.com?transactionId={$request.body#/id}&email={$request.body#/email}':
    post:
      requestBody:
        description: Callback payload
        content: 
          'application/json':
            schema:
              $ref: '#/components/schemas/SomePayload'
      responses:
        '200':
          description: webhook successfully processed and no retries will be performed
```

Note that the "expression" key above is a URL with RuntimeExpression embedded in. 

Note: I think the spec is a bit ambiguous. The spec states this: "The key that identifies the Path Item Object is a runtime expression that can be evaluated in the context of a runtime HTTP request/response to identify the URL to be used for the callback request." 

The runtime expression in the spec is strictly defined. There's a bit of leeway with this sentence "Runtime expressions preserve the type of the referenced value. Expressions can be embedded into string values by surrounding the expression with {} curly braces.", but this sentence is in the Examples section of runtime expression section and it contradicts with the definition of runtime expression directly above: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#examples-1

I would say we are actually faithful to the spec with the current implementation - though probably not aligned with the spirit of the spec.

PerthCharern:
Fix #287 

- Allow composite expression with the following formats:
  - Strings with runtime expression embedded in { }
  - Plain strings without any runtime expression
- Strings with {$xxxx} where xxxx doesn't constitute a valid runtime expression is NOT allowed. A set of braces with a dollar sign denotes a runtime expression in a composite expression and should not be used in other random scenarios.
  - Note that the spec is a bit ambiguous in this regard. From a literal reading of the spec, the requirement is way more generous (any string should be okay). If we decide to allow that, then this requirement can be lifted later without being a breaking change.

- Adjust unit tests


PerthCharern:
@Shwetap05 @darrelmiller Do we have a consensus based on #313 on which branch to use? Are we still using vNext?

Shwetap05:
@PerthCharern we discussed to stick with vNext for preview features and yes bug fixes should go to both master and vNext

PR #328, opened 2018-10-05
Added more comparers

Shwetap05:
Added comparison logic for
- OpenApiInfo
- OpenApiExternalDocs
- OpenApiSecurityRequirements

### DIFF
--- a/src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.OpenApi.Models;
 
 namespace Microsoft.OpenApi.Services
@@ -39,6 +41,32 @@ namespace Microsoft.OpenApi.Services
                 {
                     OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
                     OpenApiComparedElementType = typeof(string),
+                    SourceValue = source,
+                    TargetValue = target,
+                    Pointer = comparisonContext.PathString
+                });
+            }
+        }
+
+        /// <summary>
+        /// Compares two Uri object.
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <param name="target">The target.</param>
+        /// <param name="comparisonContext">The context under which to compare the objects.</param>
+        internal void Compare(Uri source, Uri target, ComparisonContext comparisonContext)
+        {
+            if (source == null && target == null)
+            {
+                return;
+            }
+
+            if (source != target)
+            {
+                comparisonContext.AddOpenApiDifference(new OpenApiDifference
+                {
+                    OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
+                    OpenApiComparedElementType = typeof(Uri),
                     SourceValue = source,
                     TargetValue = target,
                     Pointer = comparisonContext.PathString
@@ -135,6 +163,125 @@ namespace Microsoft.OpenApi.Services
                     TargetValue = target,
                     Pointer = comparisonContext.PathString
                 });
+            }
+        }
+
+        /// <summary>
+        /// Compares <see cref="OpenApiReference"/> object.
+        /// </summary>
+        /// <param name="sourceReference">The source.</param>
+        /// <param name="targetReference">The target.</param>
+        /// <param name="comparisonContext">The context under which to compare the objects.</param>
+        internal void Compare<TReference>(
+            OpenApiReference sourceReference,
+            OpenApiReference targetReference,
+            ComparisonContext comparisonContext)
+        {
+            if (sourceReference == null && targetReference == null)
+            {
+                return;
+            }
+
+            if (sourceReference == null || targetReference == null)
+            {
+                comparisonContext.AddOpenApiDifference(
+                    new OpenApiDifference
+                    {
+                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
+                        SourceValue = sourceReference,
+                        TargetValue = targetReference,
+                        OpenApiComparedElementType = typeof(OpenApiReference),
+                        Pointer = comparisonContext.PathString
+                    });
+
+                return;
+            }
+
+            if (sourceReference.Id != targetReference.Id || sourceReference.Type != targetReference.Type)
+            {
+                WalkAndAddOpenApiDifference(
+                    comparisonContext,
+                    OpenApiConstants.DollarRef,
+                    new OpenApiDifference
+                    {
+                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
+                        SourceValue = sourceReference,
+                        TargetValue = targetReference,
+                        OpenApiComparedElementType = typeof(OpenApiReference)
+                    });
+
+                return;
+            }
+
+            var source = (TReference) comparisonContext.SourceDocument.ResolveReference(
+                sourceReference);
+
+            var target = (TReference) comparisonContext.TargetDocument.ResolveReference(
+                targetReference);
+
+            comparisonContext
+                .GetComparer<TReference>()
+                .Compare(source, target, comparisonContext);
+        }
+
+        /// <summary>
+        /// Compares <see cref="IDictionary{TKey,TValue}"/> where TKey is <see cref="string"/> and TValue is
+        /// <see cref="string"/>.
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <param name="target">The target.</param>
+        /// <param name="comparisonContext">The context under which to compare the objects.</param>
+        internal void Compare(IDictionary<string, string> source, IDictionary<string, string> target,
+            ComparisonContext comparisonContext)
+        {
+            if (source == null && target == null)
+            {
+                return;
+            }
+
+            if (source == null || target == null)
+            {
+                comparisonContext.AddOpenApiDifference(
+                    new OpenApiDifference
+                    {
+                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
+                        SourceValue = source,
+                        TargetValue = target,
+                        OpenApiComparedElementType = typeof(IDictionary<string, T>),
+                        Pointer = comparisonContext.PathString
+                    });
+
+                return;
+            }
+
+            var newKeysInTarget = target.Keys.Except(source.Keys).ToList();
+
+            foreach (var newKeyInTarget in newKeysInTarget)
+            {
+                comparisonContext.AddOpenApiDifference(
+                    new OpenApiDifference
+                    {
+                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Add,
+                        TargetValue = new KeyValuePair<string, string>(
+                            newKeyInTarget,
+                            target[newKeyInTarget]),
+                        OpenApiComparedElementType = typeof(KeyValuePair<string, string>)
+                    });
+            }
+
+            var removedKeysFromSource = source.Keys.Except(target.Keys).ToList();
+
+            foreach (var removedKeyFromSource in removedKeysFromSource)
+            {
+                comparisonContext.AddOpenApiDifference(
+                    new OpenApiDifference
+                    {
+                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Add,
+                        TargetValue = new KeyValuePair<string, string>(
+                            removedKeyFromSource,
+                            source[removedKeyFromSource]),
+                        OpenApiComparedElementType = typeof(KeyValuePair<string, string>)
+                    });
             }
         }
 
--- a/src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
@@ -167,64 +167,6 @@ namespace Microsoft.OpenApi.Services
         }
 
         /// <summary>
-        /// Compares <see cref="OpenApiReference"/> object.
-        /// </summary>
-        /// <param name="sourceReference">The source.</param>
-        /// <param name="targetReference">The target.</param>
-        /// <param name="comparisonContext">The context under which to compare the objects.</param>
-        internal void Compare<TReference>(
-            OpenApiReference sourceReference,
-            OpenApiReference targetReference,
-            ComparisonContext comparisonContext)
-        {
-            if (sourceReference == null && targetReference == null)
-            {
-                return;
-            }
-
-            if (sourceReference == null || targetReference == null)
-            {
-                comparisonContext.AddOpenApiDifference(
-                    new OpenApiDifference
-                    {
-                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Update,
-                        SourceValue = sourceReference,
-                        TargetValue = targetReference,
-                        OpenApiComparedElementType = typeof(OpenApiReference),
-                        Pointer = comparisonContext.PathString
-                    });
-
-                return;
-            }
-
-            if (sourceReference.Id != targetReference.Id || sourceReference.Type != targetReference.Type)
-            {
-                WalkAndAddOpenApiDifference(
-                    comparisonContext,
-                    OpenApiConstants.DollarRef,
-                    new OpenApiDifference
-                    {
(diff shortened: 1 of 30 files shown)


test/Microsoft.OpenApi.Tests/Services/OpenApiInfoComparerTests.cs
```diff
@@ -0,0 +1,297 @@
...
+                    Explode = true,
+                    AllowReserved = true
+                },
+                new OpenApiEncoding
+                {
+                    ContentType = "image/jpeg",
+                    Style = ParameterStyle.Form,
+                    Explode = false,
+                    AllowReserved = false
+                },
```
PerthCharern: This doesn't look right. Copy-paste error?

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
```diff
@@ -138,6 +166,125 @@ internal void Compare<TEnum>(Enum source, Enum target, ComparisonContext compari
...
+            var source = (TReference) comparisonContext.SourceDocument.ResolveReference(
+                sourceReference);
+
+            var target = (TReference) comparisonContext.TargetDocument.ResolveReference(
+                targetReference);
+
+            comparisonContext
+                .GetComparer<TReference>()
+                .Compare(source, target, comparisonContext);
+        }
```
PerthCharern: This seems like something that should be in its own class OpenApiReferenceComparer rather than the base class

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
```diff
@@ -138,6 +166,125 @@ internal void Compare<TEnum>(Enum source, Enum target, ComparisonContext compari
...
+            {
+                comparisonContext.AddOpenApiDifference(
+                    new OpenApiDifference
+                    {
+                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Add,
+                        TargetValue = new KeyValuePair<string, string>(
+                            newKeyInTarget,
+                            target[newKeyInTarget]),
+                        OpenApiComparedElementType = typeof(KeyValuePair<string, string>)
+                    });
```
PerthCharern: Based on this [http://jsonpatch.com/](http://jsonpatch.com/), this should be

                WalkAndAddOpenApiDifference(
                    comparisonContext,
                    newKeyInTarget,
                    new OpenApiDifference
                    {
                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Add,
                        TargetValue = target[newKeyInTarget],
                        OpenApiComparedElementType = typeof(string)
                    });

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
```diff
@@ -138,6 +166,125 @@ internal void Compare<TEnum>(Enum source, Enum target, ComparisonContext compari
...
+            {
+                comparisonContext.AddOpenApiDifference(
+                    new OpenApiDifference
+                    {
+                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Add,
+                        TargetValue = new KeyValuePair<string, string>(
+                            removedKeyFromSource,
+                            source[removedKeyFromSource]),
+                        OpenApiComparedElementType = typeof(KeyValuePair<string, string>)
+                    });
```
PerthCharern: WalkAndAddOpenApiDifference(
                    comparisonContext,
                    removedKeyFromSource,
                    new OpenApiDifference
                    {
                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Remove,
                        SourceValue = source[removedKeyFromSource],
                        OpenApiComparedElementType = typeof(string)
                    });

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
```diff
@@ -138,6 +166,125 @@ internal void Compare<TEnum>(Enum source, Enum target, ComparisonContext compari
...
+
+            var source = (TReference) comparisonContext.SourceDocument.ResolveReference(
+                sourceReference);
+
+            var target = (TReference) comparisonContext.TargetDocument.ResolveReference(
+                targetReference);
+
+            comparisonContext
+                .GetComparer<TReference>()
+                .Compare(source, target, comparisonContext);
```
PerthCharern: What's the example case of this? If the references are indeed pointing to the same object, wouldn't these two objects always be the same?

src/Microsoft.OpenApi/Services/OpenApiSecurityRequirementComparer.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                WalkAndAddOpenApiDifference(
+                    comparisonContext,
+                    newSecuritySchemeInTarget.Reference.Id,
+                    new OpenApiDifference
+                    {
+                        OpenApiDifferenceOperation = OpenApiDifferenceOperation.Add,
+                        TargetValue = new KeyValuePair<OpenApiSecurityScheme, IList<string>>(
+                            newSecuritySchemeInTarget,
+                            targetSecurityRequirement[newSecuritySchemeInTarget]),
```
PerthCharern: This should not be a key-value pair. We are already pointing the the value itself - just use the IList<string>.

src/Microsoft.OpenApi/Services/OpenApiSecurityRequirementComparer.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            var newSecuritySchemesInTarget = targetSecurityRequirement.Keys
+                .Where(targetReq => sourceSecurityRequirement.Keys.All(
+                    sourceReq => sourceReq.Reference.Id != targetReq.Reference.Id)).ToList();
+
+            foreach (var newSecuritySchemeInTarget in newSecuritySchemesInTarget)
+            {
+                WalkAndAddOpenApiDifference(
+                    comparisonContext,
+                    newSecuritySchemeInTarget.Reference.Id,
```
PerthCharern: > newSecuritySchemeInTarget.Reference.Id, [](start = 19, length = 40)

Does this yield the right string (i.e. just the security scheme name, not the json pointer)?

src/Microsoft.OpenApi/Services/OpenApiSecurityRequirementComparer.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    WalkAndAddOpenApiDifference(
+                        comparisonContext,
+                        sourceSecurityScheme.Reference.Id,
+                        new OpenApiDifference
+                        {
+                            OpenApiDifferenceOperation = OpenApiDifferenceOperation.Remove,
+                            SourceValue = new KeyValuePair<OpenApiSecurityScheme, IList<string>>(
+                                sourceSecurityScheme,
+                                sourceSecurityRequirement[sourceSecurityScheme]),
```
PerthCharern: similar to above. This should not be a Key Value Pair

src/Microsoft.OpenApi/Services/OpenApiSecurityRequirementComparer.cs
```diff
@@ -0,0 +1,96 @@
...
+                            OpenApiComparedElementType = typeof(KeyValuePair<OpenApiSecurityScheme, IList<string>>)
+                        });
+                }
+                else
+                {
+                    WalkAndCompare(comparisonContext,
+                        sourceSecurityScheme.Reference.Id,
+                        () => comparisonContext
+                            .GetComparer<OpenApiSecurityScheme>()
+                            .Compare(sourceSecurityScheme, targetSecurityScheme, comparisonContext));
```
PerthCharern: I don't think we should be comparing the OpenApiSecurityScheme. Shouldn't we be comparing the list of strings in the value side?

test/Microsoft.OpenApi.Tests/Services/OpenApiSecurityRequirementComparerTests.cs
```diff
@@ -0,0 +1,293 @@
...
+                    },
+                    [
+                        new OpenApiSecurityScheme
+                        {
+                            Reference = new OpenApiReference {Type = ReferenceType.SecurityScheme, Id = "scheme2"}
+                        }
+                    ] = new List<string>
+                    {
+                        "scope4",
+                        "scope5"
```
PerthCharern: Try add test cases with two different scope lists

test/Microsoft.OpenApi.Tests/Services/OpenApiInfoComparerTests.cs
Shwetap05: sorry what doesn't seem right?

---
In reply to: [223510742](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223510742) [](ancestors = 223510742)

PerthCharern:
This file should be comparing Info objects, not Encoding objects.

src/Microsoft.OpenApi/Services/OpenApiSecurityRequirementComparer.cs
Shwetap05: why not compare Security scheme? if source and target reference two different types of security schemes there should be a diff for it.

---
In reply to: [223516846](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223516846) [](ancestors = 223516846)

Shwetap05: That said i was thinking whether to compare scope list or not, and if i do compare how the pointer should be ?

e.g. if source is 


---
In reply to: [223852204](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223852204) [](ancestors = 223852204,223516846)

test/Microsoft.OpenApi.Tests/Services/OpenApiSecurityRequirementComparerTests.cs
Shwetap05: I am currently not comparing scope lists, have replied on other thread on question i have on diff of it

---
In reply to: [223517115](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223517115) [](ancestors = 223517115)

test/Microsoft.OpenApi.Tests/Services/OpenApiInfoComparerTests.cs
Shwetap05: oh got it yes copy paste error for sure

---
In reply to: [223851085](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223851085) [](ancestors = 223851085,223510742)

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
Shwetap05: I do not see a dictionary example for json patch , so i am not sure if this is correct, can you point me to a dictionary example that you might have found?

---
In reply to: [223511856](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223511856) [](ancestors = 223511856)

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
Shwetap05: if Id are same we can't assume that the referenced object will be same too.

---
In reply to: [223515635](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223515635) [](ancestors = 223515635)

src/Microsoft.OpenApi/Services/OpenApiSecurityRequirementComparer.cs
Shwetap05: Yes i have test for it in OpenApiComparerTests

---
In reply to: [223516346](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223516346) [](ancestors = 223516346)

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
PerthCharern: C# Dictionary should be treated as a JSON object.

The first example on [http://jsonpatch.com/](http://jsonpatch.com/)
See the second line:

 { "op": "add", "path": "/hello", "value": ["world"] }

and the third line:

{ "op": "remove", "path": "/foo" }

---
In reply to: [223857178](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223857178) [](ancestors = 223857178,223511856)

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
PerthCharern: That could be true in the C# sense, but from the way we use the Reference object, how would that be possible? If the ID is the same, doesn't that mean this refers to exactly the same object. In fact, the reference object will only be serialized as $ref (or plain string in case of tag) with id only.

---
In reply to: [223858079](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223858079) [](ancestors = 223858079,223515635)

src/Microsoft.OpenApi/Services/OpenApiSecurityRequirementComparer.cs
PerthCharern: - We should not compare SecurityScheme here because we only use its name as a key. From the OpenAPI document standpoint, only the key (id of the ref) will show up. Even if we manually craft the DOM to actually be different (say, add some random other things into the Security Scheme), the resulting serialized document will be exactly the same for this SecurityRequirement object.

- Ideally, diff in the above case should be Remove /scheme1/0. However, from the way you handle the Ordered List (in OpenApiOrderedListComparer), this can be
  
   - Update /scheme1/0 from scope1 to scope2
   - Update /scheme1/1 from scope2 to scope3
   - Remove /scheme1/2

  which is also okay (and correct).



---
In reply to: [223852979](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223852979) [](ancestors = 223852979,223852204,223516846)

PerthCharern: If we want to be really fancy, we can use an algo for "minimum edit distance" to find the best list of update/add/remove (think about one scope being one letter in the original problem).

---
In reply to: [223900360](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223900360) [](ancestors = 223900360,223852979,223852204,223516846)

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
PerthCharern: Chatted with Shweta. We will do a deep comparison here, so that it's useful for API consumer to see what exactly has changed to an API. Same principle applies to the Security Scheme in Security Requirement.

---
In reply to: [223899238](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223899238) [](ancestors = 223899238,223858079,223515635)

src/Microsoft.OpenApi/Services/OpenApiComparerBase.cs
Shwetap05: I tried that but didn't go it b/c even if i add as separate comparer class like below

public class OpenApiReferenceComparer<T> : OpenApiComparerBase<OpenApiReference> where T : IOpenApiReferenceable

It can't be registered under OpenApiComparerFactory, as there the type of reference is an enum and can't register varies reference comparer based on the value of enum,

That said i will still add the comparer class and reference it directly in the comparers where we diff reference.


---
In reply to: [223511051](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223511051) [](ancestors = 223511051)

src/Microsoft.OpenApi/Services/OpenApiSecurityRequirementComparer.cs
Shwetap05: I will have a separate PR to compare scopes

---
In reply to: [224182344](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r224182344) [](ancestors = 224182344,223900360,223852979,223852204,223516846)

test/Microsoft.OpenApi.Tests/Services/OpenApiSecurityRequirementComparerTests.cs
Shwetap05: Will cover in another PR

---
In reply to: [223853512](https://github.com/Microsoft/OpenAPI.NET/pull/328#discussion_r223853512) [](ancestors = 223853512,223517115)

PR #329, opened 2018-10-09
Ensure parsing null in "in" in Parameter does not fail

### LINKED ISSUE #309: NullReferenceException
I get an exception because of a parameter with the property "In" because it is null, I spent a lot of time to find the problem, a "NullReferenceException" error does not say anything, it is necessary to at least handle the errors to better identify where the problem is

PerthCharern:
Fix #309

- Ensure that null in "in" in parameter object is parsed to a null Im in the DOM.

- Add validation rule to validate the required fields in Parameter.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Extensions;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Readers.ParseNodes;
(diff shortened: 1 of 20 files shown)


src/Microsoft.OpenApi/Validations/Rules/OpenApiParameterRules.cs
```diff
@@ -0,0 +1,63 @@
...
+            new ValidationRule<OpenApiParameter>(
+                (context, item) =>
+                {
+                    // required
+                    context.Enter("required");
+                    if ( item.In == ParameterLocation.Path && !item.Required )
+                    {
+                        context.CreateError(
+                            nameof(RequiredMustBeTrueWhenInIsPath),
+                            "\"required\" must be true when parameter location is path");
```
Shwetap05: >path [](start = 82, length = 4)

in quotes

src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
```diff
@@ -229,10 +234,10 @@ public static OpenApiParameter LoadParameter(ParseNode node)
             return LoadParameter(node, false);
         }
 
-        public static OpenApiParameter LoadParameter(ParseNode node, bool evenBody)
+        public static OpenApiParameter LoadParameter(ParseNode node, bool loadRequestBody)
         {
             // Reset the local variables every time this method is called.
-            _in = null;
```
Shwetap05: >in = null; [](start = 13, length = 10)

why remove this?

PR #331, opened 2018-10-11
Open api example comparer

Shwetap05:
Open api example comparer

### DIFF
--- a/src/Microsoft.OpenApi/Services/OpenApiAnyComparer.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiAnyComparer.cs
@@ -1,0 +1,77 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System;
+using System.IO;
+using Microsoft.OpenApi.Any;
+using Microsoft.OpenApi.Writers;
+
+namespace Microsoft.OpenApi.Services
+{
+    /// <summary>
+    ///  Defines behavior for comparing properties of <see cref="IOpenApiAny"/>.
+    /// </summary>
+    public class OpenApiAnyComparer : OpenApiComparerBase<IOpenApiAny>
+    {
+        /// <summary>
+        /// Executes comparision against source and target <see cref="IOpenApiAny"/>.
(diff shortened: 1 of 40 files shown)


src/Microsoft.OpenApi/Services/OpenApiServerComparer.cs
```diff
@@ -54,8 +54,6 @@ public class OpenApiServerComparer : OpenApiComparerBase<OpenApiServer>
                 () => comparisonContext
                     .GetComparer<IDictionary<string, OpenApiServerVariable>>()
                     .Compare(sourceServer.Variables, sourceServer.Variables, comparisonContext));
-
-            // To Do compare extensions
```
PerthCharern: Why are we removing these "// To Do compare extensions" comments?

src/Microsoft.OpenApi/Services/OpenApiAnyComparer.cs
```diff
@@ -0,0 +1,77 @@
...
+            var targetStringWriter = new StringWriter();
+            var targetWriter = new OpenApiJsonWriter(targetStringWriter);
+
+            target.Write(targetWriter, OpenApiSpecVersion.OpenApi3_0);
+            var targetValue = targetStringWriter.GetStringBuilder().ToString();
+
+            if (string.IsNullOrWhiteSpace(sourceValue) && string.IsNullOrWhiteSpace(targetValue))
+            {
+                return;
+            }
```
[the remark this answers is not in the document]
Shwetap05: i don't think so

---
In reply to: [225337275](https://github.com/Microsoft/OpenAPI.NET/pull/331#discussion_r225337275) [](ancestors = 225337275)

src/Microsoft.OpenApi/Services/OpenApiServerComparer.cs
Shwetap05: i will be tracking this work as an issue on github instead of adding these ToDo

---
In reply to: [225337695](https://github.com/Microsoft/OpenAPI.NET/pull/331#discussion_r225337695) [](ancestors = 225337695)

PR #339, opened 2018-10-25
fixes#330-Yield validation err when an invalid security req key found

senthilkumarmohan:

PR #341, opened 2018-10-30
fixes#330-Yield validation err when an invalid security req key found

PerthCharern:
Merging changes from @senthilkumarmohan in #339 into master branch since it's a bug fix.


senthilkumarmohan:
Thanks @Shwetap05, @PerthCharern

PR #343, opened 2018-10-31
Parsing IOpenApiAny should respect format and type

### LINKED ISSUE #306: String parameter get a OpenApiDateTime type
Hi,

I tried to parse a document with a parameter with this properties:

in: "query",
name: "version",
required: true,
default: "2018-07-13",
pattern: "^\d{4}-\d{2}-\d{2}$",
type: "string",
description: "Release date of the API version"

This parameter got the type OpenApiDateTime although it should be a string.

Thanks

PerthCharern:
Fix #306

- Parse IOpenApiAny with respect to `type` and `format` This handles parsing `default`, `enum`, `example`, and `examples` in both V2 and V3.

- Validation rules added to validate whether there is a mismatch between the given type and the data itself.

- The code is quite involved given that the schema type/format information may not reside at the same level as the Any itself (e.g. the Any deep down in `examples` field in media type needs information form `schema` field).

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
@@ -147,7 +147,7 @@ namespace Microsoft.OpenApi.Readers.V2
             new AnyFieldMap<OpenApiParameter>
             {
                 {
-                    "default",
+                    OpenApiConstants.Default,
                     new AnyFieldMapParameter<OpenApiParameter>(
                         p => p.Schema.Default,
                         (p, v) => p.Schema.Default = v,
@@ -159,7 +159,7 @@ namespace Microsoft.OpenApi.Readers.V2
             new AnyListFieldMap<OpenApiParameter>
             {
                 {
-                    "enum",
+                    OpenApiConstants.Enum,
                     new AnyListFieldMapParameter<OpenApiParameter>(
                         p => p.Schema.Enum,
                         (p, v) => p.Schema.Enum = v,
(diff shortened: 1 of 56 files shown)


PerthCharern:
> 

This rule is actually a bit too strict since the spec allows example to be represented as string for every type:

"To represent examples of media types that cannot naturally be represented in JSON or YAML, a string value can contain the example with escaping where necessary."

In YAML, anything can be a string, so anything is technically valid. In JSON, we should actually be able to tell if the given value is a string, but we don't have that info once we read the node with GetScalarValue.

My proposal would be to not add this rule and not perform any validation on the default/example/enum values. Otherwise, we are enforcing something the spec doesn't require. 

The ideal solution would be for the OpenApiAnyConverter and this validator to understand all the context (media type, data type, data format), but that is way more involved, and I think we can tackle that in a separate PR. 

---
Refers to: src/Microsoft.OpenApi/Validations/Rules/OpenApiSchemaRules.cs:276 in afd19c9. [](commit_id = afd19c927a1b5b2a3ab49e5d9138ac248c69f1f0, deletion_comment = False)

PerthCharern:
This changeset turned out to be much larger than I originally expected, as I kept going down the rabbit hole after discovering more and more nuances from the spec and the V2/V3 discrepancies. 

@darrelmiller It's ready for review now. Apologies for the PR size - on the upside, half of the files are tests and the remaining, except a few files, contain straightforward changes, so hopefully this is not too bad to review.

src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
```diff
@@ -149,6 +143,30 @@ internal static partial class OpenApiV2Deserializer
...
+                        (p, v) => p.Schema.Default = v,
+                        p => p.Schema)
+                }
+            };
+
+        private static readonly AnyListFieldMap<OpenApiParameter> _parameterAnyListFields =
+            new AnyListFieldMap<OpenApiParameter>
+            {
+                {
+                    "enum",
```
darrelmiller: Probably would be good to have these as constants

PerthCharern:
Thanks for the review @darrelmiller!

PR #366, opened 2018-12-07
Supporting all header fields inside components/headers

VitaliyKurokhtin:
Specification says HeaderObject is a ParameterObject with some minor differences: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#headerObject

components/headers at the same time is defined as Map[string, Header Object \| Reference Object]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#componentsObject

which leads me to believe that components/headers should support full set of properties allowed for ParameterObject with exception of "name" and "in"

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V3/OpenApiHeaderDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V3/OpenApiHeaderDeserializer.cs
@@ -34,6 +34,12 @@ namespace Microsoft.OpenApi.Readers.V3
                 }
             },
             {
+                "allowEmptyValue", (o, n) =>
+                {
+                    o.AllowEmptyValue = bool.Parse(n.GetScalarValue());
+                }
+            },
+            {
                 "allowReserved", (o, n) =>
                 {
                     o.AllowReserved = bool.Parse(n.GetScalarValue());
@@ -46,11 +52,29 @@ namespace Microsoft.OpenApi.Readers.V3
                 }
             },
             {
+                "explode", (o, n) =>
+                {
+                    o.Explode = bool.Parse(n.GetScalarValue());
+                }
+            },
+            {
                 "schema", (o, n) =>
                 {
                     o.Schema = LoadSchema(n);
                 }
-            }
+            },
+            {
+                "examples", (o, n) =>
+                {
+                    o.Examples = n.CreateMap(LoadExample);
+                }
+            },
+            {
+                "example", (o, n) =>
+                {
+                    o.Example = n.CreateAny();
+                }
+            },
         };
 
         private static readonly PatternFieldMap<OpenApiHeader> _headerPatternFields = new PatternFieldMap<OpenApiHeader>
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi.Readers/V3/OpenApiHeaderDeserializer.cs
```diff
@@ -45,12 +51,30 @@ internal static partial class OpenApiV3Deserializer
...
+                    o.Explode = bool.Parse(n.GetScalarValue());
+                }
+            },
             {
                 "schema", (o, n) =>
                 {
                     o.Schema = LoadSchema(n);
                 }
-            }
+            },
```
PerthCharern: This is still missing the property "content". 
```suggestion
                },
                {
                    "content", (o, n) =>
                    {
                        o.Content = n.CreateMap(LoadMediaType);
                    }
                },
```

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiDocumentTests.cs
```diff
@@ -1204,5 +1204,15 @@ public void GlobalSecurityRequirementShouldReferenceSecurityScheme()
...
             }
         }
+
+        [Fact]
+        public void HeaderParameterShouldAllowExample()
+        {
+            using (var stream = Resources.GetStream(Path.Combine(SampleFolderPath, "apiWithFullHeaderComponent.yaml")))
+            {
+                var openApiDoc = new OpenApiStreamReader().Read(stream, out var diagnostic);
+                Assert.Equal(0, diagnostic.Errors.Count);
```
PerthCharern: The test is a bit too loose. I think it will pass even without adding the new properties in the deserializer (those properties will simply be ignored.)

Could you build the object and compare it against your object "openApiDoc"? See the ParseModifiedPetStoreDocumentWithTagAndSecurityShouldSucceed test for example.

PerthCharern:
Thank you for your contribution @VitaliyKurokhtin 😄 

I agree with your reading of the spec. Our Header deserializer should recognize all the properties in the parameter object except name and in.

Could you help fix the test a bit? - and I'll help check this in.

VitaliyKurokhtin:
Test improved.

src/Microsoft.OpenApi.Readers/V3/OpenApiHeaderDeserializer.cs
VitaliyKurokhtin: Spec does say "content" should be allowed, but "content" is a map keyed of media-type, not sure if concept of media-types is applicable to headers. I feel spec is loose here.

PerthCharern: That's fair. I'm not certain how it will be used either, but I was just reading off the spec. We can leave this be.

PerthCharern:
Thanks for writing this @VitaliyKurokhtin!

PR #368, opened 2018-12-13
Duplicate response processing causes schema corruption

VitaliyKurokhtin:


PerthCharern:
@VitaliyKurokhtin 

Any chance you have an example handy to illustrate what exactly is turning out incorrectly? 

I believe @darrelmiller added this change to allow reading `produces` if it comes after `responses`, but seems like it's causing some kind of regression. (#267)

VitaliyKurokhtin:
I'll add a test, but in short the problem is that ProcessProduces fetches schema by context.GetFromTempStorage<OpenApiSchema>(TempStorageKeys.ResponseSchema) which is, if responses are already processed, the last schema. And that is the schema that gets assigned to all responses. I'm not sure how current implementation helps in supporting global produces, since produces is fetched from temp storage as well and If that information is available after operation and its responses are parsed, it must be available while they're being parsed as well. I'll talk to @darrelmiller.

darrelmiller:
I tested this before and after the fix and was able to repro the problem and confirm the fix solves it.  Much appreciated @VitaliyKurokhtin !

test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
```diff
@@ -12,6 +12,10 @@
         <SignAssembly>true</SignAssembly>
         <AssemblyOriginatorKeyFile>..\..\src\Microsoft.OpenApi.snk</AssemblyOriginatorKeyFile>
     </PropertyGroup>
+    <ItemGroup>
```
PerthCharern: Could you remove this?

PerthCharern:
Thanks @darrelmiller & @VitaliyKurokhtin 

@VitaliyKurokhtin Could you help resolve the merge conflict by removing all the <None Remove> tags and verify that there is no impact on your code and tests? Once you confirm that, I'll check in the code. Thank you for this fix :)

VitaliyKurokhtin:
Sure, done!

tobindh:
There is an issue with this fix if there are multiple produces. It clears the temp storage for the responseSchema in temp storage inside of the produces foreach, so if there are multiple content types (i.e. xml and json), only the first content type gets the schema assigned.  Shouldn't this clear the temp storage outside of the produces foreach?

darrelmiller:
@tobindh Moving the clear to outside the Produces loop solves the problem with multiple produces, but it then breaks the issue with multiple responses.  Not sure why.  I will continue investigating.

VitaliyKurokhtin:
@tobindh fixed by #398

PR #376, opened 2019-01-13
Fix#375 added schema rules for discriminator and required fields

senthilkumarmohan:
Hi, Please can you review this PR and share your feedback - let me know if I have missed something.

Thanks


senthilkumarmohan:
Hi Darrel, Thanks for your response. The below text from "Schema Object" section of spec covers this point. Could you please review and share your thoughts?

> When used, the discriminator will be the name of the property that decides which schema definition validates the structure of the model. As such, the discriminator field MUST be a required field. 

Happy to implement any changes if required.

Thanks

darrelmiller:
Ok, thanks.  I didn't see that wording.  However, even within JSON Schema, a required property does not have to be defined within the properties collection.  JSON schema rules are evaluated independently against the object being tested.  A required property just says that the property must be present in the object, it does not say the schema needs to define the property.

PerthCharern:
@senthilkumarmohan Feel free to send out an updated version. Once you make the following changes, I can sign off:

1) REMOVE the check for required property in properties.
2) REMOVE the check for propertyName in discriminator to be present in properties.
3) KEEP the check for propertyName in discriminator to be present in `required` list.

darrelmiller:
@PerthCharern Yes, I think the wording in the OAS spec adds the constraint that schema required property must contain the discriminator propertyName.

There has been an effort initiated to try and define a constrained JSON Schema that does a better job of defining models rather than just validating them.  There's just nobody to drive that effort at the moment.

This brings me to a related issue.  I ran into a bunch of errors the other day with the Graph OpenAPI document because a rule was generating errors that the examples did not match the schema.  This is also just a SHOULD in the spec.  I ended up removing that rule for my scenario (which was harder than it should have been), but I'm starting to wonder again if we should distinguish between errors and warnings.  Errors would be real spec violations and warnings could be best practice guidances.  This would allow us to make the above scenario a warning.

senthilkumarmohan:
@PerthCharern , @darrelmiller Thanks for your time. I will make the suggested changes.

senthilkumarmohan:
Hi @PerthCharern, I have made the suggested changes, please can you review?

Thanks.

PerthCharern:
Thanks for the PR @senthilkumarmohan. I'll check this in.

@darrelmiller  - You are right about the example matching schema being at a SHOULD level. I overlooked this. I think it makes sense to have a warning, given that the spec itself has two levels of restrictions as well (SHOULD vs MUST).

PR #381, opened 2019-02-02
Enhance OpenApiWalker to visit references

darrelmiller:
The walker currently sees references and does not visit the object unless it is a self-reference(i.e. a component).  This prevents walking reused components multiple times.  However, there was no Visit overload for the reference itself.  This PR adds a visit for an object that implements IReferenceable and is actually referenced from somewhere in the main document.

The implementation was straightforward for most scenarios however it was necessary to add some special case code to handle when a reference was only within the definition of a component.  These references were not visited because of the check for _inComponents.  I had to refine this condition to only avoid walking references for the actual component self-reference, not for all the objects within the component.  I suspect there may be a cleaner way of doing this but I have a specific need for this capability and the current solution works.  The ugly parts are all internal and can be changed without breaking anything.

The scenario that this is being used for is to take a large OpenAPI description and extract some subset of paths from it.  When moving the paths over to a new document it is essential to bring across all referenced components.  This visitor enhancement allows me to find out what things the paths being copied reference.  

I also updated the build file because dotnet build seems to do a better job of being on the path than msbuild does.


src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -1057,7 +1133,12 @@ private void Walk(string context, Action walk)
         /// </summary>
         private bool IsReference(IOpenApiReferenceable referenceable)
         {
-            return referenceable.Reference != null && !_inComponents;
+            var isReference = referenceable.Reference != null && !_inComponents;
+            if (isReference)
+            {
+                Walk(referenceable);
+            }
```
PerthCharern: This Walk should be done by each of the callers. It's a bit weird to have IsReference function do something rather than just returning the boolean.

PerthCharern:
>             if (callback == null)

check IsReference here too?

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiWalker.cs:337 in 2ac6464. [](commit_id = 2ac64648d8354174c73b624c2cc6579037b9df15, deletion_comment = False)

PerthCharern:
>                         Walk(item.Key, () => Walk(item.Value));

Per our discussion, pass in a boolean here to indicate it is a component instead of repeatedly exiting/entering components.

---
Refers to: src/Microsoft.OpenApi/Services/OpenApiWalker.cs:120 in 2ac6464. [](commit_id = 2ac64648d8354174c73b624c2cc6579037b9df15, deletion_comment = False)

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
darrelmiller: I agree, however lifting this to the calling function would cause all the code to be lifted.  I renamed it to ProcessAsReference instead.  If the element is processed as a Reference then it will return true and no further processing of the element will happen.  By renaming the function we avoid the nastiness of having a isXXX method have a sideeffect.

darrelmiller:
@PerthCharern The code is much prettier now, and it still works :-)

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -117,7 +112,7 @@ internal void Walk(OpenApiComponents components)
                 {
                     foreach (var item in components.Schemas)
                     {
-                        Walk(item.Key, () => Walk(item.Value));
+                        Walk(item.Key, () => Walk(item.Value, true));
```
PerthCharern: >true [](start = 62, length = 4)

Can we use named parameter for code clarity?

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
PerthCharern: I see. This new naming sounds good to me.

---
In reply to: [259667050](https://github.com/Microsoft/OpenAPI.NET/pull/381#discussion_r259667050) [](ancestors = 259667050)

PerthCharern:
This looks great! :D

---
In reply to: [466860001](https://github.com/Microsoft/OpenAPI.NET/pull/381#issuecomment-466860001) [](ancestors = 466860001)

test/Microsoft.OpenApi.Tests/Walkers/WalkerLocationTests.cs
```diff
@@ -150,6 +152,96 @@ public void WalkDOMWithCycles()
...
+            };
+
+            var testHeader = new OpenApiHeader() {
+                                Schema = derivedSchema,
+                                Reference = new OpenApiReference()
+                                {
+                                    Id = "test-header",
+                                    Type = ReferenceType.Header
+                                },
+                                UnresolvedReference = false
```
PerthCharern: nit, the indentation is way off here :)

chinigo:
Hi folks. Wanted to ask a question related to the convo in this thread before spinning out a new issue.

My use case sounds similar to yours:
> The scenario that this is being used for is to take a large OpenAPI description and extract some subset of paths from it. When moving the paths over to a new document it is essential to bring across all referenced components.

I'm trying to produce two versions of an `OpenApiDocument`, one for internal users at my employer, and another for external integration partners. Internal users have access to certain endpoints that external partners don't.

I start by pruning the paths that the external users don't have access to, and then I want to prune all the `OpenApiComponents` that are "orphaned." So I figured† I'd make an `OpenApiVisitorBase` subclass that would mark each referenced component it encountered, and then prune all the unencountered components in a second step.

The problem is that my schemata have recursive references (which I believe is the default behavior of [Swashbuckle](https://github.com/domaindrivendev/Swashbuckle.AspNetCore)), and the `void Walk(OpenApiComponents components)` method intentionally skips these.

So, given this input OpenAPI doc:
```
{
  "openapi": "3.0.1",
  "paths": {
    "/public_endpoint" : {
      "get": { 
        …,
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/PublicResponse"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "PublicSchema": {
        "properties": {
          "fooProp": {
            "type": "array",
            "items": "$ref": "#/components/schemas/NestedPublicItem"
          }
        }
      },
      "NestedPublicItem": {
        "properties": {
          "barProp": {
            "type": "string"
          }
        }
      },
      "OrphanedPrivateSchema": {
        "properties": {
          "bazProp": {
            "type": "string"
          }
        }
      }
    }
  }
}
```

I'd need to include both the `PublicSchema` and the `NestedPublicItem` in the output `OpenApiDocument`, but prune the `OrphanedPrivateSchema`. The problem is that the Visitor never gets invoked with the `NestedPublicItem`.

So my question is: why don't we walk through the components recursively?

I guess, in contrast to inlined components, the referenced components introduce the possibility of circular references, so you'd need to implement cycle detection. That's a bit more complicated than the existing implementation, but it should be possible, right?

Would y'all be open to a PR adding this feature (+ tests), or is this an intentional design decision? Should we allow implementers to choose *whether* they want to recurse through components?

Thanks!
~Chinigo

† Actually, my first thought was "Cool, I finally get to write a GoF Visitor implementation!" My second thought was "ahh, @darrelmiller beat me to it."

darrelmiller:
Hey @chinigo  my understanding is that we do visit references in components.  I'm pretty sure if we didn't then the CopyReferences mechanism I use here https://github.com/microsoftgraph/msgraph-openapi-introspection/blob/master/OpenAPIService/OpenApiService.cs#L259  would not work.  
We special case references that are top level elements in the components section so that we do traverse into them.

PR #382, opened 2019-02-02
Cause scalar string with colon to be quoted #380

### LINKED ISSUE #380: Terminal colon in string properties produces invalid YAML
String property values that end with a colon are not quoted/escaped when writing YAML. The code below produces the following output, which seems to be invalid (not accepted by editor.swagger.io):

```
title: Title:
description: Description
```

```csharp
using System;

namespace ConsoleApp2
{
    class Program
    {
        static void Main(string[] args)
        {
            var writer = new OpenApiYamlWriter(Console.Out);
            var schema = new OpenApiSchema
            {
                Title = "Title:",
                Description = "Description"
            };
            schema.SerializeAsV3(writer);
            Console.WriteLine();
        }
    }
}
```

darrelmiller:
A trailing colon in a string causes the YAML parser to think the scaler is a map key.  This change ensures scalars with colons are quoted.  Arguably, we could allow colons in the middle of strings as long as they are not followed by a space but, this change should be sufficient.  Fixes #380

### DIFF
--- a/src/Microsoft.OpenApi/Writers/SpecialCharacterStringExtensions.cs
+++ b/src/Microsoft.OpenApi/Writers/SpecialCharacterStringExtensions.cs
@@ -42,8 +42,7 @@ namespace Microsoft.OpenApi.Writers
         // http://www.yaml.org/spec/1.2/spec.html#style/flow/plain
         private static readonly string[] _yamlPlainStringForbiddenCombinations =
         {
-            //": ",
-            ":",  // Even though colons can be allowed, this change prevents unquoted trailing colon which causes a problem
+            ": ",
             " #",
 
             // These are technically forbidden only inside flow collections, but
@@ -53,6 +52,13 @@ namespace Microsoft.OpenApi.Writers
             "{",
             "}",
             ","
+        };
+
+        // Plain style strings cannot end with these characters.
+        // http://www.yaml.org/spec/1.2/spec.html#style/flow/plain
+        private static readonly string[] _yamlPlainStringForbiddenTerminals =
+        {
+            ":"
         };
 
         // Double-quoted strings are needed for these non-printable control characters.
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi/Writers/SpecialCharacterStringExtensions.cs
```diff
@@ -42,7 +42,8 @@ public static class SpecialCharacterStringExtensions
         // http://www.yaml.org/spec/1.2/spec.html#style/flow/plain
         private static readonly string[] _yamlPlainStringForbiddenCombinations =
         {
-            ": ",
+            //": ",
```
PerthCharern: nit, remove this commented-out line

src/Microsoft.OpenApi/Writers/SpecialCharacterStringExtensions.cs
```diff
@@ -42,7 +42,8 @@ public static class SpecialCharacterStringExtensions
...
+            ":",  // Even though colons can be allowed, this change prevents unquoted trailing colon which causes a problem
```
PerthCharern: Even though colons not followed by a space can be allowed, this change prevents unquoted trailing colon which causes a problem.

PerthCharern:
@darrelmiller

- Though this works, this will add unnecessary single quotes for a frequent usage of URL (http://example.com) since it contains a colon in the scheme. Can we create a different list of "forbidden characters at end of string" and include colon in it instead? That would add the single quotes for this specific case while still keeping the URL case as normal.

- Unit tests

darrelmiller:
@PerthCharern That makes sense.  I try and find time to work on that this week, although that might be a bit optimistic :-)

darrelmiller:
@PerthCharern I updated this PR to check for colon as a terminal character.  I have been unable to find other characters that cause a problem.  I added a test for the scenario

Issue #383, opened 2019-02-18
ValidateOneOfDiscriminator errors


michivogler:
I'm currently trying to generate an OpenAPI documentation using this library. When I try to add a discriminator in combination with `oneOf` I get a ton of errors after validation for the `ValidateOneOfDiscriminator` rule and I can't figure out what exactly is wrong.

Error message examples:
```
Composite Schema EmailAddress must contain property specified in the discriminator Type. [#/paths/~1Addresses/get/responses/200/content/application~1json/schema/items/oneOf]

Composite schema EmailAddress must contain property specified in the discriminator Type in the required field list. [#/paths/~1Addresses/get/responses/200/content/application~1json/schema/items/oneOf]
```

When I serialize the `OpenApiDocument` and copy it to https://editor.swagger.io/ everything seems to be valid. Even if I use the `OpenApiStreamReader` and validate it again there are no errors and after serializing it again there's no diff between the two documents.

Any idea what the problem might be?

Here's a short version of my openapi.yaml
```
openapi: 3.0.1
info:
  title: My API
  version: v1
paths:
  /Addresses:
    get:
      tags:
        - Addresses
      operationId: AddressesGetAll
      parameters:
        - name: pageIndex
          in: query
          schema:
            type: string
        - name: pageSize
          in: query
          schema:
            type: string
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  oneOf:
                    - $ref: '#/components/schemas/EmailAddress'
                    - $ref: '#/components/schemas/PostalAddress'
                    - $ref: '#/components/schemas/WebAddress'
                  discriminator:
                    propertyName: Type
        '401':
          description: Unauthorized
        '403':
          description: Forbidden
components:
  schemas:
    WebAddress:
      required:
        - Type
        - Uri
      type: object
      properties:
        VerifiedDate:
          type: string
          format: date-time
          readOnly: true
        Version:
          type: number
          format: int32
          readOnly: true
        CreatedTime:
          type: string
          format: date-time
          readOnly: true
        UpdatedTime:
          type: string
          format: date-time
          readOnly: true
        IsVerified:
          type: boolean
          readOnly: true
        Type:
          type: string
          readOnly: true
        VerificationCode:
          type: string
        Guid:
          type: string
          format: uuid
          readOnly: true
        IsAdvertisingAllowed:
          type: boolean
        Uri:
          type: string
        DisplayName:
          type: string
          readOnly: true
[...]

PR #384, opened 2019-02-18
Bugfix#383 remove check for existence of discriminator property in the property list

senthilkumarmohan:


senthilkumarmohan:
#383 is due to a validation rule added against anyOf/oneOf lists. I have removed this check as this check is not a MUST as per the spec. ref #376. Please can you review?

PerthCharern:
>         }

@darrelmiller @senthilkumarmohan 

This is a bit tricky. The spec says this "The expectation now is that a property with name {discriminator.propertyName} MUST be present in the response payload". 

Should we interpret "MUST be present in the response payload" to mean "MUST be a required field"? There seems to be a nuanced difference here, whether intentionally or not. The original sentence in the spec talks about the response payload only and doesn't only say anything about the property or the field itself. (Compare this with another less ambiguous wording "As such, the discriminator field MUST be a required field.") 

I suggest we remove this check altogether.


---
Refers to: src/Microsoft.OpenApi/Validations/Rules/OpenApiSchemaRules.cs:139 in fcd74c4. [](commit_id = fcd74c49dfc2e11f3bfccc4f28419d691c637ea6, deletion_comment = False)

PerthCharern:
>         }

Which means removing line 83-119 above too.

---
In reply to: [466162203](https://github.com/Microsoft/OpenAPI.NET/pull/384#issuecomment-466162203) [](ancestors = 466162203)

---
Refers to: src/Microsoft.OpenApi/Validations/Rules/OpenApiSchemaRules.cs:139 in fcd74c4. [](commit_id = fcd74c49dfc2e11f3bfccc4f28419d691c637ea6, deletion_comment = False)

senthilkumarmohan:
I agree, will update the PR with suggested changes.

senthilkumarmohan:
Hi @PerthCharern, I have updated the PR as suggested. Please can you review?

PerthCharern:
Thanks for your continuous contribution @senthilkumarmohan!

Issue #386, opened 2019-02-27
Binary in Any type writing failed


xuzhg:
If i have a OpenApiBinary object, it's not supported now in the writing the binary value.

The problem is at https://github.com/Microsoft/OpenAPI.NET/blob/master/src/Microsoft.OpenApi/Writers/OpenApiWriterAnyExtensions.cs#L173

It directly uses the value (its type is byte[]), and the underlying writer can't write the `byte[]` type.

PerthCharern:
Related issue: #379. 

This still hasn't been completely fixed I think. Reposting my comment from the PR since it already got merged: https://github.com/Microsoft/OpenAPI.NET/pull/387#discussion_r265248942

> My take would be:
> 
> Store OpenApiBinary as byte[]. The reader converts the string into byte array assuming it knows each character is a byte. I think in C#, this can be done by Encoding.ASCII.GetBytes(string);. The writer does the reverse: Encoding.ASCII.GetString(bytes);
> 
> We should also store OpenApiByte as byte[]. The reader converts the string into byte array using Convert.FromBase64String, and the writer converts the byte array into base64 string using Convert.ToBase64String.
> 
> This way we keep our principle of storing the underlying meaningful type that we does with Date/DateTime, and we allow the reader/writer to handle the (de-)serialization properly.

darrelmiller:
@PerthCharern   I'm ok with storing OpenApiBinary as a byte array, but shouldn't we do Encoding.UTF8.GetBytes(string) considering JSON is UTF-8 and probably 99.99% of yaml files are also UTF-8.  I agree with your solution to OpenApiByte.

PerthCharern:
Agree with you. The UTF-8 encoding makes more sense.

All that said, this Binary type seems pretty useless since in the doc it's just serialized as a normal string, representing itself - the only benefit our model is offering is converting and storing it into byte[]. 

From the spec standpoint, is there really any difference when someone specifies a string as string without format vs string with Binary format? One thing I can think of is it is EXPLICITLY saying that this is indeed a string that should be taken as is, not a string that represents bytes (like the Byte format).

darrelmiller:
@PerthCharern From the spec standpoint, the format means nothing.  It is purely a hint to tooling on how to interpret the data.  IT doesn't affect the wire format, or how the HTTP request is made.  Tooling can choose to validate data based on the format, but there is no requirement in the spec for this.
[...]

PR #387, opened 2019-02-27
write the binary to base64string and read it as byte[]

xuzhg:
This PR trys to fix the issue at #386 .

Before this change, the library can't write the "byte[]" types.

### DIFF
--- a/src/Microsoft.OpenApi/Writers/OpenApiWriterAnyExtensions.cs
+++ b/src/Microsoft.OpenApi/Writers/OpenApiWriterAnyExtensions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
+using System;
 using System.Collections.Generic;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Exceptions;
@@ -170,7 +171,14 @@ namespace Microsoft.OpenApi.Writers
 
                 case PrimitiveType.Binary:
                     var binaryValue = (OpenApiBinary)primitive;
-                    writer.WriteValue(binaryValue.Value);
+                    if (binaryValue == null)
+                    {
+                        writer.WriteNull();
+                    }
+                    else
+                    {
+                        writer.WriteValue(Convert.ToBase64String(binaryValue.Value));
+                    }
                     break;
 
                 case PrimitiveType.Boolean:
--- a/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
+++ b/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
@@ -195,13 +195,17 @@ namespace Microsoft.OpenApi.Readers.ParseNodes
 
             if (type == "string" && format == "byte")
             {
-                if ( byte.TryParse(value, NumberStyles.Any, CultureInfo.InvariantCulture, out var byteValue))
+                if (byte.TryParse(value, NumberStyles.Any, CultureInfo.InvariantCulture, out var byteValue))
                 {
                     return new OpenApiByte(byteValue);
                 }
             }
 
-            // TODO: Parse byte array to OpenApiBinary type.
+            // binary
+            if (type == "string" && format == "binary")
+            {
+                return new OpenApiBinary(Convert.FromBase64String(value));
+            }
 
             if (type == "string" && format == "date")
             {
--- a/test/Microsoft.OpenApi.Tests/Models/OpenApiExampleTests.cs
+++ b/test/Microsoft.OpenApi.Tests/Models/OpenApiExampleTests.cs
@@ -29,7 +29,8 @@ namespace Microsoft.OpenApi.Tests.Models
                             new OpenApiObject
                             {
                                 ["href"] = new OpenApiString("http://example.com/1"),
-                                ["rel"] = new OpenApiString("sampleRel1")
+                                ["rel"] = new OpenApiString("sampleRel1"),
+                                ["binary"] = new OpenApiBinary(new byte[] { 1, 2, 3 })
                             }
                         }
                     },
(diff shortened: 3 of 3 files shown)


src/Microsoft.OpenApi/Writers/OpenApiWriterAnyExtensions.cs
```diff
@@ -170,7 +171,14 @@ private static void WritePrimitive(this IOpenApiWriter writer, IOpenApiPrimitive
...
-                    writer.WriteValue(binaryValue.Value);
+                    if (binaryValue == null)
+                    {
+                        writer.WriteNull();
+                    }
+                    else
+                    {
+                        writer.WriteValue(Convert.ToBase64String(binaryValue.Value));
+                    }
                     break;
```
PerthCharern: @darrmi@microsoft.com  @saxu@microsoft.com 

I'm not sure if this is correct. 

- The spec says that "byte" type is base64 encoded characters. This means that our OpenApiByte which stores one byte is incorrect. Despite the singular name "byte", it seems to refer to multiple bytes. A reading from swagger.io also seems to suggest this: [https://swagger.io/docs/specification/data-models/data-types/#file](https://swagger.io/docs/specification/data-models/data-types/#file)

- Binary is defined as a sequence of octets, so it can't be Base64 string (since Base64 digit is 6-bit). I'm a bit at a loss at how to interpret this.

xuzhg: @PerthCharern in my understand, from the spec it says binary is defined as a sequence of octets, while "byte" is used to represent one octet, meanwhile, one byte is base64 encoded characters (here "s" means one byte maybe more than one character base64 encode, so, binary is a sequence of base64 encoded characters. however, i leave it for @darrelmiller to give us more inputs.

darrelmiller: I really dislike the "format" property because it isn't well defined in the spec.  I do believe that format "byte" was intended to indicate a base64 encoded string.  I think the 'binary' format is basically saying that you should treat the UTF-8 code units as octets.  Not sure where that would be useful though.  

The "byte" format isn't representing a single byte, because if that were the case we would use the underlying integer type.

darrelmiller: I'll reach out to other folks to on the TSC to confirm.

xuzhg: @darrelmiller & @PerthCharern Is there any update for this? Thanks.

darrelmiller: @xuzhg I got affirmation that a string that with a format binary should just be written out as a string, because what else could it be?  I agree it isn't very useful.

xuzhg: @darrelmiller Ok. Let me merge that. @PerthCharern, if you have further concern, let's discuss offline. Thanks.

darrelmiller: @xuzhg Sorry, I guess I was not clear.  Format=Byte should be base64 encoded.  Format=Binary should just be serialized as a string.  Doesn't your change serialize format=binary as base64?

xuzhg: @darrelmiller what kind of string for Format=Binary?

darrelmiller: @xuzhg I'm not sure I understand the question.  A UTF-8 string?  From an implmentation perspective, I think our reader/writer can basically ignore the format property when its value is binary.  If someone was then consuming that string value they would have the option of treating it like a array of chars and each char is a byte.

PerthCharern: @xuzhg Any chance you could help with creating a new PR that handles byte as base64 encoded and binary as just a simple string?

xuzhg: @darrelmiller Sorry, I am not clear for my question. my original question is "how to convert a binary to/from a string in the OpenAPI.Net library?" 

Now, the underlying type of OpenApiBinary is "byte[]", if I create a OpenApiBinary to write, the library gives me the error. So, What's the built-in conversion in OpenApi.NET for our customers to use?

PerthCharern: My take would be:

- Store OpenApiBinary as byte[]. The reader converts the string into byte array assuming it knows each character is a byte. I think in C#, this can be done by `Encoding.ASCII.GetBytes(string);`. The writer does the reverse: `Encoding.ASCII.GetString(bytes);`

- We should also store OpenApiByte as byte[]. The reader converts the string into byte array using `Convert.FromBase64String`, and the writer converts the byte array into base64 string using `Convert.ToBase64String`.

This way we keep our principle of storing the underlying meaningful type that we does with Date/DateTime, and we allow the reader/writer to handle the (de-)serialization properly.

@xuzhg  @darrelmiller Thoughts?

xuzhg: @PerthCharern 

So, do you think we will do like below:

1. Convert stringn to Binary?
`StringToBinary("fluxbytes.com")`

`// returns "01100110011011000111010101111000011000100111100101110100011001010111001100101110011000110110111101101101"`

2. Convert Binary string to string as below:

`BinaryToString("01100110011011000111010101111000011000100111100101110100011001010111001100101110011000110110111101101101");`

`// returns "fluxbytes.com"`

PerthCharern: I don't think so. I'm thinking of the logic that you already have in this PR, but with the right conversion function.

GetBytes instead of FromBase64String
GetString instead of ToBase64String

PR #391, opened 2019-03-14
Change OpenApiByte & OpenApiBinary

xuzhg:
1. Change OpenApiByte
2. Change OpenApiBinary
3. Chang the OpenApiByte & OpenApiBinary serialization/desalinization.

Below are the discussions:

@darrelmiller 
Admittedly, the schema is unnecessary, but there are examples in the spec that show this.  In this scenario the format: binary is used to indicate that there is no special serialization that is performed on the bytes.  They are raw.  If you try and present them as a string, they might not create printable text.

This is not a problem in 90% of cases because for the definition of an API you don’t need to include any data.  The spec says the parameter contains binary data and tooling knows not to do any processing on it.  That’s obviously potentially tricky if it is in a URL parameter! However, for us, the only place it causes a problem is in examples.   In examples it is important that the user chooses example data that can be serialized into printable strings.  Or not include a sample for binary data as it is somewhat redundant.

This brings me back to my suggestion that for examples of binary data, if they are present, we should just assume that they can successfully be rendered as UTF8 strings and that they will roundtrip.  If it doesn’t work, then the user should choose an example of binary data that can be stored in UTF-8 because that’s a limitation of the OpenAPI spec.

This will ensure that round tripping works just fine.  However, if someone manually creates an OpenAPI document via the DOM and assigns an arbitrary set of bytes that cannot be represented as UTF-8 characters then it will not round trip.  Ironically, this is good because it will discourage people from using format “binary”.  If we “fix” this bug to make it serialize properly then we will hide the problem and people will expect other OpenAPI implementations to process the issue properly.  But we don’t control how other implementations will interpret binary.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
+++ b/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Globalization;
 using System.Linq;
+using System.Text;
 using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Exceptions;
 using Microsoft.OpenApi.Models;
@@ -195,16 +196,23 @@ namespace Microsoft.OpenApi.Readers.ParseNodes
 
             if (type == "string" && format == "byte")
             {
-                if (byte.TryParse(value, NumberStyles.Any, CultureInfo.InvariantCulture, out var byteValue))
+                try
                 {
-                    return new OpenApiByte(byteValue);
+                    return new OpenApiByte(Convert.FromBase64String(value));
                 }
+                catch(Exception)
+                { }
             }
 
             // binary
             if (type == "string" && format == "binary")
             {
-                return new OpenApiBinary(Convert.FromBase64String(value));
+                try
+                {
+                    return new OpenApiBinary(Encoding.UTF8.GetBytes(value));
+                }
+                catch(Exception)
+                { }
             }
 
             if (type == "string" && format == "date")
--- a/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
+++ b/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
@@ -200,7 +200,7 @@ namespace Microsoft.OpenApi.Readers.ParseNodes
                 {
                     return new OpenApiByte(Convert.FromBase64String(value));
                 }
-                catch(Exception)
+                catch(FormatException)
                 { }
             }
 
@@ -211,7 +211,7 @@ namespace Microsoft.OpenApi.Readers.ParseNodes
                 {
                     return new OpenApiBinary(Encoding.UTF8.GetBytes(value));
                 }
-                catch(Exception)
+                catch(EncoderFallbackException)
                 { }
             }
 
--- a/src/Microsoft.OpenApi/Any/OpenApiByte.cs
+++ b/src/Microsoft.OpenApi/Any/OpenApiByte.cs
@@ -6,12 +6,20 @@ namespace Microsoft.OpenApi.Any
     /// <summary>
     /// Open API Byte
     /// </summary>
-    public class OpenApiByte : OpenApiPrimitive<byte>
+    public class OpenApiByte : OpenApiPrimitive<byte[]>
     {
         /// <summary>
         /// Initializes the <see cref="OpenApiByte"/> class.
         /// </summary>
         public OpenApiByte(byte value)
+            : this(new byte[] { value })
+        {
+        }
+
+        /// <summary>
+        /// Initializes the <see cref="OpenApiByte"/> class.
+        /// </summary>
+        public OpenApiByte(byte[] value)
             : base(value)
         {
         }
--- a/test/Microsoft.OpenApi.Tests/Models/OpenApiExampleTests.cs
+++ b/test/Microsoft.OpenApi.Tests/Models/OpenApiExampleTests.cs
@@ -31,7 +31,8 @@ namespace Microsoft.OpenApi.Tests.Models
                             {
                                 ["href"] = new OpenApiString("http://example.com/1"),
                                 ["rel"] = new OpenApiString("sampleRel1"),
-                                ["binary"] = new OpenApiBinary(new byte[] { 1, 2, 3 })
+                                ["bytes"] = new OpenApiByte(new byte[] { 1, 2, 3 }),
+                                ["binary"] = new OpenApiBinary(new byte[] { 41, 42, 43 })
                             }
                         }
                     },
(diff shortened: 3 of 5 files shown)


src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
```diff
@@ -195,16 +196,23 @@ public static IOpenApiAny GetSpecificOpenApiAny(IOpenApiAny openApiAny, OpenApiS
...
             // binary
             if (type == "string" && format == "binary")
             {
-                return new OpenApiBinary(Convert.FromBase64String(value));
+                try
+                {
+                    return new OpenApiBinary(Encoding.UTF8.GetBytes(value));
+                }
+                catch(Exception)
+                { }
```
PerthCharern: Try/catch unnecessary.

src/Microsoft.OpenApi/Any/OpenApiByte.cs
```diff
@@ -6,12 +6,20 @@ namespace Microsoft.OpenApi.Any
...
+    public class OpenApiByte : OpenApiPrimitive<byte[]>
     {
         /// <summary>
         /// Initializes the <see cref="OpenApiByte"/> class.
         /// </summary>
         public OpenApiByte(byte value)
+            : this(new byte[] { value })
+        {
+        }
+
```
PerthCharern: Remove this

xuzhg: Remove this is a breaking change. We should avoid breaking change.

src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
xuzhg: Leave the uplevel to process the exception?

src/Microsoft.OpenApi/Any/OpenApiByte.cs
PerthCharern: Yes, but this is technically a bug fix since our old model is incorrect

---
In reply to: [265768029](https://github.com/Microsoft/OpenAPI.NET/pull/391#discussion_r265768029) [](ancestors = 265768029)

src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
PerthCharern: Which exception case are we envisioning here?

---
In reply to: [265768186](https://github.com/Microsoft/OpenAPI.NET/pull/391#discussion_r265768186) [](ancestors = 265768186)

xuzhg: GetBytes(..) may throw the following exception:

Exceptions:
        //   T:System.ArgumentNullException:
        //     s is null.
        //
        //   T:System.Text.EncoderFallbackException:
        //     A fallback occurred (see Character Encoding in the .NET Framework for complete
        //     explanation)-and-System.Text.Encoding.EncoderFallback is set to System.Text.EncoderExceptionFallback.

src/Microsoft.OpenApi/Any/OpenApiByte.cs
xuzhg: `public OpenApiByte(byte value)` is not a technically bug. Keeping it also provides a method for customer to use.

src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
PerthCharern: Got it. Let's do this:
- The ArgNullEx one won't be thrown since we already handle it above.
- Can we specifically catch the second type of exception? 

---
In reply to: [265771477](https://github.com/Microsoft/OpenAPI.NET/pull/391#discussion_r265771477) [](ancestors = 265771477)

test/Microsoft.OpenApi.Tests/Models/OpenApiExampleTests.cs
```diff
@@ -119,7 +120,8 @@ public void SerializeAdvancedExampleAsV3JsonWorks()
           {
             ""href"": ""http://example.com/1"",
             ""rel"": ""sampleRel1"",
-            ""binary"": ""AQID""
+            ""bytes"": ""AQID"",
+            ""binary"": "")*+""
```
PerthCharern: >)*+ [](start = 26, length = 3)

For the binary case, can we test with a bit more complicated string (something that's not in ASCII)? 

Instead of putting in the byte[] specifically above for the binary case, you can use the UTF8.GetBytes function to get the right byte array.

src/Microsoft.OpenApi/Any/OpenApiByte.cs
PerthCharern: I guess that's fine. 

---
In reply to: [265771924](https://github.com/Microsoft/OpenAPI.NET/pull/391#discussion_r265771924) [](ancestors = 265771924)

test/Microsoft.OpenApi.Tests/Models/OpenApiExampleTests.cs
xuzhg: I original use the byte[] { 1, 2, 3 }, but, the output string is not displayable. How can I show that?

PerthCharern: Yeah, that's because 1, 2, 3 correspond to some control characters in ASCII (thus non-printable).

I was suggesting something along this line:

- Here in line 124, do something like ""Ñ😻😑♮Í☛oƞ♑😲☇éǋžŁ♻😟¥a´Ī♃ƠąøƩ""
- Then, in line 35, do UTF8.GetBytes("Ñ😻😑♮Í☛oƞ♑😲☇éǋžŁ♻😟¥a´Ī♃ƠąøƩ")

---
In reply to: [265777035](https://github.com/Microsoft/OpenAPI.NET/pull/391#discussion_r265777035) [](ancestors = 265777035)

xuzhg: :) I changed.

src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
xuzhg: Changed.

PerthCharern:
I will sign off, but given that this issue was quite contentious (for lack of a better word), could we get all 3 of us to agree before checking this in? That is, @darrelmiller, could we get your comments and/or sign-off here as well?

darrelmiller:
Thanks @xuzhg for fixing this and thanks @PerthCharern for catching the issue after I had signed off on it!

PerthCharern:
Thanks @darrelmiller! 

I'll merge this in. Thanks @xuzhg for detecting and addressing the issue.

PR #398, opened 2019-04-10
Fix to properly assign schemas to content of all media types of all r…

VitaliyKurokhtin:
It is an error to clean temp schema storage after processing single "produces" item since it's per response. Thus moving it outside by cleaning it only if there are any "produces" items to accomodate for scenario when "produces" is specified after "paths"

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
@@ -136,6 +136,9 @@ namespace Microsoft.OpenApi.Readers.V2
                 ProcessProduces(node.CheckMapNode("responses"), response, node.Context);
             }
 
+            // Reset so that it's not picked up later
+            node.Context.SetTempStorage(TempStorageKeys.OperationProduces, null);
+
             return operation;
         }
 
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
@@ -63,35 +63,39 @@ namespace Microsoft.OpenApi.Readers.V2
         private static void ProcessProduces(MapNode mapNode, OpenApiResponse response, ParsingContext context)
         {
             var produces = context.GetFromTempStorage<List<string>>(TempStorageKeys.OperationProduces) ??
-                context.GetFromTempStorage<List<string>>(TempStorageKeys.GlobalProduces) ?? new List<string>();
+                context.GetFromTempStorage<List<string>>(TempStorageKeys.GlobalProduces);
 
             if (response.Content == null)
             {
                 response.Content = new Dictionary<string, OpenApiMediaType>();
             }
 
-            foreach (var produce in produces)
+            if (produces != null && produces.Count > 0)
             {
-                var schema = context.GetFromTempStorage<OpenApiSchema>(TempStorageKeys.ResponseSchema, response);
-                context.SetTempStorage(TempStorageKeys.ResponseSchema, null, response);
-
-                if (response.Content.ContainsKey(produce) && response.Content[produce] != null)
+                foreach (var produce in produces)
                 {
-                    if (schema != null)
+                    var schema = context.GetFromTempStorage<OpenApiSchema>(TempStorageKeys.ResponseSchema, response);
+
+                    if (response.Content.ContainsKey(produce) && response.Content[produce] != null)
                     {
-                        response.Content[produce].Schema = schema;
-                        ProcessAnyFields(mapNode, response.Content[produce], _mediaTypeAnyFields);
+                        if (schema != null)
+                        {
+                            response.Content[produce].Schema = schema;
+                            ProcessAnyFields(mapNode, response.Content[produce], _mediaTypeAnyFields);
+                        }
+                    }
+                    else
+                    {
+                        var mediaType = new OpenApiMediaType
+                        {
+                            Schema = schema
+                        };
+
+                        response.Content.Add(produce, mediaType);
                     }
                 }
-                else
-                {
-                    var mediaType = new OpenApiMediaType
-                    {
-                        Schema = schema
-                    };
 
-                    response.Content.Add(produce, mediaType);
-                }
+                context.SetTempStorage(TempStorageKeys.ResponseSchema, null, response);
             }
         }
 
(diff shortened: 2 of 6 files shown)


src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
```diff
@@ -63,35 +63,39 @@ internal static partial class OpenApiV2Deserializer
...
-                context.GetFromTempStorage<List<string>>(TempStorageKeys.GlobalProduces) ?? new List<string>();
+                context.GetFromTempStorage<List<string>>(TempStorageKeys.GlobalProduces);
 
             if (response.Content == null)
             {
                 response.Content = new Dictionary<string, OpenApiMediaType>();
             }
 
-            foreach (var produce in produces)
+            if (produces != null && produces.Count > 0)
```
PerthCharern: nit, the >0 check not needed

test/Microsoft.OpenApi.Readers.Tests/Microsoft.OpenApi.Readers.Tests.csproj
```diff
@@ -12,6 +12,9 @@
         <SignAssembly>true</SignAssembly>
         <AssemblyOriginatorKeyFile>..\..\src\Microsoft.OpenApi.snk</AssemblyOriginatorKeyFile>
     </PropertyGroup>
+    <ItemGroup>
```
PerthCharern: Remove these lines

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiDocumentTests.cs
```diff
@@ -238,5 +239,30 @@ public void ShouldParseProducesInAnyOrder()
                 }
             }
         }
+
```
PerthCharern: Can we add one more test case with the mixture of global produces and operation-level produces?

VitaliyKurokhtin: This is already covered by twoResponses test

PerthCharern: I see. 

Looks like the twoResponses test has only `application/json` as the produce type in all places. It would be great to have `application/xml` in one of the operation-level produce to test that there are no issues with such interaction. There's no test that covers such case right now.

Any chance you could help fix the test? It should not be too big of a change.

---
In reply to: [275497172](https://github.com/Microsoft/OpenAPI.NET/pull/398#discussion_r275497172) [](ancestors = 275497172)

VitaliyKurokhtin: Sure

test/Microsoft.OpenApi.Readers.Tests/V2Tests/Samples/twoResponses.json
```diff
@@ -70,13 +71,15 @@
...
+                    "application/json",
+                    "application/xml"
                 ]
             }
         }
     },
     "produces": [
-        "application/json"
+        "application/json",
+        "application/xml"
```
PerthCharern: >       "application/xml" [](start = 1, length = 24)

Sorry I was not clear. Could you remove the `application/xml` from line 82?

Then:

- The check for get and patch should yield the same results (both json and xml)

- The check for post should yield just json.

VitaliyKurokhtin: Sure.

src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
```diff
@@ -136,6 +136,9 @@ internal static OpenApiOperation LoadOperation(ParseNode node)
                 ProcessProduces(node.CheckMapNode("responses"), response, node.Context);
             }
 
+            // Reset so that it's not picked up later
+            node.Context.SetTempStorage(TempStorageKeys.OperationProduces, null);
```
PerthCharern: Is this needed? It's already set in line 110 above

VitaliyKurokhtin: Yes, to clear it our after we exit operation scope. So that when we process global "produces" later on we do not pickup ones stored in temp operation store.

PerthCharern: Got it. I'll investigate more on what to do with line 110, but this should not block this from being checked in.

PerthCharern:
Thanks for the contribution @VitaliyKurokhtin. I'll check this in.

src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs
VitaliyKurokhtin: Both of them are needed. I'd clean other temp storages too (body, form, consumes), but don't want to increase scope of this PR. The general idea is to clean temp storage before operation is parsed, since parsing may read data from there and we don't want to use other operation's data. And we need to clean after as well if we're going to reuse same parsing logic.

In this particular test - two responses - without cleaning temp storage POST operation was being assigned two media types since they're left in temp storage after last operation and ProcessProduces prefers operation data to global.

Another approach here is to store data with scope, i.e. at line 65:

`n.Context.SetTempStorage(TempStorageKeys.OperationProduces, null, o)`

And pass operation reference to ProcessProduces and do:

`context.GetFromTempStorage<List<string>>(TempStorageKeys.OperationProduces, operation)`

This way it also would not be possible for one operation to be contaminated with other operation data. But I'm reluctant to make that change since it requires more thoroughal testing than I can do at the moment.

PerthCharern: The issue is that there are ProcessProduces in two places - in the operation and document deserializers. This leads to the data in temp storage being used incorrectly when ProcessProduces in called in document deserializer.

A couple more alternatives: 

- We technically do not need to clear OperationProduces in LoadOperation I think. We can also do it before we call ProcessProduces in the document deserializer (the same way we call it in line 110 of the operation deserializer here).

- We can make ProcessProduces take another parameter whether it's called from document or operation deserializer. Then, we only use the global produces when it's called from the document. 

- The scope method you propose should work as well with the right implementation.

VitaliyKurokhtin: - We could clear it out before calling ProcessProduces at document level, but that would mean "exposing" internal dependencies of ProcessProduces.

- Would work

- Support for scope is already there, was added before for some other place. Technically should be used throughout, but it's a more broad change.

PR #404, opened 2019-04-27
Serialize explode value properly on parameters with style "form"

mapitman:
When a parameter has the style set to "form" and explode set to false,
the v3 serializer will not write out the value for explode. Swagger UI
then defaults the explode value to true, since it is not specified and
that is the correct default value when the style is set to "form". The
OpenAPI spec for this is right around here:

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-10

I'll be glad to make any changes required to get his accepted.

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiParameter.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiParameter.cs
@@ -177,7 +177,7 @@ namespace Microsoft.OpenApi.Models
             writer.WriteProperty(OpenApiConstants.Style, Style?.GetDisplayName());
 
             // explode
-            writer.WriteProperty(OpenApiConstants.Explode, Explode, false);
+            writer.WriteProperty(OpenApiConstants.Explode, Explode, Style.HasValue && Style.Value == ParameterStyle.Form);
 
             // allowReserved
             writer.WriteProperty(OpenApiConstants.AllowReserved, AllowReserved, false);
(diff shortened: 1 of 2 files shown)


darrelmiller:
Nice catch.  It is this part of the spec that we were not doing properly:

> 
When style is form, the default value is true. For all other styles, the default value is false.

Your changes look good to me.

mapitman:
I signed the agreement, not sure why it keeps saying I haven't.

mapitman:
Looks like it finally checks out that I signed it

mapitman:
There is similar logic in the `OpenApiHeader` class. Would is make sense to implement the same change there?

https://github.com/Microsoft/OpenAPI.NET/blob/5742db6b5f0f8bd9990fbf43a7e03a8b0b6f674a/src/Microsoft.OpenApi/Models/OpenApiHeader.cs#L131

mapitman:
Any chance of merging this sometime soon?

darrelmiller:
@mapitman I will look into this next week.  Sorry for the delay

jgreywolf:
I am interested in this fix as well.  Any update?

darrelmiller:
I need to get a certificate updated before I can redeploy.  Working on it.

boekabart:
Appreciate you're busy and doing this for the community, @darrelmiller , but any update?

darrelmiller:
@boekabart My apologies.  I have the new cert.  Now I need to actually deploy it. I will try and do this later today.

PR #417, opened 2019-09-09
Use context specific storage for param parsing

### LINKED ISSUE #416: V2 Reader is not thread safe
OpenApiV2Deserializer ParameterDeserializer partial uses static state (`_isBodyOrFormData`) that produces incorrect documents when completely separate `OpenApi*Reader` instances are used in different threads.

There may be other cases of misuse of global static state that I haven't found yet.

wpbrown:
Change from a static variable to temp storage scoped to the parse
context to fix thread-safety issue in v2 parameter parsing.

Fixes #416


darrelmiller:
@wpbrown Thank you for this PR.  I'm going to include this fix in the 1.1.4 release that I am currently preparing

PR #418, opened 2019-09-15
Collapsed components that are references

### LINKED ISSUE #415: Schema reference not parsed as expected, results in circular reference
It looks like there is an issue parsing schema references of a certain structure in an OA3 json document.  The problem manifests when schema A is defined (only) as ""$ref":"#/components/schemas/B".  in this case, the OpenApiDocument.Components.Schemas entry for A, has an UnresolvedReference, but the Reference.Id is #/components/schemas/A.

To demonstrate this using PetStore (OA 3, json), I have made two changes:
1. change the response for the /pets path's Get operation to refer to AllPets:
it was:
`"$ref": "#/components/schemas/Pets"`
change to:
`"$ref": "#/components/schemas/AllPets"`
and then, 
2. add the new AllPets schema definition:
`"AllPets": {
        "$ref":"#/components/schemas/Pets"
     },`
(yes, this seems unnecessary, but in my case I don't control the OA3 json generation, I'm just trying to parse it).

Per my reading of the OA3 spec, this syntax/structure _should_ be valid: the schemas element is a map of string (schema id) to **either** a schema or a reference object.
From: [OA3 spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#componentsObject).

When I attempt to ResolveSchema() on the AllPets schema object, it fails to resolve - I would expect it to resolve to the Pets schema object:
`     string oaFilepath = GetTestFilepath ("petstore-3.0.json");

     OpenApiStreamReader reader = new OpenApiStreamReader ();
     using (Stream file = new FileStream (oaFilepath, FileMode.Open))
     {
        OpenApiDocument doc = reader.Read (file, out OpenApiDiagnostic diags);
        OpenApiSchema schema1 = doc.Components.Schemas["AllPets"];
        if (!schema1.UnresolvedReference)
           return;
        OpenApiSchema schema2 = (OpenApiSchema)doc.ResolveReference (schema1.Reference);
        if (schema2.UnresolvedReference && schema1.Reference.Id == schema2.Reference.Id)
        {
           // detected a cycle - this code gets triggered
           Log ("This is a circular reference, WTF? {0}", schema1.Reference.Id);
        }
     }
`
I am fairly new to using OpenApi.NET so I might be misunderstanding what is expected.  I tried to trace through the reader/parsing code and it looks like the SharpYaml representation has the expected reference from AllPets to Pets, but by the time the OpenApiDocument gets returned from the reader.Read() call, the Components.Schemas["AllPets"] entry is incorrect.

darrelmiller:
Attempt to fix #415 
This is only partial fix.  It does ensure that the object model is valid however, due to the way references are resolved we lose the fact that in the original document the reference was to AllPets as it gets collapsed down to Pets.  More investigation is required.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/ParseNodes/MapNode.cs
+++ b/src/Microsoft.OpenApi.Readers/ParseNodes/MapNode.cs
@@ -100,11 +100,15 @@ namespace Microsoft.OpenApi.Readers.ParseNodes
                      {
                          return null;  // Body Parameters shouldn't be converted to Parameters
                      }
-                     entry.value.Reference = new OpenApiReference()
+                     // If the component isn't a reference to another component, then point it to itself.
+                     if (entry.value.Reference == null)
                      {
-                         Type = referenceType,
-                         Id = entry.key
-                     };
+                         entry.value.Reference = new OpenApiReference()
+                         {
+                             Type = referenceType,
+                             Id = entry.key
+                         };
+                     }
                      return entry;
                  }
                  );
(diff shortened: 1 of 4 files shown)


darrelmiller:
This partial fix should be merged because it is better than not having it there.  I still don't know how we can make these references to references actually roundtrip.

Issue #428, opened 2019-11-08
OpenApiSchema.SerializeAsV2WithoutReference does not recursively serialize references


nth-commit:
Version 1.1.4

Given the document:

```
{
  "swagger": "2.0",
  "paths": {
    "/foos": {
      "get": {
        "produces": [
          "application/json"
        ],
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/Foo"
            }
          }
        }
      }
    }
  },
  "definitions": {
    "Foo": {
      "type": "object",
      "properties": {
        "bar": {
          "$ref": "#/definitions/Bar"
        }
      }
    },
    "Bar": {
      "type": "object",
      "properties": {
        "baz": {}
      }
    }
  }
}
```

When serializing the schema referenced by the path "/foos"

**Expected:**

```
{
  "type": "object",
  "properties": {
    "bar": {
      "type": "object",
      "properties": {
        "baz": {}
      }
    }
  }
}
```

**Actual:**

```
{
  "type": "object",
  "properties": {
    "bar": {
      "$ref": "#/definitions/Bar"
    }
  }
}
```

nth-commit:
I was trying to imagine a use-case where someone would call `SerializeAsV2WithoutReference` and not want nested references to be denormalized, and I couldn't. However, in the code, there is a situation where this would happen.

Given the JSON schema:
```
{
    "type": "object",
    "properties": {
        "foo": {
            "$ref": "#/$defs/foo"
        }
    }
}
```

When `SerializeAsV2` is called on the root, the code will see that there's no reference, and then go into `SerializeAsV2WithoutReference`, which will go and write the properties of that schema. As the caller, you would expect the JSON schema to be something equivalent to the above, but given the naive fix, this would break (and denormalize where not intended). What I mean by "the naive fix": https://github.com/nth-commit/OpenAPI.NET/commit/b95d871fe9c73ac34b8f123172aff925a1bd78da

DerekChasse:
I'm actually looking at accomplishing exactly this, so thank you for starting the discussion.

Is there a solution which would account for circular references?

```
{
  "swagger": "2.0",
  "paths": {
    "/foos": {
      "get": {
        "produces": [
          "application/json"
        ],
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/Foo"
            }
          }
        }
      }
    }
  },
  "definitions": {
    "foo": {
      "type": "object",
      "properties": {
[...]

PR #430, opened 2019-11-13
Bugfix #428 - Recursively denormalize references if entry point is 'serialize without references'

nth-commit:
Made some changes to the schema serialization, so that the intention of the caller is preserved. Specifically, whether or not they want reference schemas to be inlined. 

I broke some tests around OpenApiComponents, which is another example of the behaviour change, but I think the tests relied on the bug 😄 

More info in #428

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
+using System;
 using System.Collections.Generic;
 using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Interfaces;
@@ -257,7 +258,13 @@ namespace Microsoft.OpenApi.Models
                 return;
             }
 
-            SerializeAsV3WithoutReference(writer);
+            writer.WriteStartObject();
+
+            WriteSchemaPropertiesAsV3(
+                writer,
+                (w, s) => s.SerializeAsV3(w));
+
+            writer.WriteEndObject();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


nth-commit:
Haven't tested properly but pretty sure this will infinite loop on circular references. See further comments on: #428

DerekChasse:
I'm feeling that, depending on what @darrelmiller comes back with, this could be considered a breaking change.  I'm still confused as to the intent of `IOpenApiReferenceable` and the intent of `SerializeWithoutReference`.

From name alone, it does sound like it should resolve and expand references, but perhaps that's not the intent.  This begs the question on how this code would work with external references?  There are already many open bugs/questions around this. Also, when is enough? Are we chasing and expanding references until all are resolved?  What is the performance impact if we are resolving and expanding references that are 10 children deep? 20? 100? 

For a more gentle approach, I would recommend adding something like an optional `OpenApiWriterSettings` argument to the constructors of `OpenApiJsonWriter` and `OpenApiYamlWriter`  (Really in `OpenApiWriterBase`) This writer settings class would contain indicators on what to do with local references, and other things in the future.  This is a similar pattern which Newtonsoft takes with their `JsonSerializerSettings` class.

darrelmiller:
@DerekChasse @nth-commit I'm going to try and get this fix into the 1.2 release.  I agree that having a setting for that writer that choose whether to write out references as references or inline is an ideal solution.  We will also have to have a further qualification to decide what to do with external references.  I do something very similar in the reader settings.

The SerializeWithoutReference was just a special case to allow us to serialize top level components.  However, that has a bug at the moment that I need to fix first.  Currently the following is not supported properly:

```
openapi: 3.0.0
info:
  title: blag
  version: 1.0.0
paths:
  /: 
    get:
      responses:
        200:
          $ref: '#/components/responses/bob'
components:
  responses:
    bob:
      $ref: '#/components/responses/otherbob'
    otherbob:
      description: ok
```
This scenario is not that common for OpenAPI descriptions in a single file but it is much more common when referencing schemas in external files.  Once I have fixed this we can look into enabling inlining of refs.  The slight challenge is that because I didn't use the Visitor for the writer, we will have to duplicate the loop detection into the writer.  That shouldn't be a huge deal.

nth-commit:
A few years later... Just bumped into this. So glad this works now, huge thanks @darrelmiller 🙌

PR #431, opened 2019-11-18
Explicitly specified string must stay a string

VitaliyKurokhtin:
If string value was specified in YAML explicitly via quotes, or one of the other string syntaxes it must not be converted into any other type.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
+++ b/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
@@ -50,12 +50,24 @@ namespace Microsoft.OpenApi.Readers.ParseNodes
                 return newObject;
             }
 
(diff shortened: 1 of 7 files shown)


VitaliyKurokhtin:
@darrelmiller, please take a look.

darrelmiller:
Hey @VitaliyKurokhtin Thanks for the PR.  Sorry it took so long for me to get around to looking at this.  I am a bit concerned that we will lose the ability to represent DateTime values natively with this change.  There's no choice but to explicitly quote them in when using JSON style and therefore DateTimes are always going to come back as strings.  This isn't necessarily a terrible thing, but it would definitely be a breaking change.
I have no issue with treating quoted numbers and booleans as strings because JSON has a unquoted format for those.

VitaliyKurokhtin:
Good point. I'll make a change to bring back older behavior for DateTimes specifically.

Issue #407, opened 2019-05-14
format:int64 parameter throws data mismatch error


selman92:
While parsing this document:

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  description: A sample API that uses a petstore as an example to demonstrate features in the OpenAPI 3.0 specification
  termsOfService: http://swagger.io/terms/
  contact:
    name: Swagger API Team
    email: apiteam@swagger.io
    url: http://swagger.io
  license:
    name: Apache 2.0
    url: https://www.apache.org/licenses/LICENSE-2.0.html
servers:
  - url: http://petstore.swagger.io/api
paths:
  /users:
    post:
      summary: Adds a new user
      requestBody:
        content:
          application/xml:
            schema:      # Request body contents
              $ref: '#/components/schemas/User'
            example:   # Sample object
                id: 10
                name: Jessica Smith
      responses:
        '200':
          description: OK
components:
  schemas:
    User:    # Schema name
      type: object
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
```

I am getting this error in diagnostics:

`Data and type mismatch found. [#/paths/~1users/post/requestBody/content/application~1xml/example/id`

If I remove `format:int64 `in the `id` definition, the error goes away. I don't understand what's wrong here, how should int64 format look like? 10 is not a valid value in long format??

Version 1.1.1 didn't throw this error, I started getting it after updating to 1.1.3
[sample.txt](https://github.com/microsoft/OpenAPI.NET/files/3180107/sample.txt)

darrelmiller:
This definitely seems like a bug to me.  I actually don't think we should be returning errors at all for examples that don't match the schema.  They should only be warnings.  However, what you are showing is worse because you actually have a valid number.

david-jessee:
So apparently its parsing the integers at strings since it sees them as an "any type"  
This can be fixed in the current version by nesting the example inside of a value attribute (see below.  the attach function wasn't working).
However, if you do this then the swagger ui's don't work that well. When you past the below yml into https://editor.swagger.io/ it sees "value" as part of the output.


openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  description: A sample API that uses a petstore as an example to demonstrate features in the OpenAPI 3.0 specification
  termsOfService: http://swagger.io/terms/
  contact:
    name: Swagger API Team
    email: apiteam@swagger.io
    url: http://swagger.io
[...]

PR #432, opened 2019-11-18
Reevaluating any objects once schema becomes available if parsing in …

VitaliyKurokhtin:
In most general case we can't do any type conversion as we parse YAML tree since schemas may not yet be available if referenced via $ref. That is only possible when whole document is parsed and we can resolve such references to reference schema for type information.

But it is also often the case that we do not parse whole document or maybe schema is available and is not a reference.

This change still tries to convert any value as parsing is performed, but for scenarios when whole document is being parsed it also schedules delayed any value conversion. After document is fully parsed delayed conversion is performed.

This issue has surfaced numerous times for us and is reported at least once here as well #407.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
+++ b/src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
@@ -6,7 +6,7 @@ using System.Globalization;
 using System.Linq;
 using System.Text;
 using Microsoft.OpenApi.Any;
-using Microsoft.OpenApi.Exceptions;
+using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Models;
 
 namespace Microsoft.OpenApi.Readers.ParseNodes
@@ -42,7 +42,7 @@ namespace Microsoft.OpenApi.Readers.ParseNodes
                 return newObject;
             }
 
-            if ( !(openApiAny is OpenApiString))
+            if (!(openApiAny is OpenApiString))
             {
                 return openApiAny;
             }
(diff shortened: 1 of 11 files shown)


VitaliyKurokhtin:
@darrelmiller, please take a look at this one as well.

darrelmiller:
Is it this error "Data and type mismatch found" that you are getting?  I didn't think we should have added that as an error.  At most it should have been a warning.  The spec explicitly says tooling doesn't have to validate examples:

"Tooling implementations MAY choose to validate compatibility automatically,"
http://spec.openapis.org/oas/v3.0.2#fixed-fields-15

I think we should just kill this rule https://github.com/microsoft/OpenAPI.NET/blob/d1695277839502a0f2ce1908957b6c3349a658c5/src/Microsoft.OpenApi/Validations/Rules/OpenApiSchemaRules.cs#L20 until we have a way to generate warnings.

Validating examples can easily be done as a separate step after parsing has been done.

VitaliyKurokhtin:
Right now it affects not only samples, but any ANY field, those are:

- header.default
- header.enum
- parameter.default
- parameter.enum
- mediaType.example
- schema.default
- schema.example

VitaliyKurokhtin:
I think the problem is not only in validation. Value by default will be parsed as string and represented in memory as OpenApiString. That may be unexpected to client tooling going through object model to see OpenApiString where as schema clearly defines number. Reverse is worse.

darrelmiller:
@VitaliyKurokhtin I agree.  I think it was a mistake to try and load examples into a typed DOM based on the schema.  It is possible to do this using the deferred loading that you implemented while we are only considering a single document.  However, this is going to be much harder to do once we support external references.  

I think we should consider loading examples as strings and then provide other mechanisms to parse that string based on a provided schema.  We need to decouple the interpretation of examples from the parsing.

VitaliyKurokhtin:
I agree. We just got another case of the same issue not covered by change in this PR. Will work on that when I have time.

VitaliyKurokhtin:
Another case in favor of not parsing samples at all is the case when example is placed in components. It may be later referenced from multiple places possibly with different schemas.

darrelmiller:
Some of the issues in the PR were addressed in other PRs.  The problem around typing of examples will be addressed in a future major version.

PR #436, opened 2019-12-02
Always emitting description property for response object. It's required

VitaliyKurokhtin:
Right now if empty response is provided it will be serialized as empty object thus producing invalid spec.

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiResponse.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiResponse.cs
@@ -3,7 +3,6 @@
 
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Writers;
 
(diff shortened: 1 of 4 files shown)


darrelmiller:
:-)  That's a creative solution to the required description.

src/Microsoft.OpenApi/Models/OpenApiResponse.cs
```diff
@@ -79,7 +78,14 @@ public void SerializeAsV3WithoutReference(IOpenApiWriter writer)
             writer.WriteStartObject();
 
             // description
-            writer.WriteProperty(OpenApiConstants.Description, Description);
+            if (this.Reference == null)
```
darrelmiller: I'm not sure we want this condition.  The only time this method is called when reference is not null is when it is a top level object in a component map.  In those cases we would still want it to validate.

VitaliyKurokhtin:
Well, spec doesn't say that it MUST be non-null value :)

src/Microsoft.OpenApi/Models/OpenApiResponse.cs
darrelmiller: If I used the following OpenAPI Description, it does not generate a description

```
openapi: 3.0.3
info:
  version: 1.0.0
  title: hey
paths:
  /:
    get:
      responses:
        200: 
          $ref: "#/components/responses/success"
components:
  responses:
    success: {}
          
```

VitaliyKurokhtin: That is true. It was later that I realized that Reference property is used for both, to signal **reference** to the component, and to **provide** reference to use to point to the component.

Issue #441, opened 2019-12-23
OpenApiOAuthFlow serializes unescaped Uris, which creates non-compliant Uris


AlbertoSadoc:
## Simplified repro
```c#
new OpenApiOAuthFlow { AuthorizationUrl = new Uri("https://foo/?bar=a%20b") } 
// https://foo/?bar=a b
```

-------------------------------------------

When serializing the Uri objects with ToString() all characters are unescaped and a generated Uri can become non-compliant with [RCF3986](https://tools.ietf.org/html/rfc3986#appendix-B)

https://github.com/microsoft/OpenAPI.NET/blob/578e5870d299992b9cd8414c1a1e655a56e0403c/src/Microsoft.OpenApi/Models/OpenApiOAuthFlow.cs#L57-L63

darrelmiller:
The AuthorizationUrl can be a [relative reference](http://spec.openapis.org/oas/v3.0.2#oauth-flow-object) and therefore you can't use AbsoluteURI on the value.  As OpenAPI descriptions are intended for users to create by hand, it was decided not to require those users to escape references.  It should be up to the tooling to ensure the URLs are properly escaped before using them.  I think I would prefer if we had used OriginalString rather than ToString() to preserve what the user actually entered.  However, I am not sure exactly what the different between those two properties are.

darrelmiller:
I can repro the reported issue and I think OriginalString should be used here.  Reopening the issue.

PR #442, opened 2019-12-23
[441] Use AbsoluteUri instead of ToString

AlbertoSadoc:
ToString unescape the characters and create a non-compliant Uri
https://github.com/microsoft/OpenAPI.NET/issues/441


darrelmiller:
Cannot use AbsoluteURI due to relative references, but I am re-opening issue #441 to resolve the reported issue by using originalString.

PR #461, opened 2020-02-19
Safer parsing of explicit strings

VitaliyKurokhtin:
The problem with previous implementation is that explicit strings are parsed as date-time even if format is not specified. This new implementation tries to guess type of value from a more narrow set of types for explicit strings.


darrelmiller:
I'm going to pull this into our next 1.X release but I still think our 2.0 release needs to stop trying to parse examples.

PR #474, opened 2020-04-28
[472] Add path signature unique validator

### LINKED ISSUE #472: OpenApiPaths and Operations
Hi all
I have a question regarding OpenApiPaths and their operations:
Let's assume I have the following two endpoints:
```
/api/mp/v1.0/order/{ordernumber}
/api/mp/v1.0/order/{id}
```
Is this valid from an OpenApi spec point of view?
According to online.swagger.io and OpenApiDiagnostic this is no issue.
Whey are they not combined? Or why is this not a validation issue?
With the following example from an API point of view I don't know what has been called:
/api/mp/v1.0/order/10

code to generate OpenApi spec:
```
        /// <summary>
        /// Gets the resource with the specified id.
        /// </summary>
        /// <param name="id">The resource id.</param>
        /// <response code="200">The resource with the specified id.</response>
        /// <response code="404">A resource with the specified id could not be found.</response>
        [HttpGet("{id}")]
        [Authorize(PolicyNames.OrderManagement.CanRead)]
        [ProducesResponseType(typeof(OrderReadDTO), StatusCodes.Status200OK)]
        [ProducesResponseType(typeof(BaseErrorResult), StatusCodes.Status404NotFound)]
        public async Task<IActionResult> GetById(int id)
        {
            var result = await _OrderService.GetSingleAsync(id);
            return new ObjectResult(result);
        }
        /// <summary>
        /// Gets the resource with the specified order number.
        /// </summary>
        /// <param name="orderno">The resource order number.</param>
        /// <response code="200">The resource with the specified order no.</response>
        /// <response code="404">A resource with the specified order no could not be found.</response>
        [HttpGet("{orderno}")]
        [Authorize(PolicyNames.OrderManagement.CanRead)]
        [ProducesResponseType(typeof(OrderReadDTO), StatusCodes.Status200OK)]
        [ProducesResponseType(typeof(BaseErrorResult), StatusCodes.Status404NotFound)]
        public async Task<IActionResult> GetByOrderNo(int orderno)
        {
            var result = await _OrderService.GetSingleAsync(orderno);
            return new ObjectResult(result);
        }
```

fabich:
reenabled OpenApiPaths validators
added validation
added tests for new validator and existing PathsMustBeginWithSlash validator
fixes #472


src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -1036,6 +1037,7 @@ internal void Walk(IOpenApiElement element)
                 case OpenApiOAuthFlow e: Walk(e); break;
                 case OpenApiOperation e: Walk(e); break;
                 case OpenApiParameter e: Walk(e); break;
+                case OpenApiPaths e: Walk(e); break;
```
darrelmiller, line 1040: We already walk OpenAPIPaths here https://github.com/microsoft/OpenAPI.NET/blob/e6535f7380e424668163f812da80d8e8c270b649/src/Microsoft.OpenApi/Services/OpenApiWalker.cs#L48   I think this line may not be necessary.  The Visit method in the OpenAPIValidator is main thing that is needed.

darrelmiller: I was wrong.  This is used for validating fragments.

src/Microsoft.OpenApi/Properties/SRResource.resx
```diff
@@ -207,6 +207,9 @@
   <data name="Validation_PathItemMustBeginWithSlash" xml:space="preserve">
     <value>The path item name '{0}' MUST begin with a slash.</value>
   </data>
+  <data name="Validation_PathSignatureMustBeUnique" xml:space="preserve">
+    <value>The path signature '{0}' MUST begin be unique.</value>
```
darrelmiller, line 211:
```suggestion
    <value>The path signature '{0}' MUST be unique.</value>
```

baywet (approved):

PR #481, opened 2020-05-23
Add literal style support for OpenApiYamlWriter

foriequal0:
OpenAPI allows `description` to be written in Markdown. Markdown descriptions can be multiple lines.
Current `OpenApiYamlWriter` escapes all control characters and joins and serializes them in one line.

I read description from external `README.md` file, and put it to `document.Info.Description`, and publishes the generated OpenAPI document.
Users can `diff` with previous versions of documents to track any public changes.
However, it is quite daunting when you have to look these long concatenated Markdown description to find diffs.

With literal style block scalar, I can preserve my line breaks of `README.md` as is.

### DIFF
--- a/src/Microsoft.OpenApi/Writers/OpenApiWriterBase.cs
+++ b/src/Microsoft.OpenApi/Writers/OpenApiWriterBase.cs
@@ -24,7 +24,7 @@ namespace Microsoft.OpenApi.Writers
         /// <summary>
         /// The indentation string to prepand to each line for each indentation level.
         /// </summary>
-        private const string IndentationString = "  ";
+        protected const string IndentationString = "  ";
 
         /// <summary>
         /// Scope of the Open API element - object, array, property.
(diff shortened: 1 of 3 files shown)


darrelmiller:
@foriequal0 Thank you for this submission.  This is a great addition.  I am going to try it out and get it released ASAP.

darrelmiller:
There is just one issue that concerns me here. Using this input:

```yaml
openapi: 3.0.0
info:
  title: asdads 
  version: 1.0.0
paths:
  /: 
    get:
      description: |
                     Hello
                     World
      responses:
        200:
          description: ok
```

the output is

```yaml
openapi: 3.0.1
info:
  title: asdads
  version: 1.0.0
paths:
  /:
    get:
      description: |+
        Hello
        World

      responses:
        '200':
          description: ok
```

If this output is used as input then the "keep" chomping indicator will cause an extra newline to be introduced.  Would it not be better to use the "clip" chomping indicator as the default output?

foriequal0:
I also found that "clip" indicator introduces a trailling newline even if the original input doesn't have one. 
I'll push more commits to handle chomping indicator correctly.

foriequal0:
Also I realized that I don't have to preserve linebreaks. The spec clearly says:
> Line breaks inside scalar content must be normalized by the YAML processor. Each such line break must be parsed into a single line feed character. The original line break format is a presentation detail and must not be used to convey content information. 
<https://yaml.org/spec/1.2/spec.html#id2774608>

I'll remove clunky line break handlings too.

PR #491, opened 2020-06-24
Make default parameter overloads binary compatible

CumpsD:
This changes all overloads from https://github.com/microsoft/OpenAPI.NET/commit/f00c27cc0e104dd67ae26f85323950a7ae44e5d4 to be binary compatible for not having 1.2.0 break 1.1.4 upgrades


CumpsD:
Retargetting on master

PR #504, opened 2020-07-26
Corrected expected ordering for skipped unit tests.

AndrewRissing:


test/Microsoft.OpenApi.Tests/Services/OpenApiEncodingComparerTests.cs
```diff
@@ -246,14 +246,6 @@ public static IEnumerable<object[]> GetTestCasesForOpenApiEncodingComparerShould
                         SourceValue = "image/png, image/jpeg"
```
AndrewRissing, line 246: The below changes are just reordering of expected differences to align to the current output.  Since the tests expect an explicit ordering, I assume these were taken out due to ordering changes in the code base that caused these tests to break.

I verified the correct ordering for the individual comparer's and assumed the overall `OpenApiComparerTests` suffered the same issue.  If there is some underlying change that would warrant digging into those details, let me know and I can run it down.

Outside of the above, I found a few cases where a `OpenApiComparedElementType` of `Uri` was provided a `TargetValue` and `SourceValue` of type `string` rather than `Uri`, which I corrected as well.

PR #516, opened 2020-08-28
Resolve header references on read

### LINKED ISSUE #515: OpenApiHeader references are not resolved
When accessing the Headers property on OpenApiResponse, if the header is a reference to components then it isn't pre-resolved. It is necessary to call `(OpenApiHeader)document.ResolveReference(response.Headers['header'].Reference)` manually to get details about the header.

brantburnett:
Fixes #515


darrelmiller:
Thanks for this PR.  I may try one slight change once it is merged into the hotfix branch.

darrelmiller:
I changed the visitor to resolve references for any map of headers, not just the one in the response object.  This will fix the same bug in the Encoding object.  Thanks again!

brantburnett:
Thanks!

PR #520, opened 2020-09-16
OpenAPI Workspace Document Fragments

simon-pearson:
- Add `IOpenApiReferenceable ResolveReference(JsonPointer pointer, JsonPointer pointerFromRoot = null);` method to `IOpenApiElement` interface.
- Add `OpenApiElementBase` class which containing an implementation of the `ResolveReference` method used by many of the classes which implement `IOpenApiElement`.
- Update implementation of `OpenApiWorkspace.ResolveReference` method to support referencing document fragments.
- UTs covering the above.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -45,7 +45,7 @@ namespace Microsoft.OpenApi.Readers
         /// <param name="version">Version of the OpenAPI specification that the fragment conforms to.</param>
         /// <param name="diagnostic">Returns diagnostic object containing errors detected during parsing</param>
         /// <returns>Instance of newly created OpenApiDocument</returns>
-        public T ReadFragment<T>(Stream input, OpenApiSpecVersion version, out OpenApiDiagnostic diagnostic) where T : IOpenApiElement
+        public T ReadFragment<T>(Stream input, OpenApiSpecVersion version, out OpenApiDiagnostic diagnostic) where T : IOpenApiReferenceable
         {
             using (var reader = new StreamReader(input))
             {
--- a/src/Microsoft.OpenApi/Extensions/OpenApiReferencableExtensions.cs
+++ b/src/Microsoft.OpenApi/Extensions/OpenApiReferencableExtensions.cs
@@ -30,6 +30,10 @@ namespace Microsoft.OpenApi.Extensions
                 return element;
             try
             {
+                if (element.GetType() == typeof(OpenApiHeader))
+                {
+                    return ResolveReferenceOnHeaderElement((OpenApiHeader)element, jsonPointer);
+                }
                 if (element.GetType() == typeof(OpenApiParameter))
                 {
                     return ResolveReferenceOnParameterElement((OpenApiParameter)element, jsonPointer);
@@ -40,6 +44,24 @@ namespace Microsoft.OpenApi.Extensions
                 throw new OpenApiException(string.Format(SRResource.InvalidReferenceId, jsonPointer));
             }
             throw new NotImplementedException();
+        }
+
+        private static IOpenApiReferenceable ResolveReferenceOnHeaderElement(OpenApiHeader headerElement, string jsonPointer)
+        {
+            var jsonPointerTokens = jsonPointer.Split('/');
+            var propertyName = jsonPointerTokens.ElementAtOrDefault(1);
+            switch (propertyName)
+            {
+                case OpenApiConstants.Schema:
+                    return headerElement.Schema;
+                case OpenApiConstants.Examples:
+                    {
+                        var mapKey = jsonPointerTokens.ElementAtOrDefault(2);
+                        return headerElement.Examples[mapKey];
+                    }
+                default:
+                    throw new OpenApiException(string.Format(SRResource.InvalidReferenceId, jsonPointer));
+            }
         }
 
         private static IOpenApiReferenceable ResolveReferenceOnParameterElement(OpenApiParameter parameterElement, string jsonPointer)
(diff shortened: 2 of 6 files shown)


darrelmiller:
In general this PR looks really good.  I have avoided creating a base class for OpenAPIElements to date.  I tend to try and avoid implementation inheritance where possible.  I was wondering if I we could use an extension method on IOpenApiElement, but there that would mean adding ResolveReferencedProperty to IOpenApiElement.
I probably just need to get over it.  I would think OpenApiElementBase should be abstract.

test/Microsoft.OpenApi.Tests/Workspaces/OpenApiFragmentsTests.cs
```diff
@@ -0,0 +1,254 @@
...
+                new object[] { new OpenApiServerVariable() },
+                new object[] { new OpenApiXml() }
+            };
+
+        [Theory]
+        [MemberData(nameof(ElementsWhichCannotBeReferenced))]
+        public void ElementsWhichCannotBeReferencedShouldThrow(IOpenApiElement element)
+        {
+            // Arrange & Act
+            Action resolveReference = () => element.ResolveReference(new JsonPointer("/"));
```
darrelmiller, line 166: Rather than adding ResolveReference to IOpenApiElement but then checking to make sure that any non-referenceable  element just throws, could we instead move ResolveReference into IOpenApiReferencable?

src/Microsoft.OpenApi/Models/OpenApiElementBase.cs
```diff
@@ -0,0 +1,55 @@
...
+// Licensed under the MIT license. 
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.OpenApi.Exceptions;
+using Microsoft.OpenApi.Interfaces;
+
+namespace Microsoft.OpenApi.Models
+{
+    public class OpenApiElementBase : IOpenApiElement
```
darrelmiller, line 11: Can we add XML document blocks for public methods?

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -280,5 +281,45 @@ public void SerializeAsV2(IOpenApiWriter writer)
         {
             // Components object does not exist in V2.
         }
+
+        public override IOpenApiReferenceable ResolveReference(JsonPointer pointer, JsonPointer pointerFromRoot = null)
```
darrelmiller, line 285: This shouldn't be supported.  The components object is not allowed to be the target of a reference.  If someone wants a document that contains just components then they can create an external OpenApiDocument and just add the info element as well as a components element.

darrelmiller:
So, I played around a bit and I was able to move the ResolveReference into the IOpenApiReferenceable interface.  This allows us to remove it from IOpenApiElement.  This guarantees that external fragments can only be referenceable components. 

The spec isn't very clear on this but here http://spec.openapis.org/oas/v3.0.3#reference-object it says:

> A simple object to allow referencing other components in the specification, internally and externally.

This is implicitly saying you can only point to to something externally if you can also point to it internally. e.g. a component.

The ability to point into a component that lives within a component is debatable.  I'm not sure there is a strong use case for doing that.  The risk of allowing it is we create re-usable things that are not explicitly identified as re-usable things.

I know there are some tools that take a much more liberal approach to referencing.  They allow pointing to any valid structure within a document or externally.  We limited this ability in OpenAPI 3 by saying that references could only point to components.

simon-pearson:
Apologies for the slow reply to this, I've been on holiday for the last two weeks.  I'll get back to you on the above points over the next few days.

simon-pearson:
Darrel,

I've re-worked my initial PR, making the following key changes:
1. I moved the `ResolveReference` method onto the `IOpenApiReferencable` interface as you suggested.  This vastly simplifies the implementation of this method.  I do have some questions around this (below).
1. Again I've followed your suggestion and have implemented the `IOpenApiReferencable.ResolveReference` method as an extension method on `IOpenApiReferencable`, removing the need to introduce a base class.

Questions:
1. It's possible I've completely missed the point with my implementation of the `IOpenApiReferencable.ResolveReference` method but as I see it the only OpenAPI objects which are referencable themselves (i.e. implement `IOpenApiReferencable`) and also have referencable child properties are the [Header object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#headerObject) (with properties `schema` and `examples`), the [Parameter object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#parameterObject) (again `schema` and `examples`), and the [Response object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#responseObject) (properties `headers` and `links`).  This makes my implementation of the `IOpenApiReferencable.ResolveReference` method much simpler than it was previously.  Does this look right to you?
1. Do you agree with the following: if we're saying that the only OpenAPI objects which are valid document fragments are those which implement `IOpenApiReferencable` then we should change the constraint on the `OpenApiStreamReader.ReadFragment` method to restrict the method to only returning elements which implement `IOpenApiReferencable`:

```
public T ReadFragment<T>(Stream input, OpenApiSpecVersion version, out OpenApiDiagnostic diagnostic) where T : IOpenApiReferenceable
{
    ...
}
```

Do these changes better align with what you had in mind for document fragments?

darrelmiller:
@simon-pearson I think it is a reasonable assertion to say that ReadFragment can only return things that are referenceable.  I imagine there might be a few edge use cases where someone might want to construct an OpenAPIDocument from pre-existing fragments of JSON that are not necessarily referenceable, but it feels like a reasonable trade off to lose that ability for the simplicity it brings us.  Also, enabling users to have whole new way of having tooling based re-use seems like bad thing thing to encourage.  If you really want to do reuse outside of the boundaries of OpenAPI semantics, then I would suggest using some kind of text pre-processor.

I'm going to try and finish reviewing this PR today.

darrelmiller:
I just realized that this PR is against my OpenApiWorkspace branch, so I'm going to merge that in and get the OpenAPIWorkspace branch updated so it can merge into 1.3 milestone.

PR #524, opened 2020-10-01
Feature/add validator for path parameters not present in path

### LINKED ISSUE #523: Path parameter not present in the path : No error
_I changed the bug goal because I thought the parameters was not parsed but I wasn't watching the correct object._

I tried to parse this small swagger spec because the "bar" parameter is specified as a path parameter but does not appear in the path : 

```
string swagger = @"
openapi: 3.0.1
info:
  title: Sample API
  version: v1
servers:
  - url: 'api.example.com'
paths:
  '/test/{foo}':
    get:
      parameters:
        - name: foo
          in: path
          required: true
          schema:
            type: string
        - name: bar
          in: path
          required: true
          schema:
            type: string
      responses:
        '200':
          description: Ok
          content:
            application/json:
              schema:
                type: string
                
            ";
            var openApiDocument = new OpenApiStringReader();
            var temp = openApiDocument.Read(swagger, out var diagnostic);
```

But when I check the "diagnostic" variable I see there is no error.

Microsoft.OpenApi : 1.2.3
Microsoft.OpenApi.Readers : 1.2.3

Thanks :-)

DCollart:
Fixes #523

For that new rule I used the "PathString" of the context in order to do the check.

I could also maybe have put that rule at the "Path" level, what do you think ?


darrelmiller (approved):

darrelmiller:
Doing the validation at the parameter level is better.  There is an exception that was introduced in 3.1 that says if the path Item is empty, then the required parameter rule is not applied.  https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#path-templating

PR #536, opened 2020-11-25
Support for scientific notation in parsing of "multipleOf" FixedFieldMap.

EricWilson-BluePrism:
I stumbled upon this issue while trying to parse OpenAPI docs from SAP. Within their docs they use scientific notation to describe percentages (ex. discounts). When trying to parse those docs a System.FormatException would be thrown due to the scientific notation. If I manually convert those values to their decimal equivalents (ex. 1.0e-3 == 0.003) everything works fine.

After researching this a bit I found that including the NumberStyles.Float flag in the call to Decimal.Parse() would address the issue without changing the handling of whole number input (ex. 1). This essentially tacks on the AllowExponent flag.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V3/OpenApiSchemaDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V3/OpenApiSchemaDeserializer.cs
@@ -27,7 +27,7 @@ namespace Microsoft.OpenApi.Readers.V3
             {
                 "multipleOf", (o, n) =>
                 {
-                    o.MultipleOf = decimal.Parse(n.GetScalarValue(), CultureInfo.InvariantCulture);
+                    o.MultipleOf = decimal.Parse(n.GetScalarValue(), NumberStyles.Float, CultureInfo.InvariantCulture);
                 }
             },
             {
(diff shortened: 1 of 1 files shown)


EricWilson-BluePrism:
Seems I've made some sort of newb mistake when trying to submit a change to the project. Any guidance on why the "OpenApi-CI-Allbranches" checked failed? I don't see anything related to it in the wiki, and when I click "Details" in the checks window it takes me to a 401 error page.

The change I submitted is exceedingly simple, so if someone else wants to submit I'm fine with that. I just believe it needs to be addressed.

darrelmiller:
Hey @EricWilson-BluePrism thank you for the PR!  I'm not sure what happened in the branch.  I was too slow to check the build log.  Could you just make an update to the PR to trigger another build?  The change looks good to me.  I suspect there may have been a temporary issue with the vNext branch.

darrelmiller (approved):

EricWilson-BluePrism:
Hi @darrelmiller,

I hit the same issue with the OpenApi-CI-Allbranches test failing on my pull request. I submitted additional updates that I overlooked with my original fix for exponential notation support. Any chance you can look at the build log and see if there's a clue about why my pull requests fail that check?

Cheers,
Eric

PR #541, opened 2020-12-20
OpenAPIWorkspace 

darrelmiller:
This PR contains a new OpenApiWorkspace class that is used to hold a set of OpenAPI documents, fragments and artifacts that reference each other.

This PR contains the object model and provides ResolveReference methods that will resolve references from the OpenApiWorkspace.

This is only a partial solution.  Additional work needs to be done on the OpenApiWorkspaceLoader to provide a way to automatically read referenced documents and load them in the the workspace when referenced.


darrelmiller:
The following code will load a stream into an OpenApiDocument and all related documents will be loaded into a common OpenApiWorkspace.

```csharp
            var reader = new OpenApiStreamReader(new OpenApiReaderSettings()
            {
                ReferenceResolution = ReferenceResolutionSetting.ResolveAllReferences
            });

            var result = await reader.ReadAsync(stream);
```

Issue #479, opened 2020-05-19
Query Parameters Serialization - Explode don't respect the default value.


nicolastakashi:
When the `style: form` is present in the specification and `explode` not, the default value is being treated as `false` instead of `true`.

Using the example below:

```
openapi: 3.0.1
info:
  title: 'Tests v1'
  version: v1
paths:
  '/tests':
    get:
      tags:
        - Test
      summary: 'Test'
      description: |-
        Test Operation
      operationId: GetAsync
      parameters:
        - name: test
          description: "test description"
          in: query
          style: form
          schema:
            type: array
            items:
              type: integer
              format: int32
      responses:
        '200':
          description: 'OK'
          content:
            application/json:
              schema:
                type: string
```

I just looked around the code and I could see that the object is assuming the default value for a boolean property in .NET

One thing that could be done is to force the default value to be `true`, what do you think about that?

darrelmiller:
This is merged into the vnext branch and will release in 1.3

PR #544, opened 2021-01-22
Query Parameters Serialization - Explode doesn't respect the default value

simon-pearson:
- A proposed fix for issue #479.
- Add a unit test covering this behavior.


safepage:
I understand this fix but the serialization code looks wrong:

`writer.WriteProperty(OpenApiConstants.Explode, Explode, Style.HasValue && Style.Value == ParameterStyle.Form);`

Maybe it should be:

`writer.WriteProperty(OpenApiConstants.Explode, _explode);`

Then the serialized model would reflect the actual values read in (as _explode is a nullable bool?).

Also, there is now an inconsistency as Style is also not set to a default value:
“Default values (based on value of in): for query - form; for path - simple; for header - simple; for cookie - form.”

darrelmiller:
@safepage The serialization code will write out any value that was read because that is what the Explode property does by default (unless it is the default value).  Only if Explode has not be set, i.e. _explode is null will the default value be used.  If the Explode property returns a "default" then it will match with the default passed as the third parameter to WriteProperty and no Explode value will be written to the output.  This does mean that if a input document redundantly provides the default value, then it will be missing in the output.

PR #546, opened 2021-02-01
Add OpenAPI Diff to compare two OpenAPI specification files

fabich:
Hi @darrelmiller 

A long long time ago we had a quick talk about a tool to compare two OpenAPI specification files.
I finally got the time to open a PR.

Please let me know what you think!

Best regards

fabich


fabich:
we open-sourced this code in another repo until this PR can be merged.
https://github.com/LimeFlight/openapi-diff.

darrelmiller:
Hey @fabich thanks for the PR.  We currently have three different implementations of OpenAPI diff capabilities and they all take a different approach. I'm thinking that the best option is to have each implementation get distributed as a different package so that people can choose the solution that works for them.  I have failed to find the time to do the work that is needed to incorporate diff into the main library and I think I need to stop trying.  I'm going to remove the partial implementation of the comparers that are in the vNext branch.  Maybe we will pick that up later and expose as it's own package.

Anakael:
I've just started creating one more tool there: :)
https://github.com/Anakael/OpenApiChecker
It's just MVP for now, but it is able to compare not old/new specs, but generated with requried ones

baywet:

PR #547, opened 2021-02-04
Adds OpenAPITreeNode for translating OpenAPI paths into a directory structure

irvinesunday:
Proposes:
- Adding a directory structure model for translating OpenAPI paths.
- Adding tests to validate the model properties and methods.


src/Microsoft.OpenApi/Extensions/StringExtensions.cs
```diff
@@ -35,7 +36,23 @@ public static T GetEnumFromDisplayName<T>(this string displayName)
...
         }
+
+        /// <summary>
+        /// Capitalizes each letter of a word in a string.
+        /// </summary>
+        /// <param name="input">String containing the words to be capitalized, delimited by the '-' character.</param>
+        /// <returns>String value with each word capitalized and concatenated.</returns>
+        public static string ToPascalCase(this string input)
+            => string.IsNullOrEmpty(input) ? input : string.Join(null, input.Split(new char[] { '-' }, StringSplitOptions.RemoveEmptyEntries)
+                                                                            .Select(s => ToFirstCharacterUpperCase(s)));
```
ddyett, line 49:
```suggestion
                                                                            .Select(ToFirstCharacterUpperCase));
```

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
```diff
@@ -35,7 +36,23 @@ public static T GetEnumFromDisplayName<T>(this string displayName)
...
+        /// <summary>
+        /// Capitalizes the first letter of an input string.
+        /// </summary>
+        /// <param name="input">String with first letter to be capitalized. </param>
+        /// <returns>The string value with the first letter capitalized.</returns>
+        public static string ToFirstCharacterUpperCase(this string input)
```
ddyett, line 55: can be private it looks like

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
```diff
@@ -35,7 +36,23 @@ public static T GetEnumFromDisplayName<T>(this string displayName)
...
+            return default;
         }
+
+        /// <summary>
+        /// Capitalizes each letter of a word in a string.
+        /// </summary>
+        /// <param name="input">String containing the words to be capitalized, delimited by the '-' character.</param>
+        /// <returns>String value with each word capitalized and concatenated.</returns>
+        public static string ToPascalCase(this string input)
+            => string.IsNullOrEmpty(input) ? input : string.Join(null, input.Split(new char[] { '-' }, StringSplitOptions.RemoveEmptyEntries)
```
ddyett, line 48: NIT: wrap on multiple lines, so the ternary is easier to read.

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,231 @@
...
+{
+    /// <summary>
+    /// A directory structure representing the paths of an OpenAPI document.
+    /// </summary>
+    public class OpenApiUrlSpaceNode
+    {
+        /// <summary>
+        /// All the subdirectories of a node.
+        /// </summary>
+        public IDictionary<string, OpenApiUrlSpaceNode> Children { get; set; } = new Dictionary<string, OpenApiUrlSpaceNode>();
```
ddyett, line 21: can be private?  i don't think the setter gets used as well.

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Flag indicating whether a PathItem has operations.
+        /// </summary>
+        /// <returns>true or false.</returns>
+        public bool HasOperations()
+        {
+            return PathItem != null && PathItem.Operations != null && PathItem.Operations.Count > 0;
```
ddyett, line 84:
```suggestion
            return PathItem?.Operations != null && PathItem.Operations.Count > 0;
```

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,231 @@
...
+    {
+        /// <summary>
+        /// All the subdirectories of a node.
+        /// </summary>
+        public IDictionary<string, OpenApiUrlSpaceNode> Children { get; set; } = new Dictionary<string, OpenApiUrlSpaceNode>();
+
+        /// <summary>
+        /// The name tag for a group of nodes.
+        /// </summary>
+        public string Label { get; set; }
```
ddyett, line 26: should some of these properties be settable outside of this class?

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,231 @@
...
+        /// <param name="label">Name tag for labelling the nodes in the directory structure.</param>
+        /// <returns>The root node of the created directory structure.</returns>
+        public static OpenApiUrlSpaceNode Create(OpenApiDocument doc, string label = "")
+        {
+            OpenApiUrlSpaceNode root = null;
+
+            var paths = doc?.Paths;
+            if (paths != null)
+            {
+                root = new OpenApiUrlSpaceNode("");
```
ddyett, line 147:
```suggestion
                root = new OpenApiUrlSpaceNode(string.Empty);
```

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,231 @@
...
+        public bool HasOperations()
+        {
+            return PathItem != null && PathItem.Operations != null && PathItem.Operations.Count > 0;
+        }
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="segment">The subdirectory of a relative path.</param>
+        public OpenApiUrlSpaceNode(string segment)
```
ddyett, line 91: should this be private since there's a create method

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
irvinesunday: Could be reused by another client perhaps?

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
irvinesunday: They should have private setters

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
irvinesunday: I agree. 
@darrelmiller what was the use case for making this public? I can only see it's being called outside the class in one of Kiota's tests [here](https://github.com/microsoft/kiota/blob/60e831d24c8cdc8cbdc183a9d9f7a9a1343ac889/tests/kiota.core.tests/KiotaBuilderTests.cs#L12) but nowhere else.
We could instead update the `Create()` [method](https://github.com/irvinesunday/OpenAPI.NET/blob/876611bfa610893fb50b413321cb86b3f2380fe0/src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs#L142) for clients to use when they want to create a 'blank' `OpenApiUrlSpace` node as below: 
```
 public static OpenApiUrlSpaceNode Create(OpenApiDocument doc = null, string label = string.Empty)
        {
            OpenApiUrlSpaceNode root = new OpenApiUrlSpaceNode(string.Empty);

            var paths = doc?.Paths;
            if (paths != null)
            {
                root = new OpenApiUrlSpaceNode(string.Empty);

                foreach (var path in paths)
                {
                    root.Attach(path.Key, path.Value, label);
                }
            }
            return root;
        }
``` 
```
var node = OpenApiUrlSpaceNode.Create();
```

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
irvinesunday: They really shouldn't

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
irvinesunday: Could be reused by another public client perhaps?

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
irvinesunday: Resolved

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
irvinesunday: Resolved

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        /// <summary>
+        /// Flag indicating whether a node segment is a path parameter.
+        /// </summary>
+        public bool IsParameter => Segment.StartsWith("{");
+
+        /// <summary>
+        /// Flag indicating whether a node segment is a path function.
+        /// </summary>
+        public bool IsFunction => Segment.Contains("(");
```
darrelmiller, line 46: This is OData specific. It can't be in this library.

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
darrelmiller: I'm fine with making that internal/private

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,233 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Hashes a string value using a specified hash algorithm.
+        /// </summary>
+        /// <param name="hashAlgorithm">The hash algorithm to use for hashing.</param>
+        /// <param name="input">The string to hash.</param>
+        /// <returns>The hashed value.</returns>
+        private static string GetHash(HashAlgorithm hashAlgorithm, string input)
```
darrelmiller, line 114: This function is related to us generating unique identifiers.  I don't believe this should be in this library.  This is Kiota related.

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,233 @@
...
+using System.Text;
+using Microsoft.OpenApi.Extensions;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Services
+{
+    /// <summary>
+    /// A directory structure representing the paths of an OpenAPI document.
+    /// </summary>
+    public class OpenApiUrlSpaceNode
```
darrelmiller, line 16: Would OpenAPITreeNode be a better name?

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        /// <summary>
+        /// The subdirectory of a relative path.
+        /// </summary>
+        public string Segment { get; private set; }
+
+        /// <summary>
+        /// The Pascal-cased alphabet name of a segment.
+        /// </summary>
+        public string Identifier
```
darrelmiller, line 56: Remove Identifier

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
```diff
@@ -0,0 +1,233 @@
...
+    {
+        /// <summary>
+        /// All the subdirectories of a node.
+        /// </summary>
+        public IDictionary<string, OpenApiUrlSpaceNode> Children { get; } = new Dictionary<string, OpenApiUrlSpaceNode>();
+
+        /// <summary>
+        /// The name tag for a group of nodes.
+        /// </summary>
+        public string Label { get; private set; }
```
darrelmiller, line 26: Label should be attached to every node that gets created when merging a path.  We can add a Properties dictionary later if we need to store more data alongside the pathItem.

src/Microsoft.OpenApi/Services/OpenApiUrlSpaceNode.cs
darrelmiller: Yes, a create method with no parameters would also be useful if we wanted to create a tree from scratch

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
```diff
@@ -1,7 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT license. 
+// Licensed under the MIT license.
```
darrelmiller, line 2:
```suggestion
// Licensed under the MIT license. 
```
Fake update to trigger build

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -0,0 +1,225 @@
...
+        /// </summary>
+        public string Segment { get; private set; }
+
+        /// <summary>
+        /// Flag indicating whether the node's PathItems has operations.
+        /// </summary>
+        /// <returns>true or false.</returns>
+        public bool HasOperations(string label)
+        {
+            if ((bool)!PathItems?.ContainsKey(label))
```
baywet, line 52:
```suggestion
            if (!(PathItems?.ContainsKey(label) ?? false))
```

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -0,0 +1,225 @@
...
+            {
+                return false;
+            }
+
+            return PathItems[label].Operations != null && PathItems[label].Operations.Count > 0;
```
baywet, line 57:
```suggestion
            return PathItems[label].Operations?.Any() ?? false;
```

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -0,0 +1,225 @@
...
+            Segment = segment;
+        }
+
+        /// <summary>
+        /// Creates an empty structured directory of <see cref="OpenApiUrlTreeNode"/> node.
+        /// </summary>
+        /// <returns>The root node of the created <see cref="OpenApiUrlTreeNode"/> directory structure.</returns>
+        public static OpenApiUrlTreeNode Create()
+        {
+            return new OpenApiUrlTreeNode("/");
```
baywet, line 75: consider using a constant for that default value that's used multiple times

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -0,0 +1,225 @@
...
+        /// </summary>
+        /// <param name="doc">The OpenAPI document.</param>
+        /// <param name="label">Name tag for labelling the <see cref="OpenApiUrlTreeNode"/> nodes in the directory structure.</param>
+        /// <returns>The root node of the created <see cref="OpenApiUrlTreeNode"/> directory structure.</returns>
+        public static OpenApiUrlTreeNode Create(OpenApiDocument doc, string label)
+        {
+            Utils.CheckArgumentNull(doc, nameof(doc));
+            Utils.CheckArgumentNullOrEmpty(label, nameof(label));
+
+            OpenApiUrlTreeNode root = new OpenApiUrlTreeNode("/");
```
baywet, line 89: consider calling the create method to avoid duplicating the constant.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -0,0 +1,225 @@
...
+
```
baywet, line 90: in [our copy](https://github.com/microsoft/kiota/blob/b8c8c87e71d8a9db38b2a5bfc20d894350b748d5/src/Kiota.Builder/OpenApiUriSpaceNode.cs#L65) we init with empty string. I'm not suggesting any change here, I'm just curious about the reason behind the difference?

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -0,0 +1,225 @@
...
+        {
+            Utils.CheckArgumentNullOrEmpty(label, nameof(label));
+
+            if (path.StartsWith("/"))
+            {
+                // Remove leading slash
+                path = path.Substring(1);
+            }
+
+            var segments = path.Split('/');
```
baywet, line 146: remove empty entries?

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -0,0 +1,225 @@
...
+                return this;
+            }
+
+            // If the child segment has already been defined, then insert into it
+            if (Children.ContainsKey(segment))
+            {
+                return Children[segment].Attach(segments: segments.Skip(1),
+                                                pathItem: pathItem,
+                                                label: label,
+                                                currentPath: currentPath + "\\" + segment);
```
baywet, line 186: consider using a constant for the separator value instead of duplicating the definition

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        /// <summary>
+        /// The relative directory path of the current node from the root node.
+        /// </summary>
+        public string Path { get; private set; } = "";
+
+        /// <summary>
+        /// Dictionary of labels and Path Item objects that describe the operations available on a node.
+        /// </summary>
+        public IDictionary<string, OpenApiPathItem> PathItems { get; } = new Dictionary<string, OpenApiPathItem>();
```
baywet, line 29: in [our version](https://github.com/microsoft/kiota/blob/b8c8c87e71d8a9db38b2a5bfc20d894350b748d5/src/Kiota.Builder/OpenApiUriSpaceNode.cs#L16) path item is singular (not a dictionary). I'm not asking to change anything, I'm just curious about the motivation for the change.

baywet (changes requested):

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
irvinesunday: I figured that the root path segment should ideally be a `/` instead of just an empty string. In the Graph OpenAPI docs, at least, we represent it as such.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
irvinesunday: Well, we can. But in either case, it would still work okay, since we're checking for null or empty strings either way here: https://github.com/irvinesunday/OpenAPI.NET/blob/7d36d697524aeb43369477d3b297996381256045/src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs#L168

image

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
baywet: In [the powershell open api descriptions](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/openApiDocs/v1.0/Mail.yml#L0-L1) the servers already contain a trailing slash. I'm not sure whether this is just us or a standard practice in the community. What I'm wondering is whether starting with a slash here will lead to a double slash and/or to consumers of the open api lib to have to control for that?
The paths also always start with a slash.

If we look at [pet store](https://petstore.swagger.io/v2/swagger.json) which is v2, they use host + base path with no trailing slash, but the paths also all start with a slash.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
irvinesunday: Having it as a dictionary of Labels and PathItem objects will enable us to easily represent the different OpenAPI operations exposed by a single path in e.g. both V1.0 and Beta versions of the Graph metadata. In this case, the versions (labels) will be dictionary keys.

baywet: Thanks for providing more details. I'm not sure I follow entirely here. As beta and v1.0 are base paths for the API shouldn't we end up in a situation where we either have:
- a gigantic open api description with all the paths, everything starting with beta or v1
- two open api descriptions, one for beta, one for v1, with a basepath

In which case, why do we need that labeling capability?

irvinesunday: Both scenarios are feasible with this new structure. 

**Scenario 1**: If for example you start with translating `v1.0` OpenAPI doc. into an `OpenApiUrlTreeNode` object, you can later attach the `beta` document onto it. If a path in the `beta` version has already been defined in the `v1.0` version, no new path will be created, but rather its operations (`OpenApiPathItem`) will be appended to the existing `v1.0` path in the node, and this will be labelled as `beta` (for example) within the `PathItems` dictionary. Agora makes use of this scenario.

**Scenario 2**: If you want to have separate `OpenApiUrlTreeNode` objects, one representing `v1.0 `and the other `beta`, that's perfectly doable. In each case, all the segments for a given `OpenApiUrlTreeNode` object  will have their `PathItems` dictionary key as the specified version passed in the `Create(OpenApiDocument doc, string label)` method.

baywet: perfect! thanks for going through all the details with me!

baywet (approved):

darrelmiller (approved):

PR #551, opened 2021-02-16
Serializer Refactoring

ddrdoubled:
Serializer Refactoring to support DI, reduce Code Duplication and simplify Future Updates of OpenAPI Version.
Unfortunately I do not know whether it is possible to use DI in Framework 4.6. thats why the Public API Test is failing.
Let me know if you agree with general concept and i would add Xml Comments and mark IOpenApiSerializable as Obsolete.
I could also write for Framework 4.6. an Explicit Serializer without DI.


darrelmiller:
I don't really understand why we would want to do this.  There is not likely going to be more than one version update per year to the OpenAPI specification.  Only two major versions are required to be supported by OpenAPI tooling.  Adding the complexity of DI doesn't seem to add much value for such a significant refactor.  Maybe there is more going on that I am missing.

Also, keeping the serialization code close to the model classes was an intentional design choice.  Also, making it extensible was a non-goal as the shape of the output is explicitly designed by a specification and creating the ability for the library to be changed to generate non-compliant output is a concern.

baywet:
closing due to author inactivity

PR #559, opened 2021-03-14
Add support for $ref in PathItems

### LINKED ISSUE #288: $ref is not a valid property path item
Hi,
I tried 1.1.0-preview.3 reader version with this sample:

```yaml
components:
  securitySchemes:
    Token:
      description: This API uses OAuth 2 
      flows:
        implicit:
          authorizationUrl: https://.......
          scopes: []
          tokenUrl: https://.......
      type: oauth2
info:
  title: API
  version: 0.9.0
openapi: 3.0.0
paths:
  /float:
    $ref: paths/float.yaml#/collection
  /float/{id}:
    $ref: paths/float.yaml#/individual
  /float/{id}/tags:
    $ref: paths/tags.yaml#/network_collection
security:
- Token: []
servers:
- url: /v1
```

And I get errors like **$ref is not a valid property at #/paths//float**
I checked the specification and it should be permitted.

Thanks

darrelmiller:
Fixes #288


darrelmiller:
Dammit, I need .net5 on the build machine.

fieldioltd:
Hi.
I still get errors, like _$ref is not a valid property at #/info_
I am using 1.3.1-preview

PR #562, opened 2021-03-29
🚀 Tests upgraded to dotnet core 3.1

### LINKED ISSUE #548: Run tests on .NET core
The tests seem to require .NET Framework 4.6 which is unavailable on MacOS/Linux.

PureKrome:
Fixes #548 

Stumbled over this repo and noticed a request to update the tests to dotnet core. So here's a draft PR for this requirement.

- Test projects are not `netcoreapp3.1`
- `Microsoft.OpenApi.SmokeTests` project doesn't have any actual test attributes? Not sure how this could be ran?
- Cleaned up csproj files for the src. For example, renamed `<TargetFrameworks>` => `<TargetFramework>` where one framework is only listed.
- updated test related nugets.

image

Further suggestions if this PR is accepted/merged:
- build.cmd can be nuked. This is now .net core for all projects
- CI can be added (e.g. github actions)
- `Microsoft.OpenApi.Workbench` is a WPF app. It's using a nuget package that might not be .net core compatible. I've never done any WPF so i'm not investing any time into finding a suitable, free alternative.

image

If this is ok, i'll migrate this off a DRAFT PR -> PR.


darrelmiller:
CI is done in an Azure DevOps repo because it needs to be signed with a Microsoft cert due to the Microsoft namespace on Nuget.
Smoke tests take along time to run as they make a lot of network calls, so I generally run them manually after significant changes.
Until recently we were building for .net46 and .net standard 2.0, hence the <TargetFrameworks>. Thanks for catching that.
I'm wondering if we should just go straight to .net5 for the tests.  I also might be able to get the WPF app working in .NET 5.

PureKrome:
> I'm wondering if we should just go straight to .net5 for the tests. I also might be able to get the WPF app working in .NET 5.

Personally, I'm usually cautious about going to versions of .NET that are _not_ LTS. Just makes the a lot more work.

That said, the WPF app is finally xplat with .NET5 so maybe in this scenario, it's worth the extra effort to get a nice, fully xplat project migrated across. 

Moving from NET5 -> NET6 will then be a minor effort, when NET6 drops at the end of '21.

So how does that sound? I'll update this PR to make the tests go to .NET5   ?

darrelmiller:
Thank you for the contribution and push :-).  Tests are now running in 462 and 50 in vnext.

PR #576, opened 2021-04-12
Fix usings for PR 547

darrelmiller:
@irvinesunday I fixed missing usings that were preventing the PR from compiling.

I was also considering adding a dictionary to store additional properties. But after some pondering, I wonder if we should have a dictionary of labels and pathItems.


irvinesunday:
> @irvinesunday I fixed missing usings that were preventing the PR from compiling.
> 
> I was also considering adding a dictionary to store additional properties. But after some pondering, I wonder if we should have a dictionary of labels and pathItems.

What's the use case? I presume the expectation is to have only one `pathItem` and `label` per node (path)?

PR #578, opened 2021-04-12
Add continuous code security and quality analysis

### LINKED ISSUE #577: Add continuous code security and quality analysis
- [CodeQL](https://github.com/github/codeql) is an open source code scanning tool for discovering vulnerabilities across a codebase

Built with ❤ by [Pipeline Foundation](https://pipeline.foundation).

aleks-ivanov:
- add `codeql-analysis.yml` which automatically enables CodeQL code security and quality scanner. It executes on every push commit, PR, manually and every day at 8:00AM UTC. A scan check can be viewed on commits and PRs. All alerts and fix suggestion can be viewed at **Security** tab -> **Code scanning alerts** -> **CodeQL**

Resolves #577


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+          $targetsArray = @(
+            'restore',
+            'build'
+          )
+
+          $projectsArray | ForEach-Object {
+            $project = $PSItem
+            
+            $targetsArray | ForEach-Object {
+              dotnet msbuild $project `
```
baywet, line 42:
```suggestion
              dotnet build $project `
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 8 * * *'
+
+jobs:
+  analyze:
+    name: CodeQL Analysis
+    runs-on: windows-latest
```
baywet, line 12:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: Checkout repository
+        id: checkout_repo
+        uses: actions/checkout@v2
+
+      - name: Initialize CodeQL
```
baywet, line 18: missing setup dotnet step https://github.com/microsoft/kiota/blob/3054610dd7e679a099eb2b7fdd9bc6227944cfaa/.github/workflows/abstractions-dotnet.yml#L17

baywet (changes requested):
thanks for adding tooling to improve the library, here are a few suggestions

.github/workflows/codeql-analysis.yml
aleks-ivanov: We took direction for the `dotnet msbuild` from the `build.cmd` script in the project. The `dotnet build` seems to do the same as `dotnet msbuild`, but just to be sure we are not committing something invalid, do you know if there's any difference for the project between the two ? (other than `dotnet build` implicitly running `dotnet restore`)

baywet: as far as I understand it's very similar except for the way you pass arguments to msbuild. dotnet msbuild being closer to the msbuild syntax from back in the days. https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-msbuild#description

.github/workflows/codeql-analysis.yml
aleks-ivanov: Shouldn't be required, because the GH Ubuntu runner comes with [.NET SDKs preinstalled](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#net-core-sdk).
We just tested the pipeline on `ubuntu-latest` with `dotnet build` and it works flawlessly 🙂

baywet: I usually always add the dotnet install step because relying on pre-installed tools is a moving target (e.g. what are going to be the defaults when dotnet 6 comes out? will that derail this worklow?). Worst case it's a no-op (besides setting environment variables), best case it prevents workflows from stopping to work when new versions come out or default images change. But I get it's a preference thing, so I won't block on that.

baywet (approved):

.github/workflows/codeql-analysis.yml
aleks-ivanov: That makes perfect sense, actually. 
We'd add it promptly.

baywet (approved):

Issue #587, opened 2021-05-13
OpenAPI "data and type mismatch found"


Thorium:
I'm trying to read this schema:
https://developer.experian.com/system/files/2020-12/credit-profile-product-apis-v-openapi3.2.0.45.yaml

I'm getting a lot of "Data and type mismatch found" errors.
And indeed, it seems that the types vary:
permissiblePurpose/type seems to be sometimes 08 and sometimes '08' and sometimes just ''.

Is the schema totally un-usable wtih OpenAPI.NET, or is it still possible to use it by forcing the types manually (with some parameter maybe?), so that the type is "either string or int" or "force to string"?

I don't care the solution details, I do care a workaround to get this ~10k lines of type-definitions into use easily.

darrelmiller:
This was a rule that should never have been added as a rule that produces an error.

PR #588, opened 2021-05-15
Removed DataTypeMismatch as default rule

darrelmiller:
Addresses #407 and #587  by disabling the validation rule by default


Thorium:
Excellent, thanks! 
When can we expect this to NuGet?

PR #595, opened 2021-05-20
Add several syntactical, functional and logic improvements (#11)

### LINKED ISSUE #594: GitHub Actions CI/CD pipeline optimizations
Hi @baywet and @darrelmiller 🙂 

Yesterday when we got the notification about the merge, we decided to give the pipeline a glance and see if anything we've learned during the past month could be applied to optimize it.

In the connected PR we've added detailed descriptions of the changes, including one very curious finding about the tests.

aleks-ivanov:
You will see that the pipeline fails at the "Run unit tests" step and here is why.

We discovered that on `ubuntu-latest` (Ubuntu 20.04), one of the tests from the `Microsoft.OpenApi.Tests` project fails. The test is called `Microsoft.OpenApi.Tests.OpenApiWorkspaceTests.OpenApiWorkspacesShouldNormalizeDocumentLocations`.

The same test completes successful on `Windows 10` and `windows-latest` (Windows Server 2019).

Now to make it even more interesting, in the previous iteration of the pipeline [you can see in the logs](https://github.com/microsoft/OpenAPI.NET/runs/2620549887?check_suite_focus=true) that the test fails, but for some odd reason, there is no exit code and therefor the pipeline doesn't fail.

In the one this PR is for, the same test again fails BUT now [the pipeline also fails as it should](https://github.com/aleks-ivanov/OpenAPI.NET/runs/2630297552?check_suite_focus=true).

In case this fail is a false positive, we've tested the pipeline on `windows-latest` and everything runs smoothly, since the test in question does not fail, so we can switch the runner of the CI job from `ubuntu-latest` to `windows-latest`.

#

Other than that, check out in a bit the comments on each change for info on the optimizations.

Resolves #594


.github/workflows/ci-cd.yml
```diff
@@ -1,14 +1,13 @@
 name: CI/CD Pipeline
 
-on: [push, pull_request, workflow_dispatch]
+on: [ push, pull_request, workflow_dispatch ]
```
aleks-ivanov, line 3: Small change to improve legibility.

.github/workflows/ci-cd.yml
```diff
@@ -1,14 +1,13 @@
...
 
 jobs:
   ci:
     name: Continuous Integration
     runs-on: ubuntu-latest
     outputs:
       latest_version: ${{ steps.tag_generator.outputs.new_version }}
-      is_default_branch: ${{ steps.conditionals_handler.outputs.is_default_branch }}
```
aleks-ivanov, line 11: Removed unnecessary value.

.github/workflows/ci-cd.yml
```diff
@@ -23,7 +22,7 @@ jobs:
         shell: pwsh
         run: |
           # Get default branch
-          $repo = 'microsoft/OpenAPI.NET'
+          $repo = "${{ github.repository }}"
```
aleks-ivanov, line 25: Replace static value with more elegant dynamic value - `${{ github.repository }}`.

.github/workflows/ci-cd.yml
```diff
@@ -33,11 +32,22 @@ jobs:
...
           }
+          if ( $githubEventName -eq 'push' ) {
+            $isPush = 'true'
+          }
+          if ( $githubRef -like "*$defaultBranch*" -and $githubEventName -eq 'push' ) {
+            $isPushToDefaultBranch = 'true'
+          }
           Write-Output "::set-output name=is_default_branch::$(echo $isDefaultBranch)"
+          Write-Output "::set-output name=is_push::$(echo $isPush)"
+          Write-Output "::set-output name=is_push_to_default_branch::$(echo $isPushToDefaultBranch)"
```
aleks-ivanov, line 50: Added two more conditional statements:
1. `is_push` = when the workflow is triggered by a push commit
2. `is_push_to_default_branch` = when the workflow is triggered by the a push commit in the default branch

.github/workflows/ci-cd.yml
```diff
@@ -46,7 +56,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           fetch-depth: 0
 
-      - if: steps.conditionals_handler.outputs.is_default_branch == 'true'
+      - if: steps.conditionals_handler.outputs.is_push_to_default_branch == 'true'
```
aleks-ivanov, line 59: Used the newly added `is_push_to_default_branch` condition, instead of the `is_default_branch`, to further restrict the new tag generating step to only execute when a push commit to the default branch is made.

.github/workflows/ci-cd.yml
```diff
@@ -59,41 +69,40 @@ jobs:
...
         shell: pwsh
         run: |
-          $projectsArray = @(
-            '.\src\Microsoft.OpenApi\Microsoft.OpenApi.csproj',
-            '.\src\Microsoft.OpenApi.Readers\Microsoft.OpenApi.Readers.csproj',
-            '.\src\Microsoft.OpenApi.Tool\Microsoft.OpenApi.Tool.csproj'
-          )
           $gitNewVersion = if ("${{ steps.tag_generator.outputs.new_version }}") {"${{ steps.tag_generator.outputs.new_version }}"} else {$null}
-          $projectCurrentVersion = ([xml](Get-Content .\src\Microsoft.OpenApi\Microsoft.OpenApi.csproj)).Project.PropertyGroup.Version
+          $projectCurrentVersion = ([xml](Get-Content -Path .\src\Microsoft.OpenApi\Microsoft.OpenApi.csproj)).Project.PropertyGroup.Version
```
aleks-ivanov, line 73: Add `-Path` argument as best practice. (writing scripts in PowerShell is all cmdlets and arguments should be written and with their full name for optimal legibility)

.github/workflows/ci-cd.yml
```diff
@@ -59,41 +69,40 @@ jobs:
...
           $projectNewVersion = $gitNewVersion ?? $projectCurrentVersion
 
-          $projectsArray | ForEach-Object {
-            dotnet build $PSItem `
-            -c Release # `
-            # -o $env:ARTIFACTS_FOLDER `
-            # /p:Version=$projectNewVersion 
+          Get-ChildItem -Path src/ -Filter *.csproj -Exclude *Workbench* -File -Recurse | ForEach-Object {
```
aleks-ivanov, line 76: Replace static array of project locations with a more elegant solution.

.github/workflows/ci-cd.yml
```diff
@@ -59,41 +69,40 @@ jobs:
...
+            --configuration Release # `
+            # --output $env:ARTIFACTS_FOLDER `
+            # -property:Version=$projectNewVersion 
           }
 
           # Move NuGet packages to separate folder for pipeline convenience
-          # New-Item Artifacts/NuGet -ItemType Directory
-          # Get-ChildItem Artifacts/*.nupkg | Move-Item -Destination "Artifacts/NuGet"
+          # New-Item -Name Artifacts/NuGet -ItemType Directory
+          # Get-ChildItem -Path Artifacts/ -Filter *.nupkg | Move-Item -Destination Artifacts/NuGet
```
aleks-ivanov, line 85: Add several arguments explicitly as per best practice.

.github/workflows/ci-cd.yml
```diff
@@ -59,41 +69,40 @@ jobs:
...
-
-          $testProjectsArray | ForEach-Object {
-            dotnet test $PSItem `
-            -c Release
+          Get-ChildItem -Path test/ -Filter *.csproj -File -Recurse | ForEach-Object {
+            $fileBaseName = $PSItem.Basename
+            dotnet test $PSItem.FullName `
+            --configuration Release `
+            --logger "trx;LogFileName=$fileBaseName.trx;verbosity=normal" `
+            --results-directory TestResults/
```
aleks-ivanov, line 96: Replaced static array of test project locations with a more elegant solution.
Add test logger, which generates test results with normal verbosity in the root/TestResults/ folder.

.github/workflows/ci-cd.yml
```diff
@@ -59,41 +69,40 @@ jobs:
...
           }
 
+      - name: Upload test results as artifacts
+        id: ul_testresults_artifact
+        uses: actions/upload-artifact@v1
+        with:
+          name: TestResults
+          path: TestResults/
+
```
aleks-ivanov, line 105: Upload test results as pipeline artifact.

.github/workflows/ci-cd.yml
```diff
@@ -103,7 +112,7 @@ jobs:
       #     path: Artifacts/NuGet/
 
   cd:
-    if: needs.ci.outputs.is_default_branch == 'true' && needs.ci.outputs.latest_version != ''
+    if: needs.ci.outputs.latest_version != ''
```
aleks-ivanov, line 115: Removed `is_default_branch` condition, because it is logically unnecessary, since the step that dictates if the other condition `needs.ci.outputs.latest_version` is true or false, only executes when the workflow is triggered by a push commit in the default branch.

.github/workflows/ci-cd.yml
```diff
@@ -120,8 +129,8 @@ jobs:
       #   continue-on-error: true
       #   shell: pwsh
       #   run: |
-      #     Get-ChildItem NuGet/*.nupkg | ForEach-Object {
-      #       nuget push $PSItem `
+      #     Get-ChildItem -Path NuGet/ -Filter *.nupkg | ForEach-Object {
```
aleks-ivanov, line 132: Add the two arguments as per best practice.

.github/workflows/ci-cd.yml
```diff
@@ -120,8 +129,8 @@ jobs:
...
+      #       nuget push $PSItem.FullName `
```
aleks-ivanov, line 133: Add `.FullName` to ensure the command always takes the file path as value.

.github/workflows/ci-cd.yml
baywet: out of curiosity, any reason why you don't simply target the sln?

baywet (approved):

aleks-ivanov:
> out of curiosity, any reason why you don't simply target the sln?

We just took the `build.cmd` file as an example, but now that you mention it unless there is a specific reason not to, we could change the build step to building only the solution instead 🙂

.github/workflows/ci-cd.yml
aleks-ivanov: We just took the `build.cmd` file as an example and just never took a second thought about it, but now that you mention it unless there is a specific reason not to, we could change the build step to building only the solution instead 🙂

aleks-ivanov:
Also, I forgot to mention that based on this PRs resolution, we will optimize the CodeQL pipeline as well.

.github/workflows/ci-cd.yml
darrelmiller: I'm somewhat confused by this process.  First, the default branch is vnext, not master.  We don't create releases when we are updating vnext.  Also, we don't push directly to vnext, we generally PR to it.  What am I missing?

.github/workflows/ci-cd.yml
darrelmiller: We don't publish the Workbench tool, so I didn't bother building it for publishing nugets.  I don't have a problem building it as part of the pipeline.  However currently `dotnet build` can't build the Workbench project because of some error about the GenerateResource task.  Not sure what that is about.

.github/workflows/ci-cd.yml
aleks-ivanov: >  We don't create releases when we are updating vnext

If you are not making releases from your default branch (vnext), we will change it to the correct one. Is it `master` or another branch ?

Also not every push commit creates a new GitHub tag and release. In the [documentation](https://github.com/microsoft/OpenAPI.NET/blob/vnext/docs/CI-CD_DOCUMENTATION.md#3-automated-versioning) we submitted with the last PR, you can check out the exact prerequisite to trigger an output (meaning create a new tag) from the "Bump GH tag" step, which in turn triggers a GitHub release.

The path is this:

Push commit in release branch with SemVer label 
-> "Bump GH tag" step creates a new GitHub tag and pushes it to the repo 
-> CD job evaluates if there is a new GitHub tag and if there is, executes
-> the "Create and publish release" step in the CD job uses the newly created and pushed GitHub tag to create a GitHub release

> we generally PR to it

The merge of a PR triggers a push commit, so you can either create a release though your release branch, by committing into it or merging a PR.

If you have any other questions about the automated release process, please let us know! 🙂

baywet (approved):

aleks-ivanov:
@baywet it is an incredible coincidence, but today we discovered that the action for creating release in this pipeline is unmaintained and is quite unstable, so we decided to replace it with a short script using GH CLI.

I am about to commit the change and it will show in this PR, so it would more convenient for merging.

baywet (approved):

.github/workflows/ci-cd.yml
darrelmiller: We create releases off the master branch.

aleks-ivanov: @darrelmiller with the last commit:

- switch the release branch from default_branch to master
- remove the data gatherer and conditionals handler steps, since they were used specifically for the default_branch as release branch logic

In the current state of the pipeline if there are no errors present, a push commit or PR merge to the master branch with a SemVer label in the title, would deploy (publish packages and create a GitHub release).

Push commits or PR merges to the default_branch would trigger the pipeline, but not deploy.

.github/workflows/ci-cd.yml
```diff
@@ -18,82 +17,60 @@ jobs:
...
-
       - name: Checkout repository
         id: checkout_repo
         uses: actions/checkout@v2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           fetch-depth: 0
 
-      - if: steps.conditionals_handler.outputs.is_default_branch == 'true'
+      - if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
baywet, line 27:
```suggestion
      - if: github.event_name == 'push' && github.ref == 'refs/heads/vnext'
```
CC @darrelmiller

.github/workflows/ci-cd.yml
```diff
@@ -18,82 +17,60 @@ jobs:
...
         name: Bump GH tag
         id: tag_generator
         uses: mathieudutour/github-tag-action@v5.4
         with: 
           github_token: ${{ secrets.GITHUB_TOKEN }}
           default_bump: false
-          release_branches: ${{ steps.data_gatherer.outputs.default_branch }}
+          release_branches: master
```
baywet, line 34:
```suggestion
          release_branches: vnext
```

baywet (changes requested):

.github/workflows/ci-cd.yml
aleks-ivanov: @baywet I replaced this logic to point from `vnext` to `master` based on this comment https://github.com/microsoft/OpenAPI.NET/pull/595#discussion_r658865021 by @darrelmiller

baywet: ok ignore my confused self then :)

baywet (dismissed):

raghavthind2005 (approved):

baywet:

Issue #530, opened 2020-10-28
OpenApiStreamReader disposes input stream after reading and it's not configurable


jaceks2106:
Hi all,

I have to use `OpenApiStreamReader` as a one of the several steps, of handling the `Stream`. The problem is, that when I read the stream using `OpenApiStreamReader.Read(stream)` at first, then the Stream is disposed and cannot be used anymore. I cannot configure anyhow to Read the stream without disposing it, as I can do that in the base `StreamReader`.

[Link - bool leaveOpen](https://docs.microsoft.com/pl-pl/dotnet/api/system.io.streamreader.-ctor?view=netcore-3.1#System_IO_StreamReader__ctor_System_IO_Stream_System_Text_Encoding_System_Boolean_System_Int32_System_Boolean_)

Here is my example reproduction:
```
// openApiSpec is a Stream argument of the method
var openApiStreamReader = new OpenApiStreamReader();
var openApiDocument = openApiStreamReader.Read(openApiSpec, out var openApiDiagnostic);
```

darrelmiller:
This is a good call out.  We use StreamReader under the covers with a using statement, hence why dispose is called.   I think we can add a property to OpenReaderSettings to allow control of this.

PR #605, opened 2021-06-07
Adds setting to configure leaving stream open

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET/issues/530
Closes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/583

This PR:
- Adds a new `boolean` setting in `OpenApiReaderSettings.cs`: `LeaveStreamOpen` to flag whether or not to leave the stream object open after reading from an `OpenApiStreamReader` object.
Creating a `StreamReader` object within a `using` statement closes the `stream` when the `Dispose()` method of the `StreamReader` is called.
- Creates `OpenApiStreamReader` tests to validate setting the `LeaveStreamOpen` setting to either `true` or `false`.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
@@ -72,6 +72,6 @@ namespace Microsoft.OpenApi.Readers
         /// Whether to leave the <see cref="Stream"/> object open after reading
         /// from an <see cref="OpenApiStreamReader"/> object.
         /// </summary>
-        public bool LeaveStreamOpen { get; set; } = false;
+        public bool LeaveStreamOpen { get; set; }
     }
 }
--- a/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
@@ -1,5 +1,5 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT license. 
+// Licensed under the MIT license.
 
 using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Interfaces;
@@ -61,11 +61,17 @@ namespace Microsoft.OpenApi.Readers
         public Uri BaseUrl { get; set; }
 
         /// <summary>
-        /// Function used to provide an alternative loader for accessing external references.  
+        /// Function used to provide an alternative loader for accessing external references.
         /// </summary>
         /// <remarks>
         /// Default loader will attempt to dereference http(s) urls and file urls.
         /// </remarks>
         public IStreamLoader CustomExternalLoader { get; set; }
+
+        /// <summary>
+        /// Whether to leave the <see cref="Stream"/> object open after reading
+        /// from an <see cref="OpenApiStreamReader"/> object.
+        /// </summary>
+        public bool LeaveStreamOpen { get; set; } = false;
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -29,13 +29,20 @@ public OpenApiStreamReader(OpenApiReaderSettings settings = null)
...
         /// </summary>
         /// <param name="input">Stream containing OpenAPI description to parse.</param>
-        /// <param name="diagnostic">Returns diagnostic object containing errors detected during parsing</param>
-        /// <returns>Instance of newly created OpenApiDocument</returns>
+        /// <param name="diagnostic">Returns diagnostic object containing errors detected during parsing.</param>
+        /// <returns>Instance of newly created OpenApiDocument.</returns>
         public OpenApiDocument Read(Stream input, out OpenApiDiagnostic diagnostic)
         {
-            using (var reader = new StreamReader(input))
+            if (_settings.LeaveStreamOpen)
```
baywet, line 36:
```CSharp
var reader = new StreamReader(input);
var result = new OpenApiTextReaderReader(_settings).Read(reader, out diagnostic);
if (!_settings.LeaveStreamOpen)
   reader.Dispose();
return result;
```

src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
```diff
@@ -61,11 +61,17 @@ public class OpenApiReaderSettings
...
         /// <remarks>
         /// Default loader will attempt to dereference http(s) urls and file urls.
         /// </remarks>
         public IStreamLoader CustomExternalLoader { get; set; }
+
+        /// <summary>
+        /// Whether to leave the <see cref="Stream"/> object open after reading
+        /// from an <see cref="OpenApiStreamReader"/> object.
+        /// </summary>
+        public bool LeaveStreamOpen { get; set; } = false;
```
baywet, line 75:
```suggestion
        public bool KeepStreamOpen { get; set; };
```
boolean is always false by default.
nit on naming

baywet (changes requested):

src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
irvinesunday: 👍🏼 on the boolean default.

On the naming: I wanted to remain as close to the framework naming as possible: See: https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=netcore-3.1#System_IO_StreamReader__ctor_System_IO_Stream_System_Text_Encoding_System_Boolean_System_Int32_System_Boolean_
image

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
irvinesunday: Yes. Much cleaner!

irvinesunday: I like brackets after `if` statements though, even for one-liners 😄

src/Microsoft.OpenApi.Readers/OpenApiReaderSettings.cs
baywet: perfect, ignore my naming suggestion then. Thanks for doing the research

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
baywet: I won't fight coding preferences. It seems that I'm the only one on the team to like one liners with no brackets 😅

darrelmiller (approved):

PR #611, opened 2021-07-03
Made OpenApiVisitorBase Enter() and Exit() virtual, plus test

peteraritchie:
I was working on some processing ideas and using the decorator pattern to chain together different algorithms.  I realized that a decorator can't pass through to all the methods the walker would call.  So, a decorated class doesn't get all the information that an undecoreated class (i.e. `Enter` and `Exit`, `PathString` and `CurrentKeys`.

This pull request makes `Enter` and `Exit` virtual and includes a test to make sure the extra two virtual methods an involved.


darrelmiller:
I like this.  I still somewhat regret breaking the walker and visitor into two classes, but this helps.

peteraritchie:
> I like this. I still somewhat regret breaking the walker and visitor into two classes, but this helps.

Would it still be a visitor if the object structure wasn't separated from the algorithm? ;)

PR #621, opened 2021-08-25
Allow `exclusiveMinimum and exclusiveMaximum` as numbers as per newer versions of JSON schema

martinothamar:
The OpenAPI spec for 3.0 spec uses a draft version of the jsonschema spec where `exclusiveMinimum` is a boolean type. In later versions it has become a number. We parse OpenAPI documents generated by libraries in other languages, for instance FastAPI in Python. There is a bug in FastAPI OpenAPI document generation where it apparantly uses a later version of jsonschema spec leading to `exclusiveMinimum` having a number value: https://github.com/tiangolo/fastapi/issues/240

I drafted up this PR just to see if it would be possible to support the implementation of FastAPI. I'm not an OpenAPI expert, so does this look feasible at all?


darrelmiller:
OpenAPI v3 is pinned to JSON Schema draft-04.  When we implement OpenAPI 3.1, we will be decoupled from JSON Schema and support anything from JSON Schema 2020-12 onward.

PR #622, opened 2021-08-26
Added extensions generating for ref properties of definitions

win7user10:
Properties of ref type of definitions do not contain extensions though other properties do (it mean "stringProperty" will generate "extension1" property but "refProperty" will not).

```json 
"definitions": {
  "SomeType": {
    "properties": {
      "stringProperty": {
        "format": "string",
        "type": "string",
        "extension1": "value1"
      },
      "refProperty": {
        "$ref": "#/definitions/SomeRefProperty",
      }
    }
  }
}
```

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiReference.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiReference.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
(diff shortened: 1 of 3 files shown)


darrelmiller:
This was by design.  Extensions alongside a $ref are ignored  https://spec.openapis.org/oas/v3.0.0#fixed-fields-18

In 3.1 you can add properties alongside a JSON Schema $ref but not alongside a OpenAPI Reference object.  We will add support in 3.1 for this.

Issue #651, opened 2021-11-16
hidi should use default values based on the input document not the hardcoded ones.


darrelmiller:
Currently if you try to translate a YAML input file from V2 to V3, it will output as JSON because that is the default, unless you explicitly set YAML as the output.   This is probably not expected behavior. 

We should detect the defaults from the input file.

- Can we determine what the input format is?
- By the time we have determined what the import format and input versions, do we still know what the input parameters were?

PR #658, opened 2021-12-06
Use input file OpenApi format / version as the default output format / version in Hidi

danielmbaluka:
Issue: https://github.com/microsoft/OpenAPI.NET/issues/651

The  current `transform` command in Hidi defaults to `JSON` as the output format when no explicit format is specified. This is not the intended behaviour in most  cases.

This PR improves the `transform` command by using the default input file format and OpenApi version as the output format and version when no format or version is explicitly specified.

### DIFF
--- a/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
+++ b/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
@@ -23,8 +23,8 @@ namespace Microsoft.OpenApi.Hidi
         public static void ProcessOpenApiDocument(
             string input,
             FileInfo output,
-            OpenApiSpecVersion version,
-            OpenApiFormat format,
+            OpenApiSpecVersion? version,
+            OpenApiFormat? format,
             string filterByOperationIds,
             string filterByTags,
             string filterByCollection,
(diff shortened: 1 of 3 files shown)


MaggieKimani1 (commented):
Should we do the same for OpenApi version too?

danielmbaluka:
> Should we do the same for OpenApi version too?

I think we would want to. cc @darrelmiller

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -22,7 +22,7 @@ static class OpenApiService
             string input,
             FileInfo output,
             OpenApiSpecVersion version,
-            OpenApiFormat format,
```
darrelmiller, line 25: Try to use OpenApiFormat?  (nullable OpenApiFormat) so that you can distinguish between being sent a parameter or not.

danielmbaluka: @darrelmiller that's a good suggestion. I adopted it. Thank you.

danielmbaluka:
> Should we do the same for OpenApi version too?

@MaggieKimani1 @darrelmiller this has been implemented. Thank you.

MaggieKimani1 (approved):

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -198,5 +201,17 @@ internal static void ValidateOpenApiDocument(string input)
 
             Console.WriteLine(statsVisitor.GetStatisticsReport());
         }
+
+        private static OpenApiFormat GetOpenApiFormat(string input)
+        {
+            if (!input.StartsWith("http") && Path.GetExtension(input) == ".json")
```
irvinesunday, line 207:
```suggestion
            return !input.StartsWith("http") && Path.GetExtension(input) == ".json" ? OpenApiFormat.Json : OpenApiFormat.Yaml;
```
You can replace the entire `if` with a ternary operator statement.

danielmbaluka: Good suggestion. Thank you.

Issue #655, opened 2021-11-24
Add PatternProperties to OpenAPISchema


darrelmiller:
Need to add PatternProperties to OpenAPISchema
http://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.2.2

darrelmiller:
I guess I need to read the spec more closely.  PatternProperties isn't supported in V3.

PR #661, opened 2021-12-07
Added PatternProperties to OpenApiSchema

danielmbaluka:
This PR add adds "patternProperties" to the OpenApiSchema
Reference: [PatternPropeties in json-schema](http://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.2.2)
Note: PatternProperties are not supported in[ OpenAPI v2](https://swagger.io/docs/specification/data-models/keywords/)

Seeks to close https://github.com/microsoft/OpenAPI.NET/issues/655

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -165,6 +165,15 @@ namespace Microsoft.OpenApi.Models
 
         /// <summary>
         /// Follow JSON Schema definition: https://tools.ietf.org/html/draft-fge-json-schema-validation-00
+        /// PatternProperty definitions MUST be a Schema Object and not a standard JSON Schema (inline or referenced)
+        /// Each property name of this object SHOULD be a valid regular expression according to the ECMA 262 r
+        /// egular expression dialect. Each property value of this object MUST be an object, and each object MUST 
+        /// be a valid Schema Object not a standard JSON Schema.
+        /// </summary>
+        public IDictionary<string, OpenApiSchema> PatternProperties { get; set; } = new Dictionary<string, OpenApiSchema>();
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -354,6 +363,9 @@ public void SerializeAsV3WithoutReference(IOpenApiWriter writer)
             // properties
             writer.WriteOptionalMap(OpenApiConstants.Properties, Properties, (w, s) => s.SerializeAsV3(w));
 
+            // patternProperties
+            writer.WriteOptionalMap(OpenApiConstants.PatternProperties, PatternProperties, (w, s) => s.SerializeAsV3(w));
+
```
darrelmiller, line 368: Does this not also needed to be added in SerializeAsV2WithoutReference ?

danielmbaluka: @darrelmiller patternProperties field is not supported in OpenAPI v2 (https://swagger.io/docs/specification/data-models/keywords/), so I didn't add it.

darrelmiller: It turns out, neither does OpenAPI v3, which would explain why we didn't implement it.  It is however supported in v3.1, so lets hold this PR until V3.1 which we should be working on in January.
https://spec.openapis.org/oas/v3.0.3#properties

danielmbaluka: Sure, we will hold the PR until we start on v3.1

MaggieKimani1:
Superseded by https://github.com/microsoft/OpenAPI.NET/pull/1563

PR #727, opened 2022-02-08
Catch any OpenApiExceptions thrown during parsing and continue processing Schema Components

### LINKED ISSUE #698: Fails to gracefully degrade when hitting certain parsing errors
This OpenAPI https://uk.api.just-eat.io/docs fails to parse properly.

This component is invalid and we should just fail to parse it.  However, it stops the processing of the rest of the schema components.  We should trap the error and continue.
```yaml
  Index:
      $ref: '#/components/schemas'
```

MaggieKimani1:
Fixes #698

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V3/OpenApiV3VersionService.cs
+++ b/src/Microsoft.OpenApi.Readers/V3/OpenApiV3VersionService.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
 using System;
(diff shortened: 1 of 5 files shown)


src/Microsoft.OpenApi.Readers/V3/OpenApiV3VersionService.cs
```diff
@@ -88,7 +88,14 @@ internal class OpenApiV3VersionService : IOpenApiVersionService
...
                     {
                         // "$ref": "#/components/schemas/Pet"
-                        return ParseLocalReference(segments[1]);
+                        try
+                        {
+                            return ParseLocalReference(segments[1]);
+                        }
+                        catch (OpenApiException)
+                        {
+                            return null;
```
darrelmiller, line 97: We need to add an error to the errors collection in the diagnostics object.

PR #744, opened 2022-02-16
Migrate all projects to .NET6

### LINKED ISSUE #730: Upgrade Projects to use .NET 6

MaggieKimani1:
Fixes #730


test/Microsoft.OpenApi.SmokeTests/Microsoft.OpenApi.SmokeTests.csproj
```diff
@@ -1,11 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net48</TargetFrameworks>
+    <TargetFrameworks>net60</TargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <DefineConstants>TRACE;DEBUG;NET48</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;NET60</DefineConstants>
```
baywet, line 8: should that mean we have code-updates to make? if the constant was not used, just remove it

src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.0</TargetFrameworks>
+        <TargetFrameworks>net6.0</TargetFrameworks>
         <LangVersion>9.0</LangVersion>
```
baywet, line 4:
```suggestion
        <LangVersion>Latest</LangVersion>
```

src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.0</TargetFrameworks>
+        <TargetFrameworks>net6.0</TargetFrameworks>
```
baywet, line 3: this would mean we drop support for anything else. No more dotnet framework , just net 6 and up.
I personally don't have issues with that, I just wanted to call it out.
If we go this route, we should align the openapi.odata lib as well https://github.com/microsoft/OpenAPI.NET.OData/issues/139

The alternative would be (for this lib and the readers lib) to cross compile for `netstandard2.0;net462` to keep compatibility with .net framework folks if there's no new API we're using.

baywet (changes requested):
thanks for starting the work here

src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
MaggieKimani1: Thanks for pointing this out. I think the safest outcome would be to support cross compilation. Resolving..

PR #745, opened 2022-02-16
Adds Constants for the Scheme and Bearer format types

### LINKED ISSUE #632: Consider adding constants for scheme and bearer
Consider adding constants for various schemes and bearer formats.

From:
```cs
c.AddSecurityDefinition("bearerAuth", new OpenApiSecurityScheme
{
    Type = SecuritySchemeType.Http,
    Scheme = "bearer", // string
    BearerFormat = "JWT", // string
    Description = "JWT Authorization header using the Bearer scheme."
});
```

To:
```cs
c.AddSecurityDefinition("bearerAuth", new OpenApiSecurityScheme
{
    Type = SecuritySchemeType.Http,
    Scheme = Scheme.Bearer, // constant
    BearerFormat = BearerFormat.Jwt, // constant
    Description = "JWT Authorization header using the Bearer scheme."
});
```

MaggieKimani1:
Fixes #632


src/Microsoft.OpenApi/Models/AuthenticationScheme.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT license. 
+
+using Microsoft.OpenApi.Attributes;
+
+namespace Microsoft.OpenApi.Models
+{
+    /// <summary>
+    /// The type of HTTP authentication schemes
+    /// </summary>
+    public enum AuthenticationScheme
```
darrelmiller, line 11: This is a thing of beauty.  However, we can't do it this way.  The scheme property is defined as a string in the spec https://spec.openapis.org/oas/v3.0.0#fixed-fields-22 and the [IANA HTTP Auth Scheme](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml) registry defines the allow values.  That registry can have new values added at any time.
Also, changing a public member from string to enum is a breaking change, so we couldn't do it in 1.3.

src/Microsoft.OpenApi/Models/AuthenticationScheme.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Use bearer Authentication scheme
+        /// </summary>
+        [Display("bearer")] Bearer,
+
+        /// <summary>
+        /// Use OpenIdConnectUrl
+        /// </summary>
+        [Display("openIdConnectUrl")] OpenIdConnectUrl
```
darrelmiller, line 26: This is an odd one.  This isn't an HTTP auth scheme.    This is a property on the OpenAPI Security Scheme object.

src/Microsoft.OpenApi/Models/AuthenticationScheme.cs
darrelmiller: I think it should be sufficient to add Bearer and Basic as constants in this file https://github.com/microsoft/OpenAPI.NET/blob/649f0b1780c575c4780f4091eedd7a555e8ccc4a/src/Microsoft.OpenApi/Models/OpenApiConstants.cs

darrelmiller (changes requested):
Replace enums with consts

src/Microsoft.OpenApi/Models/AuthenticationScheme.cs
MaggieKimani1: Thank you for pointing this out. Resolving..

test/Microsoft.OpenApi.Tests/Models/OpenApiComponentsTests.cs
```diff
@@ -57,7 +57,7 @@ public class OpenApiComponentsTests
                 {
                     Description = "description1",
                     Type = SecuritySchemeType.OpenIdConnect,
-                    Scheme = "openIdConnectUrl",
+                    Scheme = OpenApiConstants.OpenIdConnectUrl,
```
darrelmiller, line 60: Remove this constant. It isn't required.

darrelmiller: Set the scheme to be bearer

darrelmiller (approved):

PR #753, opened 2022-02-28
 - upgrades odata conversion lib reference in hidi

baywet:


src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -198,16 +198,20 @@ string filterbycollection
...
             var edmModel = CsdlReader.Parse(XElement.Parse(csdlText).CreateReader());
 
             var settings = new OpenApiConvertSettings()
             {
+                AddSingleQuotesForStringParameters = true,
+                AddEnumDescriptionExtension = true,
+                DeclarePathParametersOnPathItem = true,
                 EnableKeyAsSegment = true,
                 EnableOperationId = true,
+                ErrorResponsesAsDefault  = false,
```
MaggieKimani1, line 214: Are these updated settings Kiota-specific or will they apply to all clients?

MaggieKimani1 (approved):

PR #767, opened 2022-03-03
Simplify openapi spec version for Hidi

### LINKED ISSUE #732: Simplify The OpenAPI version format supplied in Hidi

MaggieKimani1:
Fixes #732

### DIFF
--- a/src/Microsoft.OpenApi.Hidi/OpenApiSpecVersionExtension.cs
+++ b/src/Microsoft.OpenApi.Hidi/OpenApiSpecVersionExtension.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System;
+using System.Linq;
+
+namespace Microsoft.OpenApi.Hidi
+{
+    public static class OpenApiSpecVersionExtension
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi.Hidi/OpenApiSpecVersionExtension.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Linq;
+
+namespace Microsoft.OpenApi.Hidi
+{
+    public static class OpenApiSpecVersionExtension
+    {
+        public static OpenApiSpecVersion TryParseOpenApiSpecVersion(this string value)
```
darrelmiller, line 11: Extension methods on very commonly used types like string are not usually recommended.  If every use of a string did this it would hugely increase the surface are of the String class.  I would suggest just renaming this to OpenApiSpecVersionHelper and accept the string as a regular parameter.

MaggieKimani1: Thank you for pointing this out. Resolving...

darrelmiller (approved):

PR #776, opened 2022-03-07
Use await keyword in our asynchronous operation to prevent deadlocks

### LINKED ISSUE #775: Use await instead of GetAwaiter.GetResult
As the command handler is called asynchronously, we can use async/await instead of  GetAwaiter.GetResult.  I suspect this may be the source of some deadlocks we are seeing.

https://github.com/microsoft/OpenAPI.NET/blob/9ce2b2750324112a4c90ba4178d5957a668dce41/src/Microsoft.OpenApi.Hidi/OpenApiService.cs#L109

MaggieKimani1:
Fixes #775


darrelmiller:
Made these changes in PR #767 to avoid having to deal with merge conflicts twice.

PR #785, opened 2022-03-08
Better error handling

### LINKED ISSUE #784: Better Exception handling in Hidi
Reference : https://github.com/microsoft/kiota/issues/1269, https://github.com/microsoft/kiota/pull/1344

MaggieKimani1:
Fixes #784
- Passes the cancellation token from the system.command line host to async APIs to avoid getting task cancelled exceptions when users press ctrl + c
- Formats the logging to use templates instead of doing string interpolation so logging APIs can do sampling and metrics

### DIFF
--- a/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
+++ b/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
@@ -128,10 +128,13 @@ namespace Microsoft.OpenApi.Hidi
                 var context = result.OpenApiDiagnostic;
                 if (context.Errors.Count > 0)
                 {
+                    logger.LogTrace("{timestamp}ms: Parsed OpenAPI with errors. {count} paths found.", stopwatch.ElapsedMilliseconds, document.Paths.Count);
+
                     var errorReport = new StringBuilder();
 
                     foreach (var error in context.Errors)
                     {
+                        logger.LogError("OpenApi Parsing error: {message}", error.ToString());
                         errorReport.AppendLine(error.ToString());
                     }
                     logger.LogError($"{stopwatch.ElapsedMilliseconds}ms: OpenApi Parsing errors {string.Join(Environment.NewLine, context.Errors.Select(e => e.Message).ToArray())}");
@@ -282,9 +285,9 @@ namespace Microsoft.OpenApi.Hidi
                 catch (HttpRequestException ex)
                 {
 #if DEBUG
-                    logger.LogCritical(ex, $"Could not download the file at {input}, reason: {ex.Message}");
+                    logger.LogCritical(ex, "Could not download the file at {inputPath}, reason: {exMessage}", input, ex.Message);
 #else
-                    logger.LogCritical($"Could not download the file at {input}, reason: {ex.Message}", input, ex.Message);
+                    logger.LogCritical( "Could not download the file at {inputPath}, reason: {exMessage}", input, ex.Message);
 #endif
                     return null;
                 }
@@ -305,9 +308,9 @@ namespace Microsoft.OpenApi.Hidi
                     ex is NotSupportedException)
                 {
 #if DEBUG
-                    logger.LogCritical(ex, $"Could not open the file at {input}, reason: {ex.Message}");
+                    logger.LogCritical(ex, "Could not open the file at {inputPath}, reason: {exMessage}", input, ex.Message);
 #else
-                    logger.LogCritical($"Could not open the file at {input}, reason: {ex.Message}");
+                    logger.LogCritical("Could not open the file at {inputPath}, reason: {exMessage}", input, ex.Message);
 #endif
                     return null;
                 }
--- a/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
+++ b/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
@@ -49,42 +49,18 @@ namespace Microsoft.OpenApi.Hidi
             {
                 if (string.IsNullOrEmpty(openapi) && string.IsNullOrEmpty(csdl))
                 {
-                    throw new ArgumentNullException("Please input a file path");
+                    throw new ArgumentException("Please input a file path");
                 }
-            }
-            catch (ArgumentNullException ex)
-            {
-#if DEBUG
-                logger.LogCritical(ex, ex.Message);
-#else
-                logger.LogCritical(ex.Message);
-#endif
-                return;
-            }
-            try
-            {
                 if(output == null)
                 {
-                    throw new ArgumentException(nameof(output));
+                    throw new ArgumentNullException(nameof(output));
                 }
-            }
-            catch (ArgumentException ex)
-            {
-#if DEBUG
-                logger.LogCritical(ex, ex.Message);
-#else
-                logger.LogCritical(ex.Message);
-#endif
-                return;
-            }
-            try
-            {
                 if (output.Exists)
                 {
                     throw new IOException($"The file {output} already exists. Please input a new file path.");
                 }
             }
-            catch (IOException ex)
+            catch (Exception ex)
             {
 #if DEBUG  
                 logger.LogCritical(ex, ex.Message);
--- a/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
+++ b/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
@@ -353,7 +353,6 @@ namespace Microsoft.OpenApi.Hidi
                     foreach (var error in context.Errors)
                     {
                         logger.LogError("OpenApi Parsing error: {message}", error.ToString());
-                        Console.WriteLine(error.ToString());
                     }
                 }
 
@@ -362,7 +361,7 @@ namespace Microsoft.OpenApi.Hidi
                 walker.Walk(document);
 
                 logger.LogTrace("Finished walking through the OpenApi document. Generating a statistics report..");
-                Console.WriteLine(statsVisitor.GetStatisticsReport());
+                logger.LogInformation(statsVisitor.GetStatisticsReport());
             }
             catch(Exception ex)
             {
@@ -371,6 +370,7 @@ namespace Microsoft.OpenApi.Hidi
 #else
                 logger.LogCritical(ex.Message);
 #endif
+                return;
             }
 
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -263,11 +277,15 @@ private static async Task<Stream> GetStream(string input, ILogger logger)
...
                       DefaultRequestVersion = HttpVersion.Version20
                     };
-                    stream = await httpClient.GetStreamAsync(input);
+                    stream = await httpClient.GetStreamAsync(input, cancellationToken);
                 }
                 catch (HttpRequestException ex)
                 {
-                    logger.LogError($"Could not download the file at {input}, reason{ex}");
+#if DEBUG
+                    logger.LogCritical(ex, $"Could not download the file at {input}, reason: {ex.Message}");
```
baywet, line 285:
```suggestion
                    logger.LogCritical(ex, "Could not download the file at {input}, reason: {exMessage}", input, ex.Message);
```

(and repeat for all the LogXXX calls to remove string interpolations)

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -52,7 +54,11 @@ string filterbycollection
             }
             catch (ArgumentNullException ex)
             {
-                logger.LogError(ex.Message);
+#if DEBUG
```
baywet, line 57: just as a quick though, putting those conditional compilation directives all the way to the left might be because I'm a dinosaur and that's it
GrandmaDinosaurGIF

back in 2008 those directives would not work when indented, but I see that you originally had indented versions of them that seem to be working properly

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -327,14 +349,14 @@ private static async Task<Stream> GetStream(string input, ILogger logger)
             return requestUrls;
         }
 
-        internal static async Task ValidateOpenApiDocument(string openapi, LogLevel loglevel)
+        internal static async Task ValidateOpenApiDocument(string openapi, LogLevel loglevel, CancellationToken cancellationToken)
         {
```
baywet, line 353: you also need a root try catch for this command handler

baywet (changes requested):
thanks for starting the work here, a couple of comments.

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
MaggieKimani1: I think even in the [docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives) the preprocessor directives aren't indented.

baywet: ah, good, so I'm not the last of my kind

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -64,19 +70,27 @@ string filterbycollection
...
             }
             try
             {
                 if (output.Exists)
                 {
-                    throw new IOException("The file you're writing to already exists. Please input a new file path.");
+                    throw new IOException($"The file {output} already exists. Please input a new file path.");
                 }
             }
             catch (IOException ex)
```
baywet, line 87: sorry, I missed something on my initial review. Instead of having a bunch of try catches in this method, move to a single one that encapsulates the whole body, it'll make the code much smaller, and easier to read

MaggieKimani1: Yeah that makes sense. Resolving..

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -47,36 +49,24 @@ string filterbycollection
...
-            catch (IOException ex)
+            catch (Exception ex)
             {
-                logger.LogError(ex.Message);
+#if DEBUG  
+                logger.LogCritical(ex, ex.Message);
+#else
+                logger.LogCritical(ex.Message);
+#endif
                 return;
```
baywet, line 70:
```suggestion
                return 1;
```
And update the method prototype to Task<int> so the exit code can be captured by system.commandline

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -327,37 +328,51 @@ private static async Task<Stream> GetStream(string input, ILogger logger)
...
+                    RuleSet = ValidationRuleSet.GetDefaultRuleSet()
+                }
+                ).Read(stream, out var context);
+
+                if (context.Errors.Count != 0)
+                {
+                    foreach (var error in context.Errors)
+                    {
+                        logger.LogError("OpenApi Parsing error: {message}", error.ToString());
+                        Console.WriteLine(error.ToString());
```
baywet, line 356:
```suggestion
```

You're already using the logger, we should not have anymore console.log in the code base

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -263,11 +256,15 @@ private static async Task<Stream> GetStream(string input, ILogger logger)
...
                       DefaultRequestVersion = HttpVersion.Version20
                     };
-                    stream = await httpClient.GetStreamAsync(input);
+                    stream = await httpClient.GetStreamAsync(input, cancellationToken);
                 }
                 catch (HttpRequestException ex)
                 {
-                    logger.LogError($"Could not download the file at {input}, reason{ex}");
+#if DEBUG
+                    logger.LogCritical(ex, "Could not download the file at {inputPath}, reason: {exMessage}", input, ex.Message);
```
baywet, line 264:
```suggestion
                    throw new InvalidOperationException($"Could not download the file at {inputPath}, ex);
```
And clean the rest, replicate for a local file. Since this is a private method, and since we now have global catches that handle properly logging any exception, it's better to let the exception bubble up. (less code duplication, debugger will be able to stop here, etc...)

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -327,37 +328,51 @@ private static async Task<Stream> GetStream(string input, ILogger logger)
             return requestUrls;
         }
 
-        internal static async Task ValidateOpenApiDocument(string openapi, LogLevel loglevel)
+        internal static async Task ValidateOpenApiDocument(string openapi, LogLevel loglevel, CancellationToken cancellationToken)
```
baywet, line 331:
```suggestion
        internal static async Task<int> ValidateOpenApiDocument(string openapi, LogLevel loglevel, CancellationToken cancellationToken)
```

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -327,37 +328,51 @@ private static async Task<Stream> GetStream(string input, ILogger logger)
...
+                Console.WriteLine(statsVisitor.GetStatisticsReport());
+            }
+            catch(Exception ex)
+            {
+#if DEBUG
+                logger.LogCritical(ex, ex.Message);
+#else
+                logger.LogCritical(ex.Message);
+#endif
+            }
```
baywet, line 374:
```suggestion
               return 1;
            }
```

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -327,37 +328,51 @@ private static async Task<Stream> GetStream(string input, ILogger logger)
...
 
-            var statsVisitor = new StatsVisitor();
-            var walker = new OpenApiWalker(statsVisitor);
-            walker.Walk(document);
+                var statsVisitor = new StatsVisitor();
+                var walker = new OpenApiWalker(statsVisitor);
+                walker.Walk(document);
+
+                logger.LogTrace("Finished walking through the OpenApi document. Generating a statistics report..");
+                Console.WriteLine(statsVisitor.GetStatisticsReport());
```
baywet, line 365: this should use the logger instead

baywet (changes requested):

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
MaggieKimani1: Why `return 1` instead of `return`?

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
MaggieKimani1: Why are we returning an int?

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
MaggieKimani1: Do you mean `Task<int>`?

baywet: yep

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
baywet: so the caller knows the process terminated with an error
exit code: 0 finished and successfull, anything else is error.
this is how a script, CI (ADO, github actions...) knows the process ran into an error.
https://github.com/dotnet/command-line-api/blob/main/docs/Process-termination-handling.md

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
baywet: yep

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -47,150 +49,143 @@ string filterbycollection
...
                     {
-                        errorReport.AppendLine(error.ToString());
+                        logger.LogTrace("{timestamp}ms: Parsed OpenAPI with errors. {count} paths found.", stopwatch.ElapsedMilliseconds, document.Paths.Count);
+
+                        var errorReport = new StringBuilder();
+
+                        foreach (var error in context.Errors)
+                        {
+                            logger.LogError("OpenApi Parsing error: {message}", error.ToString());
+                            errorReport.AppendLine(error.ToString());
```
baywet, line 104:
```suggestion
                            errorReport.AppendLine(error);
```

PR #831, opened 2022-04-13
Add --terse-output Commandline parameter for producing JSON in terse format

### LINKED ISSUE #829: Enable Hidi users to create "terse" JSON output
Using the Terse option in the OpenAPIJsonWriter we can elable users to emit JSON in a terse format that removes unnecessary whitespace and produces a significantly smaller output file.  This file is also faster to load when processing.

MaggieKimani1:
Fixes #829 
Adds a --terse-ouput parameter, abbreviated as -to in Hidi to emit terse JSON output.

#### Usage:
`transform -d C:\pestore.yaml -f json -o C:\pestore.json -co -to`

### DIFF
--- a/test/Microsoft.OpenApi.Tests/PublicApi/PublicApi.approved.txt
+++ b/test/Microsoft.OpenApi.Tests/PublicApi/PublicApi.approved.txt
@@ -1297,7 +1297,7 @@ namespace Microsoft.OpenApi.Writers
     public class OpenApiJsonWriterSettings : Microsoft.OpenApi.Writers.OpenApiWriterSettings
     {
         public OpenApiJsonWriterSettings() { }
-        public bool Terse { get; set; }
+        public new bool Terse { get; set; }
     }
     public static class OpenApiWriterAnyExtensions
     {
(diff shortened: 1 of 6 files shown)


test/Microsoft.OpenApi.Tests/PublicApi/PublicApi.approved.txt
```diff
@@ -1379,6 +1379,7 @@ namespace Microsoft.OpenApi.Writers
         public bool InlineLocalReferences { get; set; }
         [System.Obsolete("Use InlineLocalReference and InlineExternalReference settings instead")]
         public Microsoft.OpenApi.Writers.ReferenceInlineSetting ReferenceInline { get; set; }
+        public bool Terse { get; set; }
```
darrelmiller, line 1382: Why is this moved up into OpenApiWriterSettings?  Terse is only something that makes sense for JSON.  You can't remove whitespace from YAML.

MaggieKimani1: Thank you for the correction, resolving...
I think the confusion was because while switching the format to JSON before writing out, the OpenApiJsonWriter consumes the OpenApiWriter settings to do the serialization hence I thought it easier to add a Terse parameter in those settings. I've updated the PR with a better approach factoring in your comment.

darrelmiller (approved):

Issue #537, opened 2020-12-14
Make OpenApiSchema partial or implement IEquatable for all classes in the model hierarchy?


OlliMartin:
Hi, 

I'm having quite some issues with comparing OpenApiSchema objects to other ones.
My workaround for now is to create Extension Methods to compare the hierarchy against another one. 
(Which leads to a different set of problems)

What are your thoughts about implenting at least IEquatable<T|OpenApiSchema> for OpenApiSchema to allow this, or at least make it partial so I don't have to write some weird workarounds inside of Extension methods?

Kind regards & thanks for your work
Olli

darrelmiller:
This sounds like a great idea to me.  ICloneable would be really useful.  IEquatable is partially useful.  We already have two submissions for comparison capabilities.  I want to be really careful about creating even more code to maintain.  
Would a comparision capability meet your needs for IEquatable?

drpohlmann:
Comparison is totally enough. 

I quickly looked through some of the classes, is there any suggested approach of how to implement ICloneable? 
For the model I think it should be mostly straight forward, but the Any models (IOpenApiAny) could be a bit trickier.

I will try to create a feat branch today and see how it's going.

OlliMartin:
Comparison is totally enough.

I quickly looked through some of the classes, is there any suggested approach of how to implement ICloneable?
For the model I think it should be mostly straight forward, but the Any models (IOpenApiAny) could be a bit trickier.

I will try to create a feat branch today and see how it's going.

darrelmiller:
Ugh, I really don't like those OpenApiAny things.  I'll take a look and see if I have any suggestions.

baywet:
A couple of updates:
- We've replaced the OpenApiAny infrastructure with JsonNode from System.Text.Json, this simplifies a lot of things and provides a much better experience.
- We've implemented copy (shallow) constructors.
- For the remaining aspect, the comparison, there's already an issue tracking this. I'll close this issue as duplicated so we can have a federated conversation over there.

PR #853, opened 2022-05-11
Adds support for Copy Constructors

### LINKED ISSUE #832: Add support for copy constructors to Microsoft.OpenApi.Models objects
As outlined in https://github.com/dotnet/aspnetcore/issues/40676 and implemented in https://github.com/dotnet/aspnetcore/pull/41238, we're introducing support for a new `Microsoft.AspNetCore.OpenApi` package that acts as a link between `Microsoft.OpenApi` and route handler endpoints.

This package exposes a `WithOpenApi` extension method that users can invoke on their endpoints to mutate the `OpenApiOperation` that we generate by default from the handler's `MethodInfo` with their own annotation. For example,

```csharp
app.MapGet("/foo", () => {})
.WithOpenApi(generatedOperation => {
  generatedOperation.OperationId = "MadeByMe";
  return generatedOperation;
});
```

One of the things we thought would be compelling was support for a copy constructor for `OpenApiOperation` so that users can invoke the following and have their modification automatically merged into the generated operation.

```csharp
app.MapGet("/foo", () => {})
.WithOpenApi(generatedOperation => new(generatedOperation) {
  OperationId = "MadeByMe"
});
```

MaggieKimani1:
Fixes #832


baywet (commented):
wouldn't it make more sense to implement [ICloneable](https://docs.microsoft.com/en-us/dotnet/api/system.icloneable?view=net-6.0) semantically instead?
Also I noticed that in every copy constructor the code is just copying the reference for properties, instead of making a copy of the object. That means that a change on `myOriginal.PropA.PropAlpha = "foo"` would also impact the copy ` myCopy.PropA.PropAlpha`  is now also valued at "foo". This could have tons of side effects.
See how I implemented it for Kiota https://github.com/microsoft/kiota/blob/d19d14e06bc6fcadac91d208d2bf1c54c1a8918c/src/Kiota.Builder/CodeDOM/CodeMethod.cs#L160

MaggieKimani1:
> wouldn't it make more sense to implement [ICloneable](https://docs.microsoft.com/en-us/dotnet/api/system.icloneable?view=net-6.0) semantically instead? Also I noticed that in every copy constructor the code is just copying the reference for properties, instead of making a copy of the object. That means that a change on `myOriginal.PropA.PropAlpha = "foo"` would also impact the copy ` myCopy.PropA.PropAlpha` is now also valued at "foo". This could have tons of side effects. See how I implemented it for Kiota https://github.com/microsoft/kiota/blob/d19d14e06bc6fcadac91d208d2bf1c54c1a8918c/src/Kiota.Builder/CodeDOM/CodeMethod.cs#L160

Does deriving from `ICloneable` provide a deep copy of the object including all the properties?

MaggieKimani1:
> wouldn't it make more sense to implement [ICloneable](https://docs.microsoft.com/en-us/dotnet/api/system.icloneable?view=net-6.0) semantically instead? Also I noticed that in every copy constructor the code is just copying the reference for properties, instead of making a copy of the object. That means that a change on `myOriginal.PropA.PropAlpha = "foo"` would also impact the copy ` myCopy.PropA.PropAlpha` is now also valued at "foo". This could have tons of side effects. See how I implemented it for Kiota https://github.com/microsoft/kiota/blob/d19d14e06bc6fcadac91d208d2bf1c54c1a8918c/src/Kiota.Builder/CodeDOM/CodeMethod.cs#L160

Do you mind expounding on the side effects?
Also implementing `ICloneable` semantically as you've done in Kiota would require type-casting when one is trying to mutate the cloned object?

matt9ucci:
Seems to be related with #537. I hope `ICloneable`, no side effects and no hierarchy problems, is implemented for all models.

src/Microsoft.OpenApi/Models/OpenApiError.cs
```diff
@@ -26,6 +26,15 @@ public OpenApiError(string pointer, string message)
             Message = message;
         }
 
+        /// <summary>
+        /// Initializes a copy of an <see cref="OpenApiError"/> object
+        /// </summary>
+        public OpenApiError(OpenApiError error)
```
baywet, line 32: this one got forgotten during the last batch of changes

src/Microsoft.OpenApi/Models/OpenApiCallback.cs
```diff
@@ -34,6 +34,20 @@ public class OpenApiCallback : IOpenApiSerializable, IOpenApiReferenceable, IOpe
         /// This object MAY be extended with Specification Extensions.
         /// </summary>
         public IDictionary<string, IOpenApiExtension> Extensions { get; set; } = new Dictionary<string, IOpenApiExtension>();
+        
+        /// <summary>
+        /// Initializes a copy of an <see cref="OpenApiCallback"/> object
+        /// </summary>
+        public object Clone()
```
baywet, line 41: the interface implementation needs to be declared (: ICloneable in the class declaration) for all models

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -64,6 +62,29 @@ public class OpenApiDocument : IOpenApiSerializable, IOpenApiExtensible
         /// </summary>
         public IDictionary<string, IOpenApiExtension> Extensions { get; set; } = new Dictionary<string, IOpenApiExtension>();
 
+        /// <summary>
+        /// Parameter-less constructor
+        /// </summary>
+        public OpenApiDocument() {}
+
```
baywet, line 69:
```suggestion
```

src/Microsoft.OpenApi/Models/OpenApiCallback.cs
```diff
@@ -34,6 +34,20 @@ public class OpenApiCallback : IOpenApiSerializable, IOpenApiReferenceable, IOpe
...
+        {
+            return new OpenApiCallback
+            {
+                PathItems = PathItems,
```
baywet, line 45: all the properties need to use the clone method before being assigned to achieve a deep copy (this only has been done for strings at the moment). Right now the code is making reference copies which could have side effects as described in an earlier comment.

baywet (changes requested):
thanks for making updates, still a few changes required

darrelmiller:
The API design that was approved for minimal APIs was to be able to use copy constructors.  By having copy constructors it allows creating a new instance of the object and then using object initializers to set values.  e.g.

```csharp
app.MapGet("/foo", () => {})
.WithOpenApi(generatedOperation => new(generatedOperation) {
  OperationId = "MadeByMe"
});
```

It is not possible to do the same with ICloneable.    Can we switch this to use copy constructors?  

Also, it is not necessary to clone strings.  Strings are immutable and so there is no harm in having the clone share a reference to the string.  If one of the properties is updated, it will not affect the other.

baywet:
sorry for the confusion, I didn't have the additional context on this one.

MaggieKimani1:
> 

Thank you for the clarification. I'll revert the PR to the state where I had implemented the copy constructors.

src/Microsoft.OpenApi/Models/OpenApiCallback.cs
```diff
@@ -35,6 +35,22 @@ public class OpenApiCallback : IOpenApiSerializable, IOpenApiReferenceable, IOpe
...
+        /// Parameter-less constructor
+        /// </summary>
+        public OpenApiCallback() { }
+
+        /// <summary>
+        /// Initializes a copy of an <see cref="OpenApiCallback"/> object
+        /// </summary>
+        public OpenApiCallback(OpenApiCallback callback)
+        {
+            PathItems = callback.PathItems;
```
baywet, line 48:
```suggestion
            PathItems = new(callback.PathItems);
```
(and all the other ones, expect for strings)

Thanks for reverting to copy constructor, with the current implementation, you're copying references. That means if a ObjectB is a copy of ObjectA and ObjectA.Foo = "bar", ObjectB.Foo also = "bar", which could have a lot of side effects.

baywet (changes requested):

src/Microsoft.OpenApi/Models/OpenApiMediaType.cs
```diff
@@ -43,6 +43,23 @@ public class OpenApiMediaType : IOpenApiSerializable, IOpenApiExtensible
...
+        /// </summary>
+        public OpenApiMediaType() {}
+
+        /// <summary>
+        /// Initializes a copy of an <see cref="OpenApiMediaType"/> object
+        /// </summary>
+        public OpenApiMediaType(OpenApiMediaType mediaType)
+        {
+            Schema = new(mediaType.Schema);
+            Example = mediaType.Example;
```
baywet, line 57: new ? but to what type? maybe get the constructor by reflection?

baywet (changes requested):

src/Microsoft.OpenApi/Models/OpenApiMediaType.cs
darrelmiller: Maybe here we need to make IOpenApiAny support ICloneable to support this particular case.

src/Microsoft.OpenApi/Models/OpenApiLicense.cs
```diff
@@ -29,6 +29,21 @@ public class OpenApiLicense : IOpenApiSerializable, IOpenApiExtensible
...
+        /// </summary>
+        public OpenApiLicense() {}
+
+        /// <summary>
+        /// Initializes a copy of an <see cref="OpenApiLicense"/> object
+        /// </summary>
+        public OpenApiLicense(OpenApiLicense license)
+        {
+            Name = license.Name;
+            Url = license.Url;
```
darrelmiller, line 43:
```suggestion
            Url = new Uri(license.Url,"");
```
I think this will work.  If not use new Uri(license.Url.OriginalString)

src/Microsoft.OpenApi/Models/RuntimeExpressionAnyWrapper.cs
```diff
@@ -16,6 +16,20 @@ public class RuntimeExpressionAnyWrapper : IOpenApiElement
...
+        /// </summary>
+        public RuntimeExpressionAnyWrapper() {}
+
+        /// <summary>
+        /// Initializes a copy of an <see cref="RuntimeExpressionAnyWrapper"/> object
+        /// </summary>
+        public RuntimeExpressionAnyWrapper(RuntimeExpressionAnyWrapper runtimeExpressionAnyWrapper)
+        {
+            Any = runtimeExpressionAnyWrapper.Any;
+            Expression = runtimeExpressionAnyWrapper.Expression;
```
darrelmiller, line 30: I believe this needs to have a copy constructor also.

MaggieKimani1: The Expression property derives from `RuntimeExpression` that doesn't have a copy constructor as it only has a readonly property `Expression`, hence why I've defaulted to a using a reference copy.

src/Microsoft.OpenApi/Any/OpenApiArray.cs
```diff
@@ -2,20 +2,30 @@
...
         /// </summary>
         public AnyType AnyType { get; } = AnyType.Array;
 
+        /// <summary>
+        /// Implement ICloneable interface to allow for deep copying
+        /// </summary>
+        /// <returns>A new copy of <see cref="OpenApiArray"/></returns>
+        public object Clone()
+        {
+            return new OpenApiArray();
```
baywet, line 26: remove clone methods, add a copy constructor when none is present

src/Microsoft.OpenApi/Any/IOpenApiAny.cs
```diff
@@ -1,14 +1,15 @@
...
+using System;
 using Microsoft.OpenApi.Interfaces;
 
 namespace Microsoft.OpenApi.Any
 {
     /// <summary>
     /// Base interface for all the types that represent Open API Any.
     /// </summary>
-    public interface IOpenApiAny : IOpenApiElement, IOpenApiExtension
+    public interface IOpenApiAny : IOpenApiElement, IOpenApiExtension, ICloneable
```
baywet, line 12: pushing more constraints on an interface is a breaking change, this needs to be reverted

src/Microsoft.OpenApi/Any/OpenApiPrimitive.cs
```diff
@@ -39,6 +50,40 @@ public OpenApiPrimitive(T value)
...
+            if (clone == null) throw new ApplicationException("There's no copy constructor defined");
+            return clone;
+        }
+
+        /// <summary>
+        /// Clones an instance of <see cref="IOpenApiPrimitive"/> object from the copy constructor
+        /// </summary>
+        /// <param name="obj">The object instance.</param>
+        /// <returns>A clone copy.</returns>
+        public static object CloneFromCopyConstructor(Object obj)
```
baywet, line 69:
```suggestion
        public static IOpenApiAny CloneFromCopyConstructor(IOpenApiAny obj)
```

src/Microsoft.OpenApi/Any/OpenApiPrimitive.cs
```diff
@@ -39,6 +50,40 @@ public OpenApiPrimitive(T value)
...
+        {
```
baywet, line 70: Move this method definition to a helper class close to the interface (IOpenApiAny)

src/Microsoft.OpenApi/Any/OpenApiPrimitive.cs
```diff
@@ -39,6 +50,40 @@ public OpenApiPrimitive(T value)
...
+                {
+                    ParameterInfo[] pi = ci.GetParameters();
+                    if (pi.Length == 1 && pi[0].ParameterType == t)
+                    {
+                        return ci.Invoke(new object[] { obj });
+                    }
+                }
+            }
+
+            return null;
```
baywet, line 84: default to returning the current object if you weren't able to find a copy constructor

src/Microsoft.OpenApi/Models/OpenApiExample.cs
```diff
@@ -66,7 +66,7 @@ public OpenApiExample(OpenApiExample example)
         {
             Summary = example.Summary;
             Description = example.Description;
-            Value = example.Value;
+            Value = (IOpenApiAny)example.Value.Clone();
```
baywet, line 69: call the CopyFromCopyConstructor method here

src/Microsoft.OpenApi/Models/OpenApiMediaType.cs
baywet: adding a new interface to the IOpenApiAny definition will be a breaking change for implementers outside of this library. Maggie started down the path of trying to find a copy constructor by reflection and defaulting back to a reference copy which I believe will be more flexible for implementers.

baywet (changes requested):

baywet (approved):

src/Microsoft.OpenApi/Models/RuntimeExpressionAnyWrapper.cs
darrelmiller: Yes, I had forgotten that Expression is an immutable object and so we can copy a reference to it.

darrelmiller (approved):

Issue #866, opened 2022-05-16
Stack overflow exception on running Hidi version 1.0.0-preview.2 and above


andrueastman:
Running hidi version 1.0.0-preview.2 and 1.0.0-preview.3 with this file as input results in a stack overflow exception. 

https://github.com/microsoftgraph/msgraph-metadata/blob/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml

Arguments are 
```shell
transform --csdl .\clean_v10_metadata\cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml --output openapi.yml --version OpenApi3_0 --loglevel Information --format yaml
```

Downgrading to 1.0.0-preview.1 however seems to not result in the same behavior.

The issue also gets fixed when I downgrade the `Microsoft.OpenApi.OData` dependency from `1.0.10` to `1.0.10-preview3` at the head of the `vNext` branch. 

This may need some investigation to understand the root cause.

cc @baywet

PR #868, opened 2022-05-17
Adds OpenAPI convert setting `ExpandDerivedTypesNavigationProperties=false`

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET/issues/866

This PR:
- Adds a new OpenAPI convert setting `ExpandDerivedTypesNavigationProperties=false` to mitigate the generation of multiple navigation property paths for OData type cast paths which cause the StackOverflow exception reported in the aforementioned issue.


baywet (changes requested):
Please bump the preview version in hidi and add a release note (csproj) so we can release shortly after.

baywet (approved):

darrelmiller (approved):

PR #876, opened 2022-05-20
- normalizes inlining parameters to kebab case

baywet:


baywet:
Also added a fix for logging as related to https://github.com/microsoft/kiota/pull/1594

PR #896, opened 2022-06-07
Bumps conversion lib version and Hidi version

irvinesunday:
Updates the `Microsoft.OpenApi.OData` lib version from `1.0.10` to `1.0.11-preview2`
Bumps up Hidi release version to `1.0.0-preview5`


baywet (changes requested):
Can you also update the chagelog on the csproj please?

baywet:
Our merging policies don't require one approval??? If you're admin, can you fix that too please?

Issue #885, opened 2022-06-02
OpenAPI v2 path parameters for a body get lost when reading


darrelmiller:
OpenAPI v2 path parameters should be converted into a requestBody for each Operation that supports requestBodies.

PR #922, opened 2022-07-04
Add support for request body path parameters in v2

darrelmiller:
see #885 

- [x] Added support for reading parameters "in: body"
- [x] Added support for reading parameters "in: formdata"

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. 
 
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.OpenApi.Extensions;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Readers.ParseNodes;
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
@@ -69,16 +69,13 @@ namespace Microsoft.OpenApi.Readers.V2
                 var requestBody = CreateRequestBody(node.Context, bodyParameter);
                 foreach(var opPair in pathItem.Operations)
                 {
-                    if (opPair.Value.RequestBody == null)
+                    switch (opPair.Key)
                     {
-                        switch (opPair.Key)
-                        {
-                            case OperationType.Post:
-                            case OperationType.Put:
-                            case OperationType.Patch:
-                                opPair.Value.RequestBody = requestBody;
-                                break;
-                        }
+                        case OperationType.Post:
+                        case OperationType.Put:
+                        case OperationType.Patch:
+                            opPair.Value.RequestBody = requestBody;
+                            break;
                     }
                 }
             }
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
+using System.Collections.Generic;
 using Microsoft.OpenApi.Extensions;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Readers.ParseNodes;
@@ -32,7 +33,7 @@ namespace Microsoft.OpenApi.Readers.V2
             {
                 "parameters", (o, n) =>
                 {
-                    o.Parameters = n.CreateList(LoadParameter);
+                    LoadPathParameters(o,n);
                 }
             },
         };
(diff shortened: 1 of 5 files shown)


darrelmiller:
I decided not to implement writing re-used request bodies as path parameters in V2 for the following reasons:
- The behaviour would be inconsistent between re-used request bodies and reused parameters
- OpenAPI v3 doesn't support setting the requestBody at the path level and therefore it is a best practice to not use this in V2.  
- An OpenAPI that was read that had identical request bodies across operations would not promote to path parameters, which would lead to inconsistent behaviour depending on how the DOM is constructed.

This means that the way it is currently implemented, we can successfully read request bodies as Path Parameters from V2 documents but they will be converted into operation parameters during the read.

src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
```diff
@@ -53,5 +54,57 @@ public static OpenApiPathItem LoadPathItem(ParseNode node)
...
+                    }
+                }
+            }
+            else
+            {
+                var formParameters = node.Context.GetFromTempStorage<List<OpenApiParameter>>(TempStorageKeys.FormParameters);
+                if (formParameters != null)
+                {
+                    var requestBody = CreateFormBody(node.Context, formParameters);
+                    foreach (var opPair in pathItem.Operations)
```
baywet, line 91:
```suggestion
                    foreach (var opPair in pathItem.Operations.Where(x => x.Value.RequestBody is null))
```

src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
```diff
@@ -53,5 +54,57 @@ public static OpenApiPathItem LoadPathItem(ParseNode node)
...
+                        {
+                            switch (opPair.Key)
+                            {
+                                case OperationType.Post:
+                                case OperationType.Put:
+                                case OperationType.Patch:
+                                    opPair.Value.RequestBody = requestBody;
+                                    break;
+                            }
+                        }
```
baywet, line 103:
```suggestion
                        switch (opPair.Key)
                        {
                            case OperationType.Post:
                            case OperationType.Put:
                            case OperationType.Patch:
                                opPair.Value.RequestBody = requestBody;
                                break;
                        }
```

src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
```diff
@@ -53,5 +54,57 @@ public static OpenApiPathItem LoadPathItem(ParseNode node)
...
+            node.Context.SetTempStorage(TempStorageKeys.FormParameters, null);
+
+            pathItem.Parameters = node.CreateList(LoadParameter);
+
+            // Build request body based on information determined while parsing OpenApiOperation
+            var bodyParameter = node.Context.GetFromTempStorage<OpenApiParameter>(TempStorageKeys.BodyParameter);
+            if (bodyParameter != null)
+            {
+                var requestBody = CreateRequestBody(node.Context, bodyParameter);
+                foreach(var opPair in pathItem.Operations)
```
baywet, line 70:
```suggestion
                foreach(var opPair in pathItem.Operations.Where(x => x.Value.RequestBody is null))
```

src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
```diff
@@ -53,5 +54,57 @@ public static OpenApiPathItem LoadPathItem(ParseNode node)
...
+                    {
+                        switch (opPair.Key)
+                        {
+                            case OperationType.Post:
+                            case OperationType.Put:
+                            case OperationType.Patch:
+                                opPair.Value.RequestBody = requestBody;
+                                break;
+                        }
+                    }
```
baywet, line 82:
```suggestion
                    switch (opPair.Key)
                    {
                        case OperationType.Post:
                        case OperationType.Put:
                        case OperationType.Patch:
                            opPair.Value.RequestBody = requestBody;
                            break;
                    }
```

baywet (changes requested):
+ changelog entry in the CSproj

src/Microsoft.OpenApi.Readers/V2/OpenApiPathItemDeserializer.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
+using System.Collections.Generic;
```
baywet, line 4:
```suggestion
using System.Collections.Generic;
using System.Linq;
```

darrelmiller:
> * changelog entry in the CSproj

I would rather use the release notes in the CSProj to point to the release notes on GitHub than worry about updating this file with a description of every change.

baywet:
> > * changelog entry in the CSproj
> 
> I would rather use the release notes in the CSProj to point to the release notes on GitHub than worry about updating this file with a description of every change.

Then may I suggest to log a new issue and have somebody on the team take care of that?

baywet (approved):

MaggieKimani1 (approved):

darrelmiller:
> > > * changelog entry in the CSproj
> > 
> > 
> > I would rather use the release notes in the CSProj to point to the release notes on GitHub than worry about updating this file with a description of every change.
> 
> Then may I suggest to log a new issue and have somebody on the team take care of that?

Done https://github.com/microsoft/OpenAPI.NET/issues/923

PR #933, opened 2022-07-07
Fixes extensionless output

### LINKED ISSUE #877: hidi transform creates an extensionless `output` file
## Steps to reproduce

`hidi transform --inlineLocal -d https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/resources/resource-manager/Microsoft.Resources/stable/2021-04-01/resources.json`

## Expected result

Have a default file with the extension associated with the format it contains (`output.json` or `output.yaml`)

## Actual result

File called `output` without any extension.

image

MaggieKimani1:
At the moment, if a user fails to provide an output path while using ` hidi transform`, we default to `./output`.
This PR proposes:
  - Appending the Input path's extension to the resulting output path, if one is not provided i.e `./output.json`.
  - Moving Hidi tests to its own test project as a cleaner approach

Fixes #877


Microsoft.OpenApi.Hidi.Tests/Microsoft.OpenApi.Hidi.Tests.csproj
```diff
@@ -0,0 +1,54 @@
...
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
```
baywet, line 12:
```suggestion
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
```

(and check other dependencies, I believe they are outdated as well)

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -575,6 +574,21 @@ private static OpenApiFormat GetOpenApiFormat(string input, ILogger logger)
...
         }
 
+        private static string GetInputPathExtension(string openapi = null, string csdl = null)
+        {
+            var extension = String.Empty;
+            if (!string.IsNullOrEmpty(openapi))
+            {
+                extension = Path.GetExtension(openapi);
+            }
+            if (!string.IsNullOrEmpty(csdl))
```
baywet, line 584: else if? or check if extension is still null or empty?

baywet (changes requested):

baywet (approved):

PR #985, opened 2022-08-17
Compute hash of OpenApi document

### LINKED ISSUE #799: Create hash of OpenAPI Document
In order to prevent tooling like Kiota from doing redundant work, it would be valuable if we can generate a hash of a OpenAPIDocument instance and its member properties.

MaggieKimani1:
Fixes #799

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiDiagnostic.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiDiagnostic.cs
@@ -30,6 +30,6 @@ namespace Microsoft.OpenApi.Readers
         /// <summary>
         /// The unique hash code of the generated OpenAPI document
         /// </summary>
-        public int HashCode { get; set; }
+        public string HashCode { get; set; }
     }
 }
--- a/src/Microsoft.OpenApi.Readers/OpenApiDiagnostic.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiDiagnostic.cs
@@ -26,5 +26,10 @@ namespace Microsoft.OpenApi.Readers
         /// Open API specification version of the document parsed.
         /// </summary>
         public OpenApiSpecVersion SpecificationVersion { get; set; }
+
+        /// <summary>
+        /// The unique hash code of the generated OpenAPI document
+        /// </summary>
+        public int HashCode { get; set; }
     }
 }
--- a/src/Microsoft.OpenApi.Readers/OpenApiDiagnostic.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiDiagnostic.cs
@@ -26,10 +26,5 @@ namespace Microsoft.OpenApi.Readers
         /// Open API specification version of the document parsed.
         /// </summary>
         public OpenApiSpecVersion SpecificationVersion { get; set; }
-
-        /// <summary>
-        /// The unique hash code of the generated OpenAPI document
-        /// </summary>
-        public string HashCode { get; set; }
     }
 }
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -45,8 +45,6 @@ namespace Microsoft.OpenApi.Readers
             var reader = new StreamReader(input);
             var result = new OpenApiTextReaderReader(_settings).Read(reader, out diagnostic);
 
-            diagnostic.HashCode = OpenApiDocument.GenerateHashValue(input);
-
             if (!_settings.LeaveStreamOpen)
             {
                 reader.Dispose();
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.IO;
+using System.Security.Cryptography;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Models;
@@ -42,6 +44,17 @@ namespace Microsoft.OpenApi.Readers
         {
             var reader = new StreamReader(input);
             var result = new OpenApiTextReaderReader(_settings).Read(reader, out diagnostic);
+
+            HashAlgorithm sha = SHA512.Create();
+            byte[] data = sha.ComputeHash(input);
+            StringBuilder sb = new StringBuilder();
+            for (int i = 0; i < data.Length; i++)
+            {
+                sb.Append(data[i].ToString("X2"));
+            }
+
+            diagnostic.HashCode = sb.ToString();
+
             if (!_settings.LeaveStreamOpen)
             {
                 reader.Dispose();
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -3,8 +3,6 @@
 
 using System;
 using System.IO;
-using System.Security.Cryptography;
-using System.Text;
 using System.Threading.Tasks;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Models;
@@ -44,7 +42,6 @@ namespace Microsoft.OpenApi.Readers
         {
             var reader = new StreamReader(input);
             var result = new OpenApiTextReaderReader(_settings).Read(reader, out diagnostic);
-
             if (!_settings.LeaveStreamOpen)
             {
                 reader.Dispose();
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -45,15 +45,17 @@ namespace Microsoft.OpenApi.Readers
             var reader = new StreamReader(input);
             var result = new OpenApiTextReaderReader(_settings).Read(reader, out diagnostic);
 
-            HashAlgorithm sha = SHA512.Create();
-            byte[] data = sha.ComputeHash(input);
-            StringBuilder sb = new StringBuilder();
-            for (int i = 0; i < data.Length; i++)
-            {
-                sb.Append(data[i].ToString("X2"));
-            }
+            //HashAlgorithm sha = SHA512.Create();
+            //byte[] data = sha.ComputeHash(input);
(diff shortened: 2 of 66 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -375,6 +375,27 @@ public IOpenApiReferenceable ResolveReference(OpenApiReference reference)
...
 
+        /// <summary>
+        /// Computes the hash code for an OpenApiDocument and its property values.
+        /// </summary>
+        /// <returns> The hash code.</returns>
+        public override int GetHashCode()
+        {
+            // select two random prime numbers e.g 1 and 3 and use them to compute hash codes
+            int hash = 1;
+            hash = hash * 3 + (Workspace == null ? 0 : Workspace.GetHashCode());
```
baywet, line 386: I think you're supposed to select different prime numbers.

baywet (changes requested):
I don't think we should be using the GetHashCode override for multiple reasons:
- it's not its intended use, especially if caller want to store the hash somewhere for later comparison
- it can yield equality between unequal objects
- it will vary depending on chip instruction set (x86, 64, arm, arm64)
- it requires also overriding the Equals method

I'd instead recommend using a proper hashing algorithm (configurable?) like sha256 or up.

see the documentation of get hash code.
https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-6.0#remarks

MaggieKimani1:
> I don't think we should be using the GetHashCode override for multiple reasons:
> 
> * it's not its intended use, especially if caller want to store the hash somewhere for later comparison
> * it can yield equality between unequal objects
> * it will vary depending on chip instruction set (x86, 64, arm, arm64)
> * it requires also overriding the Equals method
> 
> I'd instead recommend using a proper hashing algorithm (configurable?) like sha256 or up.
> 
> see the documentation of get hash code. https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-6.0#remarks

I leaned more towards overriding `GetHashCode()` and tried getting the hash value for each of the property values to avoid collision as I believed the goal was to perform a simple object lookup as opposed to encryption which comes with its own set of complications:
- You have to convert the OpenApi document object into a byte array for hash computation
- We have to mark all object model types, `OpenApiAny` and expression types as `Serializable` for the hashing to take place..

baywet:
> le object lookup as opposed to encryption which comes with its own set of complications:
> 
> * You have to convert the OpenApi document object into a byte array for hash computation
> * We have to mark all object model types, `OpenApiAny` and `Expression` types as `Serializable` for the hashing to take place..

Nitpicking: hashing and encryption are two completely different things :) I'm not suggesting to rely on any encryption algorithm, simply on standard hashing. 

The nice additional thing with using industry standard hashing is that if we document the process correctly, other libs (from other languages) could interop with it and generate the exact same hash for the same document.

Serialize and convert: well that's the beauty of it, we already have the infrastructure to serialize the document to JSON/YAML, let's use that to produce a terse representation (so white-spacing doesn't impact the result), convert to binary, hash, and voilà! You don't even need to implement changes on the whole object model anymore, just on the document or root object.

What do you think? Also @darrelmiller, any objection with this approach?

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -376,25 +379,25 @@ public IOpenApiReferenceable ResolveReference(OpenApiReference reference)
...
 
         /// <summary>
-        /// Computes the hash code for an OpenApiDocument and its property values.
+        /// Uses the stream input to generate the hash value of an OpenApi document 
         /// </summary>
-        /// <returns> The hash code.</returns>
-        public override int GetHashCode()
+        /// <param name="input">Stream containing OpenAPI description to hash.</param>
+        /// <returns>The hash value.</returns>
+        public static string GenerateHashValue(Stream input)
```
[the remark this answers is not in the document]
MaggieKimani1: Actually here I'll need to expose this by moving the implementation from the `OpenApi.Models` project to the `OpenApi.Readers` project to make it `internal`

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -42,6 +44,9 @@ public OpenApiDocument Read(Stream input, out OpenApiDiagnostic diagnostic)
         {
             var reader = new StreamReader(input);
             var result = new OpenApiTextReaderReader(_settings).Read(reader, out diagnostic);
+
+            diagnostic.HashCode = OpenApiDocument.GenerateHashValue(input);
+
```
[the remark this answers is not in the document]
baywet: Actually why is this property in diagnostics at the first place?  Moving it to the document would make more sense IMHO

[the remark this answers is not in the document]
MaggieKimani1: The initial idea was to store the hash value here to keep track of it after reading in the document before serialization but I can move it to the OpenApi document object model

src/Microsoft.OpenApi.Readers/OpenApiDiagnostic.cs
```diff
@@ -26,5 +26,10 @@ public class OpenApiDiagnostic : IDiagnostic
         /// Open API specification version of the document parsed.
         /// </summary>
         public OpenApiSpecVersion SpecificationVersion { get; set; }
+
+        /// <summary>
+        /// The unique hash code of the generated OpenAPI document
+        /// </summary>
+        public string HashCode { get; set; }
```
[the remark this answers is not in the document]
MaggieKimani1: The hashing method takes in an input stream as a parameter hence why I called it in the stream reader class which begs the question on how I'll access the input stream to compute the hash if I call it here.

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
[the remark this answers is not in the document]
baywet: I think the scenario we're trying to cover here is :
1. read the document
2. generate a hash
3. (store the hash somewhere)
4. read the document again
5. compare the hash with the previously stored one to know whether the document has changed since the last time

I don't think that involves the serialization at any point (beside leveraging it for hashing purpose). Does that make sense?

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -42,6 +44,7 @@ public OpenApiDocument Read(Stream input, out OpenApiDiagnostic diagnostic)
         {
             var reader = new StreamReader(input);
             var result = new OpenApiTextReaderReader(_settings).Read(reader, out diagnostic);
+
```
baywet, line 47:
```suggestion
```

src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
```diff
@@ -3,6 +3,8 @@
 
 using System;
 using System.IO;
+using System.Security.Cryptography;
+using System.Text;
```
baywet, line 7:
```suggestion
```

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -376,25 +379,25 @@ public IOpenApiReferenceable ResolveReference(OpenApiReference reference)
...
-            hash = hash * 3 + (Paths == null ? 0 : Paths.GetHashCode());
-            hash = hash * 3 + (Components == null ? 0 : Components.GetHashCode());
-            hash = hash * 3 + (SecurityRequirements == null ? 0 : SecurityRequirements.GetHashCode());
-            hash = hash * 3 + (Tags == null ? 0 : Tags.GetHashCode());
-            hash = hash * 3 + (ExternalDocs == null ? 0 : ExternalDocs.GetHashCode());
-            hash = hash * 3 + (Extensions == null ? 0 : Extensions.GetHashCode());
-
-            return hash;
-        }        
+            HashAlgorithm sha = SHA512.Create();
```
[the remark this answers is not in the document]
baywet: the using is still missing

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -375,6 +383,41 @@ public IOpenApiReferenceable ResolveReference(OpenApiReference reference)
...
 
+        /// <summary>
+        /// Takes in an OpenApi document instance and generates its hash value 
+        /// </summary>
+        /// <param name="doc">The OpenAPI description to hash.</param>
+        /// <returns>The hash value.</returns>
+        public static string GenerateHashValue(OpenApiDocument doc)
+        {
+            HashAlgorithm sha = SHA512.Create();
+            using var memoryStream = new MemoryStream();
```
baywet, line 394:
```suggestion
```

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -375,6 +383,41 @@ public IOpenApiReferenceable ResolveReference(OpenApiReference reference)
...
+
+            using var cryptoStream = new CryptoStream(memoryStream, sha, CryptoStreamMode.Write);
```
baywet, line 396:
```suggestion
            using var cryptoStream = new CryptoStream(Stream.Null, sha, CryptoStreamMode.Write);
```

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -375,6 +383,41 @@ public IOpenApiReferenceable ResolveReference(OpenApiReference reference)
...
+            using var streamWriter = new StreamWriter(cryptoStream);
+
+            var openApiJsonWriter = new OpenApiJsonWriter(streamWriter, new OpenApiJsonWriterSettings { Terse = true });
+            doc.SerializeAsV3(openApiJsonWriter);
+            openApiJsonWriter.Flush();
+
+            var hash = memoryStream.ToArray();
```
baywet, line 403:
```suggestion
            cryptoStream.FlushFinalBlock();
            var hash = sha.Hash;
```

according to https://stackoverflow.com/questions/4725507/calculate-hash-when-writing-to-stream

baywet (changes requested):
thanks for making the changes, here are a few final remarks

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
MaggieKimani1: Noted.

baywet (approved):

PR #987, opened 2022-08-24
Cater for file content transferred with base64 encoding

### LINKED ISSUE #747: multipart/form-data in V2 is not properly handled
Hi, Team.

I found that there had been an issue #509 and fix #511 with regards to the file upload through the `multipart/form-data` content type. However, when I checked the source code, it only partially covers the cases.

1. According to the [OpenAPI v3.0.1 spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#considerations-for-file-uploads), file input/output can be the schema type/format of `string/binary` or `string/base64`. However, the codebase only looks after the `string/binary` pair.

    https://github.com/microsoft/OpenAPI.NET/blob/a636ec2c97ee62ddd42bcd74ffca1dfc7141a256/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L233-L236

2. According to the [3.0.1 spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#media-type-object), it's valid to have schema reference (the `$ref` value) under the `requestBody/content/<media-type>/schema` object instead of the `properties` object. However, the current codebase only looks after the `properties` object.

    https://github.com/microsoft/OpenAPI.NET/blob/a636ec2c97ee62ddd42bcd74ffca1dfc7141a256/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L229

Therefore, these two issues need to be resolved for proper V3 ➡️ V2 conversion. The [Azure Functions OpenAPI extension](https://github.com/Azure/azure-functions-openapi-extension) has addressed [this issue](https://github.com/Azure/azure-functions-openapi-extension/issues/365), which relies on this OpenAPI.NET package.

MaggieKimani1:
Fixes #747

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiOperation.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiOperation.cs
@@ -255,7 +255,7 @@ namespace Microsoft.OpenApi.Models
                     {
                         var paramName = property.Key;
                         var paramSchema = property.Value;
-                        if (paramSchema.Type == "string" && paramSchema.Format == "binary" || paramSchema.Format == "base64") {
+                        if (paramSchema.Type == "string" && (paramSchema.Format == "binary" || paramSchema.Format == "base64")) {
                             paramSchema.Type = "file";
                             paramSchema.Format = null;
                         }
--- a/src/Microsoft.OpenApi/Models/OpenApiOperation.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiOperation.cs
@@ -253,9 +253,11 @@ namespace Microsoft.OpenApi.Models
                 {
                     foreach (var property in RequestBody.Content.First().Value.Schema.Properties)
                     {
-                        var paramName = property.Key;
                         var paramSchema = property.Value;
-                        if (paramSchema.Type == "string" && (paramSchema.Format == "binary" || paramSchema.Format == "base64")) {
+                        if ("string".Equals(paramSchema.Type, StringComparison.OrdinalIgnoreCase) 
+                            && ("binary".Equals(paramSchema.Format, StringComparison.OrdinalIgnoreCase) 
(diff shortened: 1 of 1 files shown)


irvinesunday (dismissed):

src/Microsoft.OpenApi/Models/OpenApiOperation.cs
```diff
@@ -255,7 +255,7 @@ public void SerializeAsV2(IOpenApiWriter writer)
                     {
                         var paramName = property.Key;
                         var paramSchema = property.Value;
-                        if (paramSchema.Type == "string" && paramSchema.Format == "binary") {
+                        if (paramSchema.Type == "string" && paramSchema.Format == "binary" || paramSchema.Format == "base64") {
```
rkodev, line 258:
```c#
if (paramSchema.Type == "string" && paramSchema.Format == "binary" || paramSchema.Format == "base64")
```

perhaps you meant 

```c#
if (paramSchema.Type == "string" && (paramSchema.Format == "binary" || paramSchema.Format == "base64"))
```

or 

```c#
if ((paramSchema.Type == "string" && paramSchema.Format == "binary") || paramSchema.Format == "base64")
```

rkodev: var x = (false && (false || true)); // evaluates to false 
var x = (false && false || true); // evaluates to true

Adding the brackets makes it simpler to decipher the intention

MaggieKimani1: Yes I stand corrected.

baywet: I also made a suggestion regarding the string comparisons themselves. The suggestion I made doesn't apply to the boolean algebra, and the expectation is that you'll combine both suggestions.

baywet (approved):

PR #993, opened 2022-09-01
Retrieve Format property when serializing as V2

irvinesunday:
Patch fix to @MaggieKimani1 branch

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -630,6 +630,10 @@ internal void WriteAsItemsProperties(IOpenApiWriter writer)
             }
 
             // format
+            Format ??= AllOf?.FirstOrDefault(x => x.Format != null)?.Format ??
+                    AnyOf?.FirstOrDefault(x => x.Format != null)?.Format ??
+                    OneOf?.FirstOrDefault(x => x.Format != null)?.Format;
```
baywet, line 635:
```suggestion
            Format ??= AllOf?.FirstOrDefault(static x => x.Format != null)?.Format ??
                    AnyOf?.FirstOrDefault(static x => x.Format != null)?.Format ??
                    OneOf?.FirstOrDefault(static x => x.Format != null)?.Format;
```

baywet (changes requested):

baywet:
Can we get a preview release for this one soon so we can test the other fix that depends on it in the conversion library please?

MaggieKimani1:
This was actually supposed to go out with v1.4.0 which I'm to release tomorrow.

baywet:
We can also wait until tomorrow ;-)

Issue #992, opened 2022-08-31
Append `Format` property value to root schema when serializing as v2 if contained in `anyOf`, `allOf`, `oneOf` of child schemas


irvinesunday:
When a primitive type schema of the example:

```
var schema = new OpenApiSchema();
if (context.Settings.IEEE754Compatible)
{
    schema.OneOf = new List<OpenApiSchema>
    {
        new OpenApiSchema { Type = "number", Format = "decimal" },
        new OpenApiSchema { Type = "string" },
    };
}
```
The **expected** serialization for **v2** should look like below:

```
Age:
  format: int64
```
and the **expected** serialization for **v3**:

```
Age:
  oneOf:
    - type: integer
       format: int64
    - type: string
  nullable: true
```

This should be applicable to all primitive type schemas.

baywet:
Thanks for creating the issue here.
Before we proceed with a fix, I'd like to make sure we're on the same page on what to implement first.
I do think the backward compatible aspects with V2 should be the sole responsibility of this library and it should not push decisions on dependents for that aspect.
In different terms, I don't think that the OpenAPI.net.OData lib should have to set the format on the root schema "just in case we end up serializing to v2". But rather set things as they are expected to be in v3, and have this library handle the compatibility aspects.
When this library needs to serialize as v2, it should see a schema is a composition (one/any of) and take the first type and format it can find to use from them.
Does that make sense?

MaggieKimani1:
When reproducing the above issue using this schema: 
`Age :

    oneOf :
       - type : number
       - format : decimal
    type : string`

MaggieKimani1:
When reproducing the above issue using the following schema:
```
Age :
  oneOf :
    - type: number
      format: decimal
    - type: string
```
I've noticed that during V2 serialization, any `OneOf/AnyOf` schema objects are written out as an `allOf` with only the first schema in the list as an attempt to downgrade gracefully and the rest are completely ignored.
So the above schema would result to:
```
Age :
  allOf :
    - format : decimal
      type : number

```
with the `type : string` schema stripped out. Is this the intended behaviour @darrelmiller, @baywet and if so, why is that so?

baywet:
I believe the first example in Irvine's initial post doesn't match the following ones and should instead have been
```CSharp
if (context.Settings.IEEE754Compatible)
                    {
                        schema.OneOf = new List<OpenApiSchema>
                        {
                            new OpenApiSchema { Type = "integer", Format = "int64" },
[...]

PR #1001, opened 2022-09-07
Appends `format` property value to root schema if contained in `anyOf`, `oneOf` or `allOf` when serializing as v2

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET/issues/992

This PR:
- Retrieves the `format` property value from the child schemas of `anyOf`, `oneOf` or `allOf` and appends this to the root schema when serializing as v2 (when writing items as properties). 
- Updates test to validate the above.

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -566,10 +566,13 @@ namespace Microsoft.OpenApi.Models
             writer.WriteProperty(OpenApiConstants.Type, Type);
 
             // format
-            Format ??= AllOf?.FirstOrDefault(static x => x.Format != null)?.Format ??
+            if (string.IsNullOrEmpty(Format))
+            {
+                Format = AllOf?.FirstOrDefault(static x => x.Format != null)?.Format ??
                     AnyOf?.FirstOrDefault(static x => x.Format != null)?.Format ??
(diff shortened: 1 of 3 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -566,6 +566,10 @@ internal void WriteAsItemsProperties(IOpenApiWriter writer)
             writer.WriteProperty(OpenApiConstants.Type, Type);
 
             // format
+            Format ??= AllOf?.FirstOrDefault(static x => x.Format != null)?.Format ??
+                    AnyOf?.FirstOrDefault(static x => x.Format != null)?.Format ??
+                    OneOf?.FirstOrDefault(static x => x.Format != null)?.Format;
```
baywet, line 571: I'm not super familiar with the code base here. But shouldn't this have a condition to only do it when we are serializing to v2?

Also, the conditions should be string.IsNullOrEmpty, not just a null comparison.

baywet (changes requested):

baywet:
also we should probably bump the preview version to we can use it in the conversion lib PR

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
irvinesunday: This method is only called by methods that serialize as v2

baywet (approved):

Issue #1007, opened 2022-09-09
Paths are not copied over in the copy constructor of `OpenApiPaths`


irvinesunday:
To repro:

In the `OpenApiDocumentTest.cs`, add the below test:
```
[Fact]
private void ShouldCopyDocument()
{
    var doc = new OpenApiDocument(AdvancedDocument);
    Assert.NotEmpty(doc.Paths);
    Assert.Equal(2, doc.Paths.Count);
}
```
Note that the above test fails on both assertions.

PR #1008, opened 2022-09-12
Ensures `Paths` are copied in the copy constructor

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET/issues/1007

This PR:
- Ensures `Paths` are copied in the copy constructor of `OpenApiDocument`.
- Adds test to validate the above.

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
@@ -83,7 +83,7 @@ namespace Microsoft.OpenApi.Models
             Workspace = document?.Workspace != null ? new(document?.Workspace) : null;
             Info = document?.Info != null ? new(document?.Info) : null;
             Servers = document?.Servers != null ? new List<OpenApiServer>(document.Servers) : null;
-            Paths = document?.Paths;
+            Paths = document?.Paths != null ? new(document?.Paths) : null;
             Components = document?.Components != null ? new(document?.Components) : null;
             SecurityRequirements = document?.SecurityRequirements != null ? new List<OpenApiSecurityRequirement>(document.SecurityRequirements) : null;
             Tags = document?.Tags != null ? new List<OpenApiTag>(document.Tags) : null;
--- a/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
@@ -83,7 +83,7 @@ namespace Microsoft.OpenApi.Models
             Workspace = document?.Workspace != null ? new(document?.Workspace) : null;
             Info = document?.Info != null ? new(document?.Info) : null;
             Servers = document?.Servers != null ? new List<OpenApiServer>(document.Servers) : null;
-            if (document.Paths != null)
+            if (document?.Paths != null)
             {
                 Paths = new();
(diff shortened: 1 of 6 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -83,7 +83,11 @@ public OpenApiDocument(OpenApiDocument document)
             Workspace = document?.Workspace != null ? new(document?.Workspace) : null;
             Info = document?.Info != null ? new(document?.Info) : null;
             Servers = document?.Servers != null ? new List<OpenApiServer>(document.Servers) : null;
-            Paths = document?.Paths != null ? new(document?.Paths) : null;
+            if (document?.Paths != null)
+            {
+                Paths = new();
```
baywet, line 88:
```suggestion
```

baywet (changes requested):

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
irvinesunday: Thanks. I've simplified it to: `Paths = document?.Paths`

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -83,7 +83,7 @@ public OpenApiDocument(OpenApiDocument document)
             Workspace = document?.Workspace != null ? new(document?.Workspace) : null;
             Info = document?.Info != null ? new(document?.Info) : null;
             Servers = document?.Servers != null ? new List<OpenApiServer>(document.Servers) : null;
-            Paths = document?.Paths != null ? new(document?.Paths) : null;
+            Paths = document?.Paths;
```
baywet, line 86: I'm not sure why we're changing what we originally had. This will copy the reference which could have negative side effect (i.e. changing the dictionary content in the copy will impact the original).
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.-ctor?view=net-6.0#system-collections-generic-dictionary-2-ctor(system-collections-generic-idictionary((-0-1)))

```suggestion
            Paths = document?.Paths != null ? new(document.Paths) : null;
```

baywet (changes requested):

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
irvinesunday: We don't assign that parameter value to anything in the `OpenApiPaths` ctor: https://github.com/microsoft/OpenAPI.NET/blob/0b710d553eb5fe4bfceb0b1452923c09c6b2d098/src/Microsoft.OpenApi/Models/OpenApiPaths.cs#L19
The new document will always have its `Paths` empty

baywet: ah I see, and that masks the copy constructor from dictionary. We should probably call the base constructor.
`public OpenApiPaths(OpenApiPaths paths): base(paths) {}`

And in `OpenApiExtensibleDictionary` have a copy constructor that calls the base, but also assigns the extensions.

baywet (dismissed):

baywet (approved):

PR #1014, opened 2022-09-19
fix: invariant culture for version field

Ne4to:
Writing version field depends on current system culture 
`version: '1.13.0'` for `en-US`
`version: 1.13.0` for `en-GB`

existing workaround:
set `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` environment variable 
https://learn.microsoft.com/en-us/dotnet/core/runtime-config/globalization#invariant-mode


Ne4to:
It looks like one of the checks is not able to report status. How can I force to restart the check?

baywet (approved):
@Ne4to Thanks for the contribution.
This check actually won't work for you, compliance policies prevent us from running some of the checks on PRs from forks, let me see how I can unblock this.

PR #1032, opened 2022-10-06
Bump version for preview release

MaggieKimani1:


baywet (approved):
just checking that readers doesn't need a bump as well?

MaggieKimani1:
> just checking that readers doesn't need a bump as well?

Oh, actually I think it does. Creating a separate PR for this.

PR #1033, opened 2022-10-06
Fix $ref serialization errors for requestBodies and responses

millicentachieng:


src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
```diff
@@ -144,5 +146,47 @@ public void SerializeAsV2WithoutReference(IOpenApiWriter writer)
...
+        {
+            var bodyParameter = new OpenApiBodyParameter
+            {
+                Description = Description,
+                // V2 spec actually allows the body to have custom name.
+                // To allow round-tripping we use an extension to hold the name
+                Name = "body",
+                Schema = Content.Values.FirstOrDefault()?.Schema ?? new OpenApiSchema(),
+                Required = Required,
+                Extensions = Extensions.ToDictionary(k => k.Key, v => v.Value)  // Clone extensions so we can remove the x-bodyName extensions from the output V2 model.
```
baywet, line 160:
```suggestion
                Extensions = Extensions.ToDictionary(static k => k.Key, static v => v.Value)  // Clone extensions so we can remove the x-bodyName extensions from the output V2 model.
```

baywet (approved):

PR #1046, opened 2022-10-24
Add webhooks property to the root document

MaggieKimani1:
This PR:
- Adds a root webhooks property to the OpenAPI document object model
- Adds logic to serialize and deserialize the webhooks object
- Adds tests


src/Microsoft.OpenApi/Validations/Rules/OpenApiDocumentRules.cs
```diff
@@ -28,15 +28,6 @@ public static class OpenApiDocumentRules
                             String.Format(SRResource.Validation_FieldIsRequired, "info", "document"));
                     }
                     context.Exit();
-
```
baywet, line 31: can you elaborate a bit on this change please?

MaggieKimani1: According to the 3.1 spec, the OpenAPI document MUST at least contain one `paths`, `components` or `webhooks` field hence paths is [no longer a required field](https://spec.openapis.org/oas/v3.1.0#openapi-object). 

However, as I'm typing this I've just realized that removing this would affect version 2.0 and 3.0 as well. Let me add a guard clause that ensures the rule is upheld for 2.0 and 3.0 but ignored for 3.1.

baywet (approved):

src/Microsoft.OpenApi.Readers/V3/OpenApiDocumentDeserializer.cs
```diff
@@ -48,12 +48,24 @@ internal static partial class OpenApiV3Deserializer
...
+            ValidatePathsField(openApidoc, rootNode);
 
             return openApidoc;
         }
+        
+        private static void ValidatePathsField(OpenApiDocument doc, RootNode rootNode)
+        {
+            var versionNode = rootNode.Find(new JsonPointer("/openapi")).GetScalarValue();
+            if (versionNode == null) return;
+            else if (versionNode.Contains("3.0") && doc.Paths == null)
```
baywet, line 64: couldn't the path node be present but no paths defined in it? maybe we need to guard against that too?
what about version 2?
For v3.1, shouldn't we check that either webhooks or paths are present?

baywet (changes requested):

src/Microsoft.OpenApi.Readers/V3/OpenApiDocumentDeserializer.cs
MaggieKimani1: Yeah these are all valid concerns that need to be addressed. Thanks for pointing that out.

src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -244,5 +246,21 @@ public void PopLoop(string loopid)
             }
         }
 
+        private void ValidateRequiredFields(OpenApiDocument doc, string version)
+        {
+            if ((version == "2.0" || version.StartsWith("3.0")) && (doc.Paths == null || doc.Paths.Count == 0))
```
baywet, line 251:
```suggestion
            if ((version == "2.0" || version.StartsWith("3.0")) && (doc.Paths == null || !doc.Paths.Any()))
```

src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -244,5 +246,21 @@ public void PopLoop(string loopid)
...
+            {
+                // paths is a required field in OpenAPI 3.0 but optional in 3.1
+                RootNode.Context.Diagnostic.Errors.Add(new OpenApiError("", $"Paths is a REQUIRED field at {RootNode.Context.GetLocation()}"));
+            }
+            else if (version.StartsWith("3.1"))
+            {
+                if ((doc.Paths == null || doc.Paths.Count == 0) && (doc.Webhooks == null || doc.Webhooks.Count == 0))
```
baywet, line 258:
```suggestion
                if ((doc.Paths == null || !doc.Paths.Any()) && (doc.Webhooks == null || !doc.Webhooks.Any()))
```

baywet (changes requested):

baywet (approved):
Thanks for taking the suggestions

baywet (approved):

PR #1056, opened 2022-10-31
Add reusable path items to components object

MaggieKimani1:
This PR:
- Adds a `PathItems` property to the Components object that contains paths that can be reused in `Webhooks` by using a `$ref` pointer
- Adds logic to resolve the reference pointer to the reusable pathItem
- Adds a test to validate this


src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -244,5 +246,21 @@ public void PopLoop(string loopid)
...
+        private void ValidateRequiredFields(OpenApiDocument doc, string version)
+        {
+            if ((version == "2.0" || version.StartsWith("3.0")) && (doc.Paths == null || !doc.Paths.Any()))
+            {
+                // paths is a required field in OpenAPI 3.0 but optional in 3.1
+                RootNode.Context.Diagnostic.Errors.Add(new OpenApiError("", $"Paths is a REQUIRED field at {RootNode.Context.GetLocation()}"));
+            }
+            else if (version.StartsWith("3.1"))
+            {
+                if ((doc.Paths == null || !doc.Paths.Any()) && (doc.Webhooks == null || !doc.Webhooks.Any()))
```
baywet, line 258: collapse this if with the previous else if

baywet (changes requested):

src/Microsoft.OpenApi.Readers/ParsingContext.cs
```diff
@@ -63,12 +63,14 @@ internal OpenApiDocument Parse(YamlDocument yamlDocument)
                     VersionService = new OpenApiV2VersionService(Diagnostic);
                     doc = VersionService.LoadDocument(RootNode);
                     this.Diagnostic.SpecificationVersion = OpenApiSpecVersion.OpenApi2_0;
+                    ValidateRequiredFields(doc, version);
                     break;
 
-                case string version when version.StartsWith("3.0"):
+                case string version when version.StartsWith("3.0") || version.StartsWith("3.1"):
```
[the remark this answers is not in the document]
MaggieKimani1: Yeah I think that would help optimize and promote reusability in the string comparison cases.

src/Microsoft.OpenApi.Readers/OpenApiVersionExtensionMethods.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        /// <summary>
+        /// Extension method for Spec version 2.0
+        /// </summary>
+        /// <param name="version"></param>
+        /// <returns></returns>
+        public static bool is2_0(this string version)
+        {
+            bool result = false;
+            if (version.Equals("2.0"))
```
baywet, line 19: Ordinal ignore case string comparison ( same for other two methods)

baywet (changes requested):

baywet (approved):

PR #1064, opened 2022-11-07
Enable Hidi users configure CSDL to OpenAPI translation

MaggieKimani1:


src/Microsoft.OpenApi.Hidi/Program.cs
```diff
@@ -41,6 +47,9 @@ static async Task Main(string[] args)
             var terseOutputOption = new Option<bool>("--terse-output", "Produce terse json output");
             terseOutputOption.AddAlias("--to");
 
+            var settingsFileOption = new Option<string>("--settingsFile", "The configuration file with CSDL conversion settings.");
```
baywet, line 50:
```suggestion
            var settingsFileOption = new Option<string>("--settings-path", "The configuration file with CSDL conversion settings.");
```

src/Microsoft.OpenApi.Hidi/Program.cs
```diff
@@ -110,9 +121,20 @@ static async Task Main(string[] args)
...
+
+            //await new CommandLineBuilder(rootCommand)
+            //        .UseHost(_ => Host.CreateDefaultBuilder(), 
+            //        host => {
+            //            var config = host.Services.GetRequiredService<IConfiguration>();
+            //        })
+            //        .UseDefaults()
+            //        .Build()
+            //        .InvokeAsync(args);
+
```
baywet, line 137:
```suggestion
```

test/Microsoft.OpenApi.Hidi.Tests/Services/OpenApiServiceTests.cs
```diff
@@ -51,5 +53,19 @@ public async Task ReturnFilteredOpenApiDocBasedOnOperationIdsAndInputCsdlDocumen
             Assert.NotEmpty(subsetOpenApiDocument.Paths);
             Assert.Equal(expectedPathCount, subsetOpenApiDocument.Paths.Count);
         }
+        
+        [Fact]
+        public void ReturnOpenApiConvertSettings()
+        {
+            // Arrange
+            var filePath = "C:/Users/v-makim/source/repos/OpenAPI.NET/test/Microsoft.OpenApi.Hidi.Tests/UtilityFiles/appsettingstest.json";
```
baywet, line 61: hardcoded path, use a relative one instead

test/Microsoft.OpenApi.Hidi.Tests/Services/OpenApiServiceTests.cs
```diff
@@ -51,5 +53,19 @@ public async Task ReturnFilteredOpenApiDocBasedOnOperationIdsAndInputCsdlDocumen
             Assert.NotEmpty(subsetOpenApiDocument.Paths);
             Assert.Equal(expectedPathCount, subsetOpenApiDocument.Paths.Count);
         }
+        
+        [Fact]
+        public void ReturnOpenApiConvertSettings()
```
baywet, line 58: add a test for when the file is not found

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -304,37 +307,52 @@ private static Stream ApplyFilter(string csdl, string entitySetOrSingleton, XslC
             }
         }
 
+        public static IConfiguration GetConfiguration(string settingsFile)
```
baywet, line 310: switch to private or internal and use friendly assemblies for unit tests. We don't want people to take a dependency on that.

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -304,37 +307,52 @@ private static Stream ApplyFilter(string csdl, string entitySetOrSingleton, XslC
...
         /// <returns>An OpenAPI document.</returns>
-        public static async Task<OpenApiDocument> ConvertCsdlToOpenApi(Stream csdl)
+        public static async Task<OpenApiDocument> ConvertCsdlToOpenApi(Stream csdl, string settingsFile = null)
         {
             using var reader = new StreamReader(csdl);
             var csdlText = await reader.ReadToEndAsync();
             var edmModel = CsdlReader.Parse(XElement.Parse(csdlText).CreateReader());
+            
+            var config = GetConfiguration(settingsFile);
+            var settings = config.GetSection("OpenApiConvertSettings").Get<OpenApiConvertSettings>();
```
baywet, line 333: why not just binding to avoid multiple defaults? (i.e. the defaults from the lib, the defaults in the configuration file, the defaults below)

```CSharp

var settings = new OpenApiConvertSettings{ // all the default values)
config.GetSecion("OpenApiConvertSettings).Bind(settings);
```

baywet (changes requested):

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
MaggieKimani1: Does binding the default settings allow one to customize them by providing their own settings file?

baywet (approved):
thanks for taking in the suggestions.

PR #1072, opened 2022-11-08
- upgrades to dotnet 7

baywet:

### DIFF
--- a/.azure-pipelines/ci-build.yml
+++ b/.azure-pipelines/ci-build.yml
@@ -31,9 +31,9 @@ stages:
     - job: build
       steps:
       - task: UseDotNet@2
-        displayName: 'Use .NET 6'
+        displayName: 'Use .NET'
         inputs:
-          version: 6.x
+          version: 7.x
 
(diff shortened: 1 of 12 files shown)


MaggieKimani1:
@baywet I'm curious, was ESRP not working with dotnet 6?

MaggieKimani1:
> Yeah that's what I'm looking into, I think what's actually happening is that net6 "allowed" some older versions, and net7 disables them (even though supported by net6) unless explicitly installed. ESRP is on 2.1....

But I see we're using v1.0 of the ESRP code signing task here https://github.com/microsoft/OpenAPI.NET/blob/6c994a589fdda06b2c86514709fbb2fdde4fd9d0/.azure-pipelines/ci-build.yml#L194

baywet:
yeah I'm not sure how the internals of that task work. Effectively what's happening at this point is the task doesn't even start, so I don't believe it has the time to parse the configuration.

MaggieKimani1 (approved):

PR #1077, opened 2022-11-14
Add summary and description to OpenApiReference object

MaggieKimani1:
- Adds `summary` and `description` fields to the Reference object whose values by default SHOULD override those of the referenced component.
- Adds test to validate this.

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiV2VersionService.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiV2VersionService.cs
@@ -225,7 +225,7 @@ namespace Microsoft.OpenApi.Readers.V2
         /// <inheritdoc />
         public string GetReferenceScalarValues(MapNode mapNode, string scalarValue)
         {
-            throw new NotImplementedException();
+            throw new InvalidOperationException();
         }
     }
 }
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiV2VersionService.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiV2VersionService.cs
@@ -134,7 +134,7 @@ namespace Microsoft.OpenApi.Readers.V2
         /// <summary>
         /// Parse the string to a <see cref="OpenApiReference"/> object.
         /// </summary>
-        public OpenApiReference ConvertToOpenApiReference(string reference, ReferenceType? type)
+        public OpenApiReference ConvertToOpenApiReference(string reference, ReferenceType? type, string summary = null, string description = null)
         {
             if (!string.IsNullOrWhiteSpace(reference))
             {
@@ -220,6 +220,12 @@ namespace Microsoft.OpenApi.Readers.V2
         public T LoadElement<T>(ParseNode node) where T : IOpenApiElement
         {
             return (T)_loaders[typeof(T)](node);
+        }
+
+        /// <inheritdoc />
+        public string GetReferenceScalarValues(MapNode mapNode, string scalarValue)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/src/Microsoft.OpenApi.Readers/V3/OpenApiExampleDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V3/OpenApiExampleDeserializer.cs
@@ -52,17 +52,12 @@ namespace Microsoft.OpenApi.Readers.V3
         public static OpenApiExample LoadExample(ParseNode node)
         {
             var mapNode = node.CheckMapNode("example");
-            string description = null;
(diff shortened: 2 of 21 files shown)


src/Microsoft.OpenApi.Readers/V2/OpenApiV2VersionService.cs
```diff
@@ -221,5 +221,11 @@ public OpenApiDocument LoadDocument(RootNode rootNode)
         {
             return (T)_loaders[typeof(T)](node);
         }
+
+        /// <inheritdoc />
+        public string GetReferenceScalarValues(MapNode mapNode, string scalarValue)
+        {
+            throw new NotImplementedException();
```
baywet, line 228: not implemented or invalid operation?

src/Microsoft.OpenApi.Readers/V3/OpenApiV3VersionService.cs
```diff
@@ -151,7 +162,17 @@ public OpenApiDocument LoadDocument(RootNode rootNode)
             return (T)_loaders[typeof(T)](node);
         }
 
-        private OpenApiReference ParseLocalReference(string localReference)
+
+        /// <inheritdoc />
+        public string GetReferenceScalarValues(MapNode mapNode, string scalarValue)
+        {
+            var valueNode = mapNode.Where(x => x.Name.Equals(scalarValue))
+                .Select(x => x.Value as ValueNode).FirstOrDefault();
```
baywet, line 170:
```suggestion
                .Select(static x => x.Value).OfType<ValueNode>().FirstOrDefault();
```

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -504,31 +504,51 @@ internal IOpenApiReferenceable ResolveReference(OpenApiReference reference, bool
                 switch (reference.Type)
                 {
                     case ReferenceType.Schema:
-                        return this.Components.Schemas[reference.Id];
+                        var resolvedSchema = this.Components.Schemas[reference.Id];
+                        resolvedSchema.Description = reference.Description != null ? reference.Description : resolvedSchema.Description;
```
baywet, line 508: should it be string null or empty instead? or do we want to keep empty values?

src/Microsoft.OpenApi.Readers/V2/OpenApiV2VersionService.cs
MaggieKimani1: InvalidOperation..thanks for pointing this out

src/Microsoft.OpenApi.Readers/V3/OpenApiExampleDeserializer.cs
```diff
@@ -51,11 +52,19 @@ internal static partial class OpenApiV3Deserializer
...
         {
             var mapNode = node.CheckMapNode("example");
+            string description = null;
+            string summary = null;
 
             var pointer = mapNode.GetReferencePointer();
             if (pointer != null)
             {
-                return mapNode.GetReferencedObject<OpenApiExample>(ReferenceType.Example, pointer);
+                if (mapNode.Count() > 1)
```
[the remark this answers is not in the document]
MaggieKimani1: Because for a mapNode with a $ref pointer, it will always have a child property node(the $ref). The count() method is to check whether there are more property nodes besides the $ref i.e. looking for a summary and/or description node hence the greater than(<1) condition.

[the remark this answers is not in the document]
baywet: would it be better to use any with a filter clause for $ref instead? to magic numbers usually come back to haunt you at some point by experience :)

src/Microsoft.OpenApi.Readers/V3/OpenApiV3VersionService.cs
```diff
@@ -166,10 +166,17 @@ public OpenApiDocument LoadDocument(RootNode rootNode)
         /// <inheritdoc />
         public string GetReferenceScalarValues(MapNode mapNode, string scalarValue)
         {
-            var valueNode = mapNode.Where(x => x.Name.Equals(scalarValue))
+            var filteredList = mapNode.Where(x => x.Name != "$ref");
+
+            if (filteredList.Any())
```
baywet, line 171:
```suggestion
            if (mapNode.Any(static x => !"$ref".Equals(x.Name, StringComparison.OrdinalIgnoreCase)))
```

baywet (changes requested):
Thanks for making the changes, one last suggestion

baywet (approved):
Thanks for taking in all the changes!

PR #1080, opened 2022-11-17
Add files via upload

isekene:
added timespan type

### DIFF
--- a/IOpenApiPrimitive.cs
+++ b/IOpenApiPrimitive.cs
@@ -1,0 +1,82 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+namespace Microsoft.OpenApi.Any
+{
+    /// <summary>
+    /// Primitive type.
+    /// </summary>
+    public enum PrimitiveType
+    {
+        /// <summary>
+        /// Integer
+        /// </summary>
+        Integer,
+
+        /// <summary>
+        /// Long
+        /// </summary>
+        Long,
+
+        /// <summary>
+        /// Float
+        /// </summary>
+        Float,
+
+        /// <summary>
+        /// Double
+        /// </summary>
+        Double,
+
+        /// <summary>
+        /// String
+        /// </summary>
+        String,
+
+        /// <summary>
+        /// Byte
+        /// </summary>
+        Byte,
+
+        /// <summary>
+        /// Binary
+        /// </summary>
+        Binary,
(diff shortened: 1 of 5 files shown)


isekene:
> @microsoft-github-policy-service agree

baywet (changes requested):
@isekene thanks for the contribution. It seems that the files you added were not uploaded at the right place (under their original directory).
Could you move them to the right directory please?

darrelmiller:
There is no agreed upon format name for timespan https://spec.openapis.org/registry/format/index.html.  Without at least a registry entry, we should not add support for new formats

baywet:
@darrelmiller while this PR still needs some work (files at not at the right location for instance) before it can be merged, Kiota maps [string + duration to TimeSpan in dotnet](https://github.com/microsoft/kiota/blob/bac004b2c3688e6dde6ddf010ec0f8f272d7d080/tests/Kiota.Builder.Tests/KiotaBuilderTests.cs#L2724) (and equivalent types in other languages).
I'm happy to PR the registry with with additional type mappings we have in kiota and the ones that @MaggieKimani1 added in #1094

darrelmiller:
@baywet We need to decide why and when we are going to use these native types.  Currently we use them in the default property of OpenApiSchema, but in the future that is going to be out of our hands as it will be implemented by a JsonSchema library.  I don't think we should try an convert examples into native types during parsing.  Parsing the example into native types requires knowing the complete schema which may be referenced and may live in other files.  I think if users want to translate an OpenApiExample into a native object, then we should have separate code to do that.  I don't think we need the full set of primitive types.

baywet:
@darrelmiller the aspect that this PR tries to parse the examples was lost on me because of the formatting/files being at the wrong place.
Outside of examples, just to represent the schema, unless we decide to take a dependency on a JSON schema lib, which is not the case today, I'd argue that having types that represent known type/format combinations is actually a good thing.
People won't have to remember the exact names/combinations, they don't risk making typos, generators would be able to use type inference instead of string comparison, etc...
Thoughts?

darrelmiller:
@baywet  The problem we have is that even today, validation of examples will fail if a schema is a $ref.  That's one reason I changed the example validation to be a warning, not an error.  Also, we run into issues like this https://github.com/microsoft/OpenAPI.NET/issues/1106 where we try and represent a JSON number but there is no equivalent type in .NET.

baywet:
closing due to inactivity from author

PR #1094, opened 2022-12-01
Add support for OpenAPI schema generation for POCOs

### LINKED ISSUE #1087: Add support for OpenAPI schema generation for POCOs
This is related to https://github.com/microsoft/OpenAPI.NET/issues/836.

For certain native .NET types, it would be great if OpenAPI.NET could provide the authorative OpenApiSchema definitions for these types. Current prototypes maintain a dictionary like this:

```csharp
private static readonly Dictionary<Type, Func<OpenApiSchema>> _primitiveTypeToOpenApiSchema =
        new()
        {
            [typeof(bool)] = () => new OpenApiSchema { Type = "boolean" },
            [typeof(byte)] = () => new OpenApiSchema { Type = "string", Format = "byte" },
            [typeof(int)] = () => new OpenApiSchema { Type = "integer", Format = "int32" },
            ...
            [typeof(DateTime)] = () => new OpenApiSchema { Type = "string", Format = "date-time" },
            [typeof(DateTimeOffset)] = () => new OpenApiSchema { Type = "string", Format = "date-time" },
            [typeof(Guid)] = () => new OpenApiSchema { Type = "string", Format = "uuid" },
            [typeof(char)] = () => new OpenApiSchema { Type = "string" }
        };
```

And there a variety of open source packages/projects that maintain their own associations (see [this](https://github.com/ipjohnson/EasyRpc/blob/012cb2731b5c2ef36a909c1c2e73220a14c5fec5/src/EasyRpc.AspNetCore/Documentation/KnownOpenApiTypeMapper.cs), [this](https://github.com/vitalybibikov/AzureExtensions.Swashbuckle/blob/d4d33d526c9ce19cff2a522ad95f7d49d2eab8c9/src/AzureFunctions.Extensions.Swashbuckle/AzureFunctions.Extensions.Swashbuckle/SwashBuckle/Filters/Mapper/TypeMapper.cs), and [this](https://github.com/JamesRandall/FunctionMonkey/blob/45f4b7efc843f13a59899cdaaf470753d99bb212/Source/FunctionMonkey.Compiler.Core/Extensions/TypeExtensions.cs)).

Since this is a finite set of mappings, it would be great if they would be centralized in the OpenAPI.NET package to avoid duplication and ambiguity about the right schemas to generate for different native types in .NET.

cc: @darrelmiller  @baywet

MaggieKimani1:
Fixes #1087

### DIFF
--- a/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
+++ b/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
@@ -1,0 +1,85 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System;
+using System.Collections.Generic;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Extensions
+{
+    /// <summary>
+    /// Extension methods for <see cref="Type"/>.
+    /// </summary>
+    public static class OpenApiTypeMapper
+    {
+        private static readonly Dictionary<Type, Func<OpenApiSchema>> _simpleTypeToOpenApiSchema = new()
+        {
+            [typeof(bool)] = () => new OpenApiSchema { Type = "boolean" },
+            [typeof(byte)] = () => new OpenApiSchema { Type = "string", Format = "byte" },
+            [typeof(int)] = () => new OpenApiSchema { Type = "integer", Format = "int32" },
+            [typeof(uint)] = () => new OpenApiSchema { Type = "integer", Format = "int32" },
+            [typeof(long)] = () => new OpenApiSchema { Type = "integer", Format = "int64" },
+            [typeof(ulong)] = () => new OpenApiSchema { Type = "integer", Format = "int64" },
+            [typeof(float)] = () => new OpenApiSchema { Type = "number", Format = "float" },
+            [typeof(double)] = () => new OpenApiSchema { Type = "number", Format = "double" },
+            [typeof(decimal)] = () => new OpenApiSchema { Type = "number", Format = "double" },
+            [typeof(DateTime)] = () => new OpenApiSchema { Type = "string", Format = "date-time" },
+            [typeof(DateTimeOffset)] = () => new OpenApiSchema { Type = "string", Format = "date-time" },
+            [typeof(Guid)] = () => new OpenApiSchema { Type = "string", Format = "uuid" },
+            [typeof(char)] = () => new OpenApiSchema { Type = "string" },
+            
+            // Nullable types
+            [typeof(bool?)] = () => new OpenApiSchema { Type = "boolean", Nullable = true },
+            [typeof(byte?)] = () => new OpenApiSchema { Type = "string", Format = "byte", Nullable = true },
+            [typeof(int?)] = () => new OpenApiSchema { Type = "integer", Format = "int32", Nullable = true },
+            [typeof(uint?)] = () => new OpenApiSchema { Type = "integer", Format = "int32", Nullable = true },
+            [typeof(long?)] = () => new OpenApiSchema { Type = "integer", Format = "int64", Nullable = true },
+            [typeof(ulong?)] = () => new OpenApiSchema { Type = "integer", Format = "int64", Nullable = true },
+            [typeof(float?)] = () => new OpenApiSchema { Type = "number", Format = "float", Nullable = true },
+            [typeof(double?)] = () => new OpenApiSchema { Type = "number", Format = "double", Nullable = true },
+            [typeof(decimal?)] = () => new OpenApiSchema { Type = "number", Format = "double", Nullable = true },
+            [typeof(DateTime?)] = () => new OpenApiSchema { Type = "string", Format = "date-time", Nullable = true },
+            [typeof(DateTimeOffset?)] = () => new OpenApiSchema { Type = "string", Format = "date-time", Nullable = true },
+            [typeof(Guid?)] = () => new OpenApiSchema { Type = "string", Format = "uuid", Nullable = true },
+            [typeof(char?)] = () => new OpenApiSchema { Type = "string", Nullable = true },
+            
+            [typeof(Uri)] = () => new OpenApiSchema { Type = "string" }, // Uri is treated as simple string
+            [typeof(string)] = () => new OpenApiSchema { Type = "string" },
+            [typeof(object)] = () => new OpenApiSchema { Type = "object" }
+        };
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Extensions
+{
+    /// <summary>
+    /// Extension methods for <see cref="Type"/>.
+    /// </summary>
+    public static class OpenApiTypeMapper
+    {
+        private static readonly Dictionary<Type, Func<OpenApiSchema>> _simpleTypeToOpenApiSchema = new()
```
baywet, line 15: could we also add the opposite mapping? (from format/type to dotnet type) it'd make the life of generators much easier.
Here is a base for comparison.
https://github.com/microsoft/kiota/blob/8b211cd6a90d0211b5467728f7bb04a174bd9dee/tests/Kiota.Builder.Tests/KiotaBuilderTests.cs#L2714

baywet (changes requested):

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
MaggieKimani1: Yeah that makes sense.

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -0,0 +1,85 @@
...
+            [typeof(ulong?)] = () => new OpenApiSchema { Type = "integer", Format = "int64", Nullable = true },
+            [typeof(float?)] = () => new OpenApiSchema { Type = "number", Format = "float", Nullable = true },
+            [typeof(double?)] = () => new OpenApiSchema { Type = "number", Format = "double", Nullable = true },
+            [typeof(decimal?)] = () => new OpenApiSchema { Type = "number", Format = "double", Nullable = true },
+            [typeof(DateTime?)] = () => new OpenApiSchema { Type = "string", Format = "date-time", Nullable = true },
+            [typeof(DateTimeOffset?)] = () => new OpenApiSchema { Type = "string", Format = "date-time", Nullable = true },
+            [typeof(Guid?)] = () => new OpenApiSchema { Type = "string", Format = "uuid", Nullable = true },
+            [typeof(char?)] = () => new OpenApiSchema { Type = "string", Nullable = true },
+            
+            [typeof(Uri)] = () => new OpenApiSchema { Type = "string" }, // Uri is treated as simple string
```
captainsafia, line 46: Do we want to support `Uri?` too?

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -0,0 +1,85 @@
...
+            [typeof(int?)] = () => new OpenApiSchema { Type = "integer", Format = "int32", Nullable = true },
+            [typeof(uint?)] = () => new OpenApiSchema { Type = "integer", Format = "int32", Nullable = true },
+            [typeof(long?)] = () => new OpenApiSchema { Type = "integer", Format = "int64", Nullable = true },
+            [typeof(ulong?)] = () => new OpenApiSchema { Type = "integer", Format = "int64", Nullable = true },
+            [typeof(float?)] = () => new OpenApiSchema { Type = "number", Format = "float", Nullable = true },
+            [typeof(double?)] = () => new OpenApiSchema { Type = "number", Format = "double", Nullable = true },
+            [typeof(decimal?)] = () => new OpenApiSchema { Type = "number", Format = "double", Nullable = true },
+            [typeof(DateTime?)] = () => new OpenApiSchema { Type = "string", Format = "date-time", Nullable = true },
+            [typeof(DateTimeOffset?)] = () => new OpenApiSchema { Type = "string", Format = "date-time", Nullable = true },
+            [typeof(Guid?)] = () => new OpenApiSchema { Type = "string", Format = "uuid", Nullable = true },
```
captainsafia, line 43: Is `uuid` a valid format?

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
baywet: And if we do I think the format needs to be specified

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
baywet: Yep
https://spec.openapis.org/registry/format/ 
https://json-schema.org/draft/2020-12/json-schema-validation.html#name-vocabularies-for-semantic-c
https://json-schema.org/draft/2020-12/json-schema-validation.html#name-defined-formats

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
MaggieKimani1: The typeof operator cannot be used on `Uri?` since its a nullable reference type similar to `string?` and it does not have a standalone runtime representation https://github.com/dotnet/csharplang/discussions/3003?sort=old

baywet: I had misread Safia's original comment. This is what I meant with mine.
https://json-schema.org/draft/2020-12/json-schema-validation.html#name-resource-identifiers
```suggestion
            [typeof(Uri)] = () => new OpenApiSchema { Type = "string", Format = "uri" }, // Uri is treated as simple string
```

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// <returns>The simple type</returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        public static Type MapOpenApiPrimitiveTypeToSimpleType(this OpenApiSchema schema)
+        {
+            if (schema == null)
+            {
+                throw new ArgumentNullException(nameof(schema));
+            }
+
+            var type = (schema.Type, schema.Format, schema.Nullable) switch
```
baywet, line 98:
```suggestion
            var type = (schema.Type?.ToLowerInvariant(), schema.Format?.ToLowerInvariant(), schema.Nullable) switch
```

baywet (changes requested):

baywet (dismissed):

baywet (approved):

PR #1104, opened 2022-12-09
Fix check Extensions object for null to avoid null reference exception

quanterion:
using Microsoft.OpenApi.OData package it generates OpenApi document with OpenApiComponents.Extensions == null and usage of OpenApiWalker triggers an exception


baywet (approved):
Thank you a lot for the contribution!

PR #1113, opened 2022-12-23
Adds support for generating Mermaid diagrams from APIs

### LINKED ISSUE #820: Generate Mermaid diagram for OpenAPI
e.g. 
```mermaid
flowchart LR
    root["/"] --> Pets
    root --> Users["users"]
    Users --> User["{userId}"]
    User --> UserPets["pets"]
    root --> Store[""store]
    Pets --> Pet["{petId}"]
    Pet --> Owner[""owner] 
```

or 

```mermaid
flowchart LR
    root["/"] --  pets --> Pets
    root -- users --> Users["GET, POST"]
    Users -- "{userId}" --> User["GET,PATCH,DELETE"]
    User -- pets --> UserPets["GET,POST"]
    root -- store --> Store[""store]
    Pets --  "{petId}" --> Pet["GET,PATCH,DELETE"]
    Pet -- owner --> Owner["GET"] 
```

darrelmiller:
Added a show command to the hidi command line and also added a WriteMermaid function to the OpenApiUriTreeNode class.

Examples of the output are here https://github.com/darrelmiller/openapi-collection/tree/main/ApiDiagrams

The show command wraps the mermaid in a markdown file with a legend to explain the coloring of the nodes.  Each node is styled with a color that maps to the set of HTTP methods that are supported.  This mapping is defined by the MermaidColorScheme dictionary on OpenApiUriTreeNode.  That dictionary can be customized to change the colors or add new combinations of HTTP methods.

GitHub markdown does not render the HTML legend that the show command emits.  However, the VS code markdown preview does.

This closes #820


src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,63 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+        {
+            var path = string.IsNullOrEmpty(node.Path) ? "/" : SanitizeMermaidNode(node.Path);
+            foreach (var child in node.Children)
+            {
+                writer.WriteLine($"{path} --> {SanitizeMermaidNode(child.Value.Path)}[\"{child.Key}\"]");
+                ProcessNode(child.Value, writer);
+            }
+            var methods = String.Join("_", node.PathItems.SelectMany(p => p.Value.Operations.Select(o => o.Key))
+                .Distinct()
+                .Select(o => o.ToString().ToUpper())
```
baywet, line 280:
```suggestion
                .Select(static o => o.ToString().ToUpper())
```

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,63 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+                .OrderBy(o => o)
```
baywet, line 281:
```suggestion
                .Order()
```

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,63 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+        
+        private static void ProcessNode(OpenApiUrlTreeNode node, StreamWriter writer)
+        {
+            var path = string.IsNullOrEmpty(node.Path) ? "/" : SanitizeMermaidNode(node.Path);
+            foreach (var child in node.Children)
+            {
+                writer.WriteLine($"{path} --> {SanitizeMermaidNode(child.Value.Path)}[\"{child.Key}\"]");
+                ProcessNode(child.Value, writer);
+            }
+            var methods = String.Join("_", node.PathItems.SelectMany(p => p.Value.Operations.Select(o => o.Key))
```
baywet, line 278:
```suggestion
            var methods = String.Join("_", node.PathItems.SelectMany(static p => p.Value.Operations.Select(static o => o.Key))
```

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -532,5 +532,85 @@ private static ILoggerFactory ConfigureLoggerInstance(LogLevel loglevel)
...
+        {
+            using var loggerFactory = Logger.ConfigureLogger(logLevel);
+            var logger = loggerFactory.CreateLogger<OpenApiService>();
+            try
+            {
+                if (string.IsNullOrEmpty(openapi))
+                {
+                    throw new ArgumentNullException(nameof(openapi));
+                }
+                var stream = await GetStream(openapi, logger, cancellationToken);
```
baywet, line 546: shouldn't that be disposed?

baywet (changes requested):
a couple of comments on top of the ones already provided by codeql.
Also it seems you created your branch from master. I'd suggest rebasing from vnext for history sanity.

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
darrelmiller: Yeah, this one probably should.  I copy/pasted this code from one of the other commands, so I will fix it in the other places too.  In practice the executable is going to end after a command so it is not likely we are going to leak resources, but disposing the stream is a good habit.

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -532,5 +532,85 @@ private static ILoggerFactory ConfigureLoggerInstance(LogLevel loglevel)
...
+                    }
+                    stopwatch.Stop();
+                }
+
+                using (logger.BeginScope("Creating diagram"))
+                {
+                    // Create OpenApiUrlTree from document
+
+                    using var file = new FileStream(output.FullName, FileMode.Create);
+                    var writer = new StreamWriter(file);
```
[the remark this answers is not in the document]
darrelmiller: I did `using` on the FileStream.  Why would I do using on StreamWriter as it doesn't own any unmanaged resources?  That was sort of a rhetorical question. I hate that we have extended the use of Dispose to "also sometime do some other stuff".  All StreamWriter does in its dispose is call Flush. The only reason it implements IDisposable is because TextWriter is disposable but doesn't actually have any code at all in it's Dispose method.  I will accept this change and remove the redundant flush call.  I just felt the need to register my objection to this pattern.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
darrelmiller: What is this black magic? What is special about this lambda that it can be static?

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
darrelmiller: Ahhh, I had tried OrderBy() and that didn't work.  Thanks.

darrelmiller:
@baywet   You tricked me with that .Order() method.  It doesn't exist.  At least my version of .NET doesn't know about it.  I put the .OrderBy( o=>o) back.

baywet:
> @baywet You tricked me with that .Order() method. It doesn't exist. At least my version of .NET doesn't know about it. I put the .OrderBy( o=>o) back.

@darrelmiller ah right! it's an addition of .net 7, sorry about that, I'm in the future.
GlassesSunglassesGIF

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,152 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+                .Distinct()
+                .Select(o => o.ToString().ToUpper())
+                .OrderBy(o => o)
+                .ToList());
+        }
+
+        private static (string, string) GetShapeDelimiters(string methods)
+        {
+            
+            if (MermaidNodeStyles.ContainsKey(methods))
```
baywet, line 299:
```suggestion
            if (MermaidNodeStyles.TryGetValue(methods, out var style))
```

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,152 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+            {
+                //switch on shape
+                switch (MermaidNodeStyles[methods].Shape)
```
baywet, line 302:
```suggestion
                switch (style.Shape)
```

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,152 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+    /// </summary>
+    public class MermaidNodeStyle
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="color"></param>
+        /// <param name="shape"></param>
+        public MermaidNodeStyle(string color, MermaidNodeShape shape)
+        {
```
baywet, line 345: defensive programming? it seems that the constructor is used to make sure properties are initialized

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,152 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+                writer.WriteLine($"classDef {style.Key} fill:{style.Value.Color},stroke:#333,stroke-width:2px");
+            }
+
+            ProcessNode(this, writer);
+        }
+
+        /// <summary>
+        /// Dictionary that maps a set of HTTP methods to HTML color.  Keys are sorted, uppercased, concatenated HTTP methods.
+        /// </summary>
+        public static Dictionary<string, MermaidNodeStyle> MermaidNodeStyles = new Dictionary<string, MermaidNodeStyle>
```
baywet, line 258:
```suggestion
        public static Dictionary<string, MermaidNodeStyle> MermaidNodeStyles = new Dictionary<string, MermaidNodeStyle>(StringComparer.OrdinalIgnoreCase)
```

baywet, line 258: does this dictionary need to be public since it's only valuable to the internal process?
if yes, may I suggest to make it readonly? as well as to add it to the public APIs?

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,152 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+                    .Replace(".", "_")
+                    .Replace(";", "_")
+                    .Replace("-", "_")
+                    .Replace("default", "def_ault");  // default is a reserved word for classes
+        }
+    }
+    /// <summary>
+    /// Defines the color and shape of a node in a Mermaid graph diagram
+    /// </summary>
+    public class MermaidNodeStyle
```
baywet, line 337: does it need to be public?

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +236,152 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+        /// <summary>
+        /// 
+        /// </summary>
+        public MermaidNodeShape Shape { get;  }
+    }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public enum MermaidNodeShape
```
baywet, line 364: does it need to be public?

baywet (changes requested):

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
baywet: https://stackoverflow.com/questions/25603965/why-do-some-c-sharp-lambda-expressions-compile-to-static-methods
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/static-anonymous-functions

it's a perf thing, it effectively translates it to a static method instead of an instance method, and doesn't capture the surrounding context. But I'm now realizing that it's a C#9 feature (not available on standard 2.0)

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
darrelmiller: I specifically made it public so that users can customize the color schemes and shapes for combinations of methods.  I don't currently support all combination of methods.  So, someone can change the supported combinations.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
darrelmiller: Yes if we are to make it customizable.

baywet: then can you add it to the public APIs please?

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
baywet: then can you add it to the public APIs please? Also making it readonly so people can't affect it to null and derail the rest of your code.

darrelmiller:
OMG. I rebased successfully for the first time ever.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
darrelmiller: I changed my mind.  After playing around with the diagrams for a bit, I think the value of having consistency in diagram styles is more than the ability to customize.  We can always change our minds in the future.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
darrelmiller: I also made the ctor internal as the styles are no longer editable.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
darrelmiller: Unfortunately yes, even though I have made it read-only.  Hidi needs to read the list of styles to be able to generate the HTML legend for the diagrams.

baywet (dismissed):
Good job on the rebase, I hope you used the CLI :P

darrelmiller:
@baywet The git CLI, yes.  I was lucky because there were no merge conflicts.

MaggieKimani1:
Thank you @darrelmiller for taking this up. ☺

MaggieKimani1 (approved):

baywet (approved):

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -110,43 +111,13 @@ CancellationToken cancellationToken
...
-                        {
-                            logger.LogTrace("{timestamp}ms: Parsed OpenApi successfully. {count} paths found.", stopwatch.ElapsedMilliseconds, document.Paths.Count);
-                        }
-
-                        openApiFormat = format ?? GetOpenApiFormat(openapi, logger);
-                        openApiVersion = version != null ? TryParseOpenApiSpecVersion(version) : result.OpenApiDiagnostic.SpecificationVersion;
-                        stopwatch.Stop();
-                    }
+                    stopwatch.Restart();
+                    var result = await ParseOpenApi(openapi, logger, stream);
```
MaggieKimani1, line 115: The `TransformOpenApiDocument()` method uses these additional custom OpenApiReaderSettings in the OpenApiStreamReader which are missing in the new `ParseOpenApi()` method. Does this mean we no longer need these settings?
`LoadExternalRefs = inlineExternal`
`BaseUrl = openapi.StartsWith("http") ? new Uri(openapi) : new Uri("file:" + new FileInfo(openapi).DirectoryName + "\\")`

baywet (dismissed):

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
darrelmiller: My intent was that the behavior would be identical. I will review what settings got lost.

darrelmiller:
I'm still working on getting the test coverage up.  I'll try and get that done today or tomorrow.

darrelmiller:
Code coverage metric achieved.  :-)

src/Microsoft.OpenApi.Hidi/Program.cs
```diff
@@ -16,8 +16,19 @@ namespace Microsoft.OpenApi.Hidi
...
-        {            
-            var rootCommand = new RootCommand() {};
+        {
+            var rootCommand = CreateRootCommand();
+
+            // Parse the incoming args and invoke the handler
+            await rootCommand.InvokeAsync(args);
+
+            //// Wait for logger to write messages to the console before exiting
+            await Task.Delay(10);
```
baywet, line 26: Nope, When I ran into those issues with kiota it was usually because I was either not disposing a logging factory properly or because the command handler was using an overload that didn't use tasks, and wasn't waiting for completion of the task

baywet (changes requested):

baywet:
good job on the coverage. I encourage you to check the code smells to wrap this up (+ the comment I left) https://sonarcloud.io/project/issues?id=microsoft_OpenAPI.NET&pullRequest=1113&resolved=false&types=CODE_SMELL

src/Microsoft.OpenApi.Hidi/Program.cs
darrelmiller: Ack.  I'll see if I can find the source of this issue.

darrelmiller: Actually, this has been an issue for a while now and is unrelated to this this PR.  Let's not hold up this PR further and solve this issue independently.  I've opened an issue to track it #1127

darrelmiller:
This PR now also fixes #1124 and #1127

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -235,5 +237,155 @@ public void AddAdditionalData(Dictionary<string, List<string>> additionalData)
...
+        /// <param name="shape"></param>
+        internal MermaidNodeStyle(string color, MermaidNodeShape shape)
+        {
+            Color = color;
+            Shape = shape;
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
```
baywet, line 356: empty doc comments (multiple)

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -235,10 +217,10 @@
             return transform;
         }
 
-        private static Stream ApplyFilter(string csdl, string entitySetOrSingleton, XslCompiledTransform transform)
+        private static Stream ApplyFilterToCsdl(Stream csdlStream, string entitySetOrSingleton, XslCompiledTransform transform)
         {
             Stream stream;
-            StreamReader inputReader = new(csdl);
+            StreamReader inputReader = new(csdlStream);
```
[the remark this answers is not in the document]
baywet: the stream probably needs to be disposed

baywet (dismissed):

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
darrelmiller: Fixed the empty doc comments.  

The stream passed to ApplyFiltertoCsdl should be disposed by the caller, which it is.

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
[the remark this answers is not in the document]
baywet: I still maintain the CSDL stream should be disposed since it's not being returned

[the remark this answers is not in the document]
darrelmiller: I added the `using` to the StreamReader but I passed in `leaveOpen: true` because I don't believe that this method should dispose the `stream`.

MaggieKimani1 (approved):

baywet (approved):

PR #1116, opened 2022-12-31
verify string writer

SimonCropp:
Verify supports StringWriter. no need to convert it to a string


baywet (approved):
Thanks for the contribution!

Issue #1124, opened 2023-01-09
When CSDL filter and http resource supplied for HIDI input, it fails because ApplyFilter expects a local file but not a URL address


MazZzDaI:
When CSDL filter and http resource supplied for HIDI input, it fails because ApplyFilter expects a local file but not a URL address.

Error:
`Unhandled exception: System.InvalidOperationException: Could not transform the document, reason: The filename, directory name, or volume label syntax is incorrect. : 'C:\dev\OpenAPI.NET\src\Microsoft.OpenApi.Hidi\bin\Debug\net7.0\https:\share.dudarenko.dev\data\metadata'`

To reproduce, start HIDI with the following args:
` .\Microsoft.OpenApi.Hidi.exe transform --csdl-filter some-filter-file.txt --csdl https://share.dudarenko.dev/data/metadata`
where "https://share.dudarenko.dev/data/metadata" is the link to a CSDL file

Issue appears when both --csdl-filter and --csdl parameters specified.

Problem is in Microsoft.OpenApi.Hidi\OpenApiService.cs\ApplyFilter method. It accepts string, but from the invocation a stream passed.

PR #1125, opened 2023-01-09
Bugfix #1124 When CSDL filter and http resource supplied for HIDI input, it fails because ApplyFilter expects a local file but not a URL address

MazZzDaI:
Fix of the #1124 issue.

Overload added for _Microsoft.OpenApi.Hidi/OpenApiService.cs/ApplyFilter_ to accept both string and stream variables.

### DIFF
--- a/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
+++ b/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
@@ -98,7 +98,7 @@ namespace Microsoft.OpenApi.Hidi
                         if (!string.IsNullOrEmpty(csdlFilter))
                         {
                             XslCompiledTransform transform = GetFilterTransform();
-                            stream = ApplyFilter(csdl, csdlFilter, transform);
+                            stream = csdl.StartsWith("http") ? ApplyFilter(stream, csdlFilter, transform) : ApplyFilter(csdl, csdlFilter, transform);
                             stream.Position = 0;
                         }
 
@@ -235,10 +235,9 @@ namespace Microsoft.OpenApi.Hidi
             return transform;
         }
 
-        private static Stream ApplyFilter(string csdl, string entitySetOrSingleton, XslCompiledTransform transform)
+        private static Stream ApplyFilterForStreamReader(StreamReader inputReader, string entitySetOrSingleton, XslCompiledTransform transform)
         {
             Stream stream;
-            StreamReader inputReader = new(csdl);
             XmlReader inputXmlReader = XmlReader.Create(inputReader);
             MemoryStream filteredStream = new();
             StreamWriter writer = new(filteredStream);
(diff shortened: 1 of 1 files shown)


MazZzDaI:
@microsoft-github-policy-service agree [company="Olk inc."]

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -98,7 +98,7 @@ CancellationToken cancellationToken
                         if (!string.IsNullOrEmpty(csdlFilter))
                         {
                             XslCompiledTransform transform = GetFilterTransform();
-                            stream = ApplyFilter(csdl, csdlFilter, transform);
+                            stream = csdl.StartsWith("http") ? ApplyFilter(stream, csdlFilter, transform) : ApplyFilter(csdl, csdlFilter, transform);
```
baywet, line 101:
```suggestion
                            stream = csdl.StartsWith("http", StringComparison.OrdinalIgnoreCase) ? ApplyFilter(new StreamReader(stream), csdlFilter, transform) : ApplyFilter(new StreamReader(csdl), csdlFilter, transform);
```

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -235,20 +235,32 @@ private static XslCompiledTransform GetFilterTransform()
...
+            StreamReader inputReader = new(csdl);
+            return ApplyFilterForStreamReader(inputReader, entitySetOrSingleton, transform);
+        }
+
+        private static Stream ApplyFilter(Stream csdlStream, string entitySetOrSingleton, XslCompiledTransform transform)
+        {
+            StreamReader inputReader = new(csdlStream);
+            return ApplyFilterForStreamReader(inputReader, entitySetOrSingleton, transform);
+        }
+
```
baywet, line 263:
```suggestion
```

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -235,20 +235,32 @@ private static XslCompiledTransform GetFilterTransform()
             return transform;
         }
 
-        private static Stream ApplyFilter(string csdl, string entitySetOrSingleton, XslCompiledTransform transform)
+        private static Stream ApplyFilterForStreamReader(StreamReader inputReader, string entitySetOrSingleton, XslCompiledTransform transform)
```
baywet, line 238:
```suggestion
        private static Stream ApplyFilter(StreamReader inputReader, string entitySetOrSingleton, XslCompiledTransform transform)
```

baywet (changes requested):
Thanks for the contribution, here are a couple of comments to help improve it

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
darrelmiller: The original code here seems wrong.  GetStream does all the logic to determine if stream  is coming from a file or via HTTP.  Currently the code is getting the stream and then ignoring it if there is a filter.  The solution should be to change ApplyFilter to just accept the stream that comes from GetStream.

darrelmiller:
I was able to resolve this issue in PR #1113 in this commit https://github.com/microsoft/OpenAPI.NET/pull/1113/commits/7aac03ffd6f5e1fbe6d6c8cb07a280c1bacc47c4  
I also added support for using CSDL with the show command.

darrelmiller:
Thank you @MazZzDaI for identifying this issue.  PR #1113 has the fixes you are looking for and other related improvements.

PR #1138, opened 2023-01-23
Add root property jsonSchemaDialect

MaggieKimani1:
This PR:
- Adds a root property `jsonSchemaDialect` field and logic to serialize and deserialize
- Updates the `SerializeAsV3()` method by adding an optional `OpenApiSpecVersion` parameter in order to write out version-specific properties
- Adds a test for validation and refactors/cleans up existing tests


src/Microsoft.OpenApi/Interfaces/IOpenApiSerializable.cs
```diff
@@ -14,7 +14,8 @@ public interface IOpenApiSerializable : IOpenApiElement
         /// Serialize Open API element to v3.0.
         /// </summary>
         /// <param name="writer">The writer.</param>
-        void SerializeAsV3(IOpenApiWriter writer);
+        /// <param name="specVersion">The OpenApi specification version.</param>
```
baywet, line 17: What's the benefit of making it optional here? It would be miss-aligned with IOpenAPIExtension

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -293,8 +293,10 @@ public void SerializeAsV3(IOpenApiWriter writer)
                     }
                 });
             
-            // pathItems
-            writer.WriteOptionalMap(
+            // pathItems - only present in v3.1
+            if(version == OpenApiSpecVersion.OpenApi3_1)
```
baywet, line 297: should it be >= instead? or should we introduce a "futureMajorVersion" enum member to facilitate upgrading the code when 4.0 comes around?

src/Microsoft.OpenApi/Interfaces/IOpenApiSerializable.cs
MaggieKimani1: We don't really need the spec version when serializing most of the components in v3; we only use it to make a clear distinction when writing out  properties in v3.1 hence the need to make it optional.

baywet: My concern here is that by making it optional we'll get undesired default behaviour or that the actual value from the parent will be forgotten to be passed from the parent, loosing some context.

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
MaggieKimani1: Will moonwalk(4.0) have all the additional elements in 3.1 e.g. webhooks?

baywet: I think there are some breaking improvements around enums and other aspects like that.

src/Microsoft.OpenApi/Interfaces/IOpenApiSerializable.cs
darrelmiller: There should be a `SerializeAsV31`.   If the implementation of that method simply calls SerializeAsV3 that is fine.  The reason we created this as an interface was to guarantee that we explicitly covered every model object.

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -112,19 +118,36 @@ public void SerializeAsV3(IOpenApiWriter writer)
...
+
             // servers
             writer.WriteOptionalCollection(OpenApiConstants.Servers, Servers, (w, s) => s.SerializeAsV3(w));
 
             // paths
             writer.WriteRequiredObject(OpenApiConstants.Paths, Paths, (w, p) => p.SerializeAsV3(w));
 
             // webhooks
-            writer.WriteOptionalMap(
+            if (version == OpenApiSpecVersion.OpenApi3_1)
```
darrelmiller, line 148: We should try and avoid these kinds of conditional statements if possible.  We could refactor the common code into a common method and share it across the v3.1 and v3.0 serialization.  We don't need to worry about the order of properties.

src/Microsoft.OpenApi/Interfaces/IOpenApiSerializable.cs
darrelmiller: Feel free to challenge me on this.  That was the intended design, but now is the right time to question it.

baywet: The challenge with that approach is that it introduces a bunch of fork trees along the model.
And logic might need to be duplicated along the way. E.G. all the serialize as v2 need to call serialize as v2, all the v3 to call v3 and 31 to call 31. If one thing has a different behaviour on a leaf, I now need to stop calling serialize as v3 in 31 all along the tree for that simple leaf.

I'd be in favor of having a single serialize method, which accepts a mandatory parameter for the target version. Yes it'll require us to add a bunch of conditions but we won't have fork trees anymore.
Thoughts?

baywet: or a better alternative would be to have the serialize as v2, v3, v31 methods for each implementation and a Serialize method at the most common ancestor that would act as a router. This way we avoid the conditions AND the fork trees.
The serialize as v2/3/31 methods from OpenAPIDocument for example would call in the Serialize (not as something) method from the router, which in turns would call the right serialize as v from the implementer.
Thoughts?

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
MaggieKimani1: It seems like the order of properties changes when we use this approach. Here is a snippet of the resulting document after refactoring to use a common method:

image

Result:
```
    openapi: '3.1.0'
    info:
      title: JsonSchemaDialectTest
      version: 1.0.0
    paths: { }
    jsonSchemaDialect: http://json-schema.org/draft-07/schema#
```

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -95,14 +96,50 @@ public OpenApiComponents(OpenApiComponents components)
...
+            writer.WriteOptionalMap(
+            OpenApiConstants.PathItems,
+            PathItems,
+            (w, key, component) =>
             {
-                throw Error.ArgumentNull(nameof(writer));
-            }
+                if (component.Reference != null &&
+                    component.Reference.Type == ReferenceType.Schema &&
+                    component.Reference.Id == key)
```
baywet, line 114: using a string comparison with a string comparer instead

src/Microsoft.OpenApi/Models/OpenApiExtensibleDictionary.cs
```diff
@@ -38,15 +38,31 @@ public abstract class OpenApiExtensibleDictionary<T> : Dictionary<string, T>,
...
-            }
+            Serialize(writer);
+        }
+        
+        /// <summary>
+        /// Serialize to Open Api v3.0
+        /// </summary>
+        public void Serialize(IOpenApiWriter writer)
+        {
+            writer = writer ?? throw Error.ArgumentNull(nameof(writer));
```
baywet, line 65: move the body of this method to a serializeinternal method that accepts the version as a parameter so you can pass the correct version to the extensions below

src/Microsoft.OpenApi/Models/OpenApiMediaType.cs
```diff
@@ -85,7 +98,7 @@ public void SerializeAsV3(IOpenApiWriter writer, OpenApiSpecVersion version = Op
             writer.WriteOptionalMap(OpenApiConstants.Encoding, Encoding, (w, e) => e.SerializeAsV3(w));
 
             // extensions
-            writer.WriteExtensions(Extensions, version);
+            writer.WriteExtensions(Extensions, OpenApiSpecVersion.OpenApi3_0);
```
baywet, line 101: move to serialize internal

src/Microsoft.OpenApi/Models/OpenApiMediaType.cs
```diff
@@ -85,7 +98,7 @@ public void SerializeAsV3(IOpenApiWriter writer, OpenApiSpecVersion version = Op
             writer.WriteOptionalMap(OpenApiConstants.Encoding, Encoding, (w, e) => e.SerializeAsV3(w));
```
baywet, line 98: the callback should be an argument as well.

src/Microsoft.OpenApi/Models/OpenApiOAuthFlows.cs
```diff
@@ -87,7 +100,7 @@ public void SerializeAsV3(IOpenApiWriter writer, OpenApiSpecVersion version = Op
                 (w, o) => o.SerializeAsV3(w));
 
             // extensions
-            writer.WriteExtensions(Extensions, version);
+            writer.WriteExtensions(Extensions, OpenApiSpecVersion.OpenApi3_0);
```
baywet, line 103: move to internal

src/Microsoft.OpenApi/Models/OpenApiOperation.cs
```diff
@@ -186,7 +199,7 @@ public void SerializeAsV3(IOpenApiWriter writer, OpenApiSpecVersion version = Op
             writer.WriteOptionalCollection(OpenApiConstants.Servers, Servers, (w, s) => s.SerializeAsV3(w));
 
             // specification extensions
-            writer.WriteExtensions(Extensions, version);
+            writer.WriteExtensions(Extensions, OpenApiSpecVersion.OpenApi3_0);
```
baywet, line 202: move to internal

baywet (changes requested):

src/Microsoft.OpenApi/Interfaces/IOpenApiSerializable.cs
baywet: @MaggieKimani1 calling your attention on this comment seeing your recent changes. Let me know if you want me to expand a bit on it.

src/Microsoft.OpenApi/Models/OpenApiExtensibleDictionary.cs
MaggieKimani1: I can just update the Serialize method to accept the version as a parameter

baywet: so the IOpenApiSerializable interface would go back to having a single Serialize method that takes a mandatory version parameter (so we don't miss passing it along at places) and implements can either have a single method with if/else when their serialization logic is simple or break apart in multiple specialized private methods?

MaggieKimani1: No. We retain the serialize methods for v2, v3 and v31. However, since most of the logic between v3 and v31 is similar, I put that common logic in a Serialize method([as per Darrel's comment](https://github.com/microsoft/OpenAPI.NET/pull/1138#discussion_r1097984754)) but it will have a mandatory version parameter.

baywet: right, and that serialize method will be private? In that case I think we were talking about the same thing with different names in my original comment on this thread.

src/Microsoft.OpenApi/Models/OpenApiMediaType.cs
MaggieKimani1: If I understand you correctly, according to the [spec](https://spec.openapis.org/oas/v3.0.1#media-type-object), the media type object doesn't include a callback object.

src/Microsoft.OpenApi/Models/OpenApiExtensibleDictionary.cs
MaggieKimani1: Yes it will be private.

src/Microsoft.OpenApi/Models/OpenApiMediaType.cs
baywet: That's not what I was referring to.
this `(w, e) => e.SerializeAsV3(w)` should be a parameter of the method instead of being hard coded is what I meant.

MaggieKimani1: I think this approach might make it a bit convoluted, the action delegate accepts different types as arguments when writing out different objects or mappings. I think it should stay as is.

baywet: maybe but wouldn't keeping the hardcoded value lead to the implementer thinking they are supposed to serialize for v3 when we might be serializing for 3.1 for example?

MaggieKimani1: Oh, I see your point.

src/Microsoft.OpenApi/Interfaces/IOpenApiReferenceable.cs
```diff
@@ -25,7 +26,7 @@ public interface IOpenApiReferenceable : IOpenApiSerializable
         /// <summary>
         /// Serialize to OpenAPI V3 document without using reference.
         /// </summary>
-        void SerializeAsV3WithoutReference(IOpenApiWriter writer);
+        void SerializeAsV3WithoutReference(IOpenApiWriter writer, OpenApiSpecVersion version, SerializeDelegate callback);
```
baywet, line 29: shouldn't this get another method SerializeAsV31WithoutReference to be consistent?

In which case we wouldn't need the version or the callback arguments anymore and that'd become implementation details

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -15,6 +15,8 @@ namespace Microsoft.OpenApi.Extensions
     /// </summary>
     public static class OpenApiSerializableExtensions
     {
+        public delegate void SerializeDelegate(IOpenApiWriter writer, IOpenApiSerializable element);
```
baywet, line 18: can you provide some context on the introduction of the delegate here?

src/Microsoft.OpenApi/Interfaces/IOpenApiReferenceable.cs
MaggieKimani1: I can add that too.

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
MaggieKimani1: This was the only place I could think of that would expose the delegate to all implementers of the IOpenApiSerializable interface.

baywet: instead of using a delegate, why not using an Action?

```CSharp
public void foo(Action<IOpenApiSerializable, IOpenApiWriter> callback)
```

Sorry if I wasn't clear in my earlier comments, I never meant to expose the callback outside of the library, I meant that as an alternative to having to copy/paste the serialize method implementation.

MaggieKimani1: Noted.

baywet (commented):
Thanks for making these updates. The only remaining thing is this comment I had left earlier.
https://github.com/microsoft/OpenAPI.NET/pull/1138#discussion_r1114586942

Also, I'm not asking you to change that, but you could have simply passed `callback` instead of `(w, o) => callback(w, o)` in a lot of places. Saying that for future references for you :)

MaggieKimani1:
> Thanks for making these updates. The only remaining thing is this comment I had left earlier. [#1138 (comment)](https://github.com/microsoft/OpenAPI.NET/pull/1138#discussion_r1114586942)
> 
> Also, I'm not asking you to change that, but you could have simply passed `callback` instead of `(w, o) => callback(w, o)` in a lot of places. Saying that for future references for you :)

Yes, I'm working on that too. I have tests failing locally due to writer scope issues when writing start and end objects with all the serialization refactoring I've implemented. I want to fix these as well and the PR will be good to go. 
Also, thanks for the pointer above.

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -336,8 +331,28 @@ public void SerializeAsV3(IOpenApiWriter writer)
...
         }
 
+        private void RenderComponents(IOpenApiWriter writer)
+        {
+            var loops = writer.GetSettings().LoopDetector.Loops;
+            writer.WriteStartObject();
+            if (loops.TryGetValue(typeof(OpenApiSchema), out List<object> schemas))
+            {
+                var openApiSchemas = schemas.Cast<OpenApiSchema>().Distinct().ToList()
+                    .ToDictionary<OpenApiSchema, string>(k => k.Reference.Id);
```
baywet, line 344:
```suggestion
```

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -336,8 +331,28 @@ public void SerializeAsV3(IOpenApiWriter writer)
...
+
+                writer.WriteOptionalMap(
+                   OpenApiConstants.Schemas,
+                   Schemas,
+                   (w, key, component) => {
```
baywet, line 349:
```suggestion
                   static (w, key, component) => {
```

baywet (changes requested):

baywet (approved):

PR #1150, opened 2023-02-01
Update ValidationRuleSet.Rules to be an IEnumerable collection not IList

### LINKED ISSUE #1139: ValidationRuleSet.Rules should be IEnumerable not IList
Not only this performs unnecessary allocs, it leads people to think they can do operations like clearing the list, which won't impact the source. This fix would represent a breaking change and needs to be addressed at part of the work for v2.

https://github.com/microsoft/OpenAPI.NET/blob/1a42bc4b8c58617c4da0f417eb570993d21882dd/src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs#L113

MaggieKimani1:
Fixes #1139

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiYamlDocumentReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiYamlDocumentReader.cs
@@ -68,7 +68,7 @@ namespace Microsoft.OpenApi.Readers
             }
 
             // Validate the document
-            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count > 0)
+            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count() > 0)
             {
                 var openApiErrors = document.Validate(_settings.RuleSet);
                 foreach (var item in openApiErrors.OfType<OpenApiValidatorError>())
@@ -112,7 +112,7 @@ namespace Microsoft.OpenApi.Readers
             }
 
             // Validate the document
-            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count > 0)
+            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count() > 0)
             {
                 var openApiErrors = document.Validate(_settings.RuleSet);
                 foreach (var item in openApiErrors.OfType<OpenApiValidatorError>())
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi.Readers/OpenApiYamlDocumentReader.cs
```diff
@@ -68,7 +68,7 @@ public OpenApiDocument Read(YamlDocument input, out OpenApiDiagnostic diagnostic
             }
 
             // Validate the document
-            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count > 0)
+            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count() > 0)
```
baywet, line 71:
```suggestion
            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Any())
```

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -110,7 +110,7 @@ public ValidationRuleSet(IEnumerable<ValidationRule> rules)
         /// <summary>
         /// Gets the rules in this rule set.
         /// </summary>
-        public IList<ValidationRule> Rules
+        public IEnumerable<ValidationRule> Rules
```
baywet, line 113: To list below needs to be removed

src/Microsoft.OpenApi.Readers/OpenApiYamlDocumentReader.cs
```diff
@@ -193,7 +193,7 @@ private void ResolveReferences(OpenApiDiagnostic diagnostic, OpenApiDocument doc
             }
 
             // Validate the element
-            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count > 0)
+            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Count() > 0)
```
baywet, line 196:
```suggestion
            if (_settings.RuleSet != null && _settings.RuleSet.Rules.Any())
```

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -110,11 +110,11 @@ public ValidationRuleSet(IEnumerable<ValidationRule> rules)
...
```
baywet, line 113: _rules and empty rules could both be readonly, would you mind correcting this as well while at it?

baywet (changes requested):
Thanks of taking the changes, a couple more remarks

baywet (approved):
Thanks for making the changes

PR #1167, opened 2023-02-20
Chore: reduce default settings

thewahome:
### Summary
This is meant to support the work done in the [MSGraph-SDK-Code-Generator](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/944) repository

The configuration experience is already built in to the generation pipeline where new configurations are passed in depending on the platform's styling.

It is prudent to prune the default settings so that the overriding settings can be simplified and to guarantee that unneeded toggles are not propagated to Hidi

### DIFF
--- a/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
+++ b/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
@@ -328,22 +328,13 @@ namespace Microsoft.OpenApi.Hidi
             {
                 AddSingleQuotesForStringParameters = true,
                 AddEnumDescriptionExtension = true,
-                DeclarePathParametersOnPathItem = true,
                 EnableKeyAsSegment = true,
                 EnableOperationId = true,
-                ErrorResponsesAsDefault = false,
                 PrefixEntityTypeNameBeforeKey = true,
                 TagDepth = 2,
                 EnablePagination = true,
-                EnableDiscriminatorValue = true,
                 EnableDerivedTypesReferencesForRequestBody = false,
                 EnableDerivedTypesReferencesForResponses = false,
-                ShowRootPath = false,
-                ShowLinks = false,
-                ExpandDerivedTypesNavigationProperties = false,
-                EnableCount = true,
-                UseSuccessStatusCodeRange = true,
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -328,22 +328,13 @@ public static async Task<OpenApiDocument> ConvertCsdlToOpenApi(Stream csdl, stri
...
                 EnableKeyAsSegment = true,
                 EnableOperationId = true,
-                ErrorResponsesAsDefault = false,
                 PrefixEntityTypeNameBeforeKey = true,
                 TagDepth = 2,
                 EnablePagination = true,
-                EnableDiscriminatorValue = true,
                 EnableDerivedTypesReferencesForRequestBody = false,
                 EnableDerivedTypesReferencesForResponses = false,
-                ShowRootPath = false,
```
baywet, line 341: those settings are critical to Kiota based SDKs. If we want to move them somewhere else we first need to provide some kind of configuration experience on hidi's CLI and update the weekly generation pipeline in tandem. Otherwise we'll get a lot of side-effects.

baywet (changes requested):

thewahome:
Upon closer investigation of the processes, this pull request is not going to solve the problem.

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
thewahome: Kiota's configurations in the weekly generation pipeline have been saved in a file that will be consumed during generation.

baywet: thanks for doing that. At this point why do we need to change anything in hidi anymore then?
Or if we change something, wouldn't it be better to remove any default from hidi and rely on the defaults from OpenApi.net.OData when no settings are provided?

thewahome:
@baywet 
I've advanced the settings files in the metadata repository which means we would not need to do any more work anywhere else regarding this.
I am going to close this as the work here is no longer necessary

PR #1191, opened 2023-03-24
Suggested path for JsonSchema.Net integration

gregsdennis:
This branch is my recommended approach for using JsonSchema.Net.  It will require a few paradigm shifts for the internals of OpenApi.Net, but these changes are necessary to take full advantage of what JsonSchema.Net has to offer.

Opening this as a draft as it's not nearly ready and so that we can discuss the changes.

I think with the approach I have here, we can completely remove the Readers library.


test/Microsoft.OpenApi.Tests/Validations/OpenApiSchemaValidationTests.cs
```diff
@@ -279,37 +279,18 @@ public void ValidateOneOfSchemaPropertyNameContainsPropertySpecifiedInTheDiscrim
...
+                            .Type(SchemaValueType.Array)
+                            .Discriminator("type", null, null)
+                            .OneOf(
+                                new JsonSchemaBuilder()
+                                    .Properties(
+                                        ("type", new JsonSchemaBuilder().Type(SchemaValueType.Array))
+                                    )
+                                    .Ref("#/components/schemas/Person")
+                            )
+                            .Ref("Person")
```
gregsdennis, line 292: I'm not sure what the difference between these refs is supposed to be.

The original code is

```c#
                                    Reference = new OpenApiReference	
                                    {	
                                        Type = ReferenceType.Schema,	
                                        Id = "Person"	
                                    }	
                                }	
                            },	
                            Reference = new OpenApiReference { Id = "Person" }
```

What does this reference do?  My `.Ref()` extension creates a `$ref` keyword, but I don't think that's really what's supposed to be happening here.

test/Microsoft.OpenApi.Tests/Models/OpenApiComponentsTests.cs
```diff
@@ -144,144 +116,80 @@ public class OpenApiComponentsTests
...
-                        new OpenApiSchema
-                        {
-                            Type = "string"
-                        },
+                        new JsonSchemaBuilder().Type(SchemaValueType.String),
                         null,
                         null
-                    }
-                }
+                    )
```
gregsdennis, line 132: This is defining the `BrokenComponents` object for this test, but submitting `null` here is more than broken; it's completely invalid.  I'm not sure JsonSchema.Net will even support doing this, and even if it does, you'll likely get a `NullReferenceException`.

src/Microsoft.OpenApi.Readers/OpenApiTextReaderReader.cs
```diff
@@ -50,7 +51,12 @@ public OpenApiDocument Read(TextReader input, out OpenApiDiagnostic diagnostic)
                 return new OpenApiDocument();
             }
 
-            return new OpenApiYamlDocumentReader(this._settings).Read(yamlDocument, out diagnostic);
+            var asJsonNode = yamlDocument.ToJsonNode();
```
darrelmiller, line 54: I believe we are going to make the switch to do Yaml => JsonNodes => OpenAPIDocument.  This will facilitate us to use the JsonSchema.Net Examples and hopefully get rid of OpenApiAny and all of the OpenApiPrimitive classes.

src/Microsoft.OpenApi.Readers/OpenApiTextReaderReader.cs
```diff
@@ -50,7 +51,12 @@ public OpenApiDocument Read(TextReader input, out OpenApiDiagnostic diagnostic)
...
+
+            diagnostic = null; // TBD
+            return asJsonNode.Deserialize<OpenApiDocument>();
```
darrelmiller, line 57: I honestly don't see how using Deserialize<T> is going to work for us.  Using an automatic deserialization mechanism strongly couples the serialized format to the in-memory model.  We specifically wrote the deserializer classes so we can deserialize any version of OpenAPI into the "latest" in-memory model.  

It is not clear to me from this PR how you are proposing we continue allowing our users to import OpenAPI without caring what version of OAS the input format is.  And to clarify, this is one of our most important features.

src/Microsoft.OpenApi.Readers/YamlConverter.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Linq;
+using System.Text.Json.Nodes;
+using Json.More;
+using SharpYaml;
+using SharpYaml.Serialization;
+
+namespace Microsoft.OpenApi.Readers
+{
+    /// <summary>
+    /// Provides extensions to convert YAML models to JSON models.
```
darrelmiller, line 12: This is awesome.  Thanks!  Maggie is currently doing a spike to try and convert the ParseNode classes to work off JsonNode instead of YamlNode.  If she is able to do that, then this class will enable us to continue supporting YAML parsing.

src/Microsoft.OpenApi/Models/OpenApiHeader.cs
```diff
@@ -82,17 +77,17 @@ public class OpenApiHeader : IOpenApiSerializable, IOpenApiReferenceable, IOpenA
         /// <summary>
         /// Examples of the media type.
         /// </summary>
-        public IDictionary<string, OpenApiExample> Examples { get; set; } = new Dictionary<string, OpenApiExample>();
+        public IDictionary<string, OpenApiExample> Examples { get; set; }
```
darrelmiller, line 80: It was an intentional design choice to ensure collections were non-null so that library users did not have to check if collections had been created before adding properties to the collection.  However, due to the performance cost we should do a lazy instantiation of these properties in the get accessor instead of doing it at create time.

test/Microsoft.OpenApi.Tests/Validations/OpenApiSchemaValidationTests.cs
darrelmiller: Components have a self-reference.  It's an ugly implementation detail due to the way that we load $refs in as a dummy empty object (because the target of the reference hasn't been parsed yet) and then post-reading we run a reference resolver to replace the dummy objects with the target object.  This means the inline object and the component object are actually the same object instance.  Therefore, objects in the components collection also need to have a Reference property set or the writers cannot write out the $ref.

This model is going to change OpenAPI.Net v2 to a proxy pattern.  We need to do this because in OpenAPI v3.1 you can now have summary and description properties that sit along side the $ref.  With the current model we have no place to store that additional information.

Anyway, I think the proxy pattern for handling references is going to be a much better model anyway as it will not require doing the post-reading reference resolution step.  It also makes writing much easier.  You can see my attempt to implement the proxy pattern here https://github.com/microsoft/OpenAPI.NET/blob/dm/ProxyReference/src/Microsoft.OpenApi/Models/OpenApiSchemaReference.cs However, I did it on OpenAPISchema which is no longer appropriate since 3.1 made the distinction between OpenAPI reference objects and JSON Schema $ref.

gregsdennis: Okay, so I can just remove all of those, probably.  I'll need to go through and re-update all of those tests.  (I'll probably end up closing this and cherry-picking it to another branch for a different PR.)

src/Microsoft.OpenApi.Readers/OpenApiTextReaderReader.cs
gregsdennis: Strictly speaking, this change isn't necessary because this code will be deleted.

> Using an automatic deserialization mechanism strongly couples the serialized format to the in-memory model.

How so?  I'm doing the actual parsing with YamlSharp; both YAML and JSON are supported through this.  It parses into the `YamlDocument` model, which I then translate to `JsonNode`.  Then I just use the built-in `JsonSerializer` to deserialize it into `OpenApiDocument`.

Going the other way, if you want JSON, you just use `JsonSerializer.Serialize(openApiDoc)`.  If you want YAML, you use [this method](https://github.com/microsoft/OpenAPI.NET/pull/1191/files#diff-83c749e84579435c1783f1791fc5271c6ba649262675829f0c78b2d322d5d335R219) which we'll need to expose as an extension.

> We specifically wrote the deserializer classes so we can deserialize any version of OpenAPI into the "latest" in-memory model.

This supports the same thing.  I can drop a 2.0 or a 3.1 schema into my `DevTest` test and it still deserializes just fine.  Users don't need to care which OpenAPI version they have.

src/Microsoft.OpenApi.Readers/YamlConverter.cs
gregsdennis: This code is basically copied from my YamlToJsonNode package.  That lib uses YamlDotNet which is the basis for YamlSharp, so it's the same code for "different" types.

src/Microsoft.OpenApi/Models/OpenApiHeader.cs
gregsdennis: These initializations cause these properties to serialize.  Lazy initialization will do the same.  Not using custom JSON converters is preferred, where possible.

For ease-of-use, personally, I think a fluent builder is better.

src/Microsoft.OpenApi.Readers/OpenApiTextReaderReader.cs
darrelmiller: How does the JsonDeserializer.Deserialize convert body parameters into a requestBody object?  How does the JsonSerializer.Deserialize know how to convert produces/consumes into media type objects?  We have many of these kinds of translations that System.Text.Json knows nothing about.

gregsdennis: I don't follow.  The paths use `OpenApiParameter` which isn't associated with `OpenApiRequestBody` anywhere that I can see.

darrelmiller: In our V2 Deserializer, we convert parameters that are of type `body` or `form` into a request body.  See code here:

https://github.com/microsoft/OpenAPI.NET/blob/vnext/src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs#L126-L138

src/Microsoft.OpenApi.Readers/YamlConverter.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        private static JsonArray ToJsonArray(this YamlSequenceNode yaml)
+        {
+            var node = new JsonArray();
+            foreach (var value in yaml)
+            {
+                node.Add(value.ToJsonNode());
+            }
+
+            return node;
```
baywet, line 97:
```suggestion
            return new JsonArray(yaml.Select(static x => x.ToJsonNode()).ToArray());
```

src/Microsoft.OpenApi/Models/OpenApiCallback.cs
```diff
@@ -19,7 +19,6 @@ public class OpenApiCallback : IOpenApiSerializable, IOpenApiReferenceable, IOpe
         /// A Path Item Object used to define a callback request and expected responses.
         /// </summary>
         public Dictionary<RuntimeExpression, OpenApiPathItem> PathItems { get; set; }
```
baywet, line 21: NRT is going to complain if we remove default initializations without adding the ?

gregsdennis: NRT isn't enabled for some reason, so I left it.

src/Microsoft.OpenApi.Readers/YamlConverter.cs
```diff
@@ -0,0 +1,133 @@
...
+    public static class YamlConverter
+    {
+        /// <summary>
+        /// Converts all of the documents in a YAML stream to <see cref="JsonNode"/>s.
+        /// </summary>
+        /// <param name="yaml">The YAML stream.</param>
+        /// <returns>A collection of nodes representing the YAML documents in the stream.</returns>
+        public static IEnumerable<JsonNode> ToJsonNode(this YamlStream yaml)
+        {
+            return yaml.Documents.Select(x => x.ToJsonNode());
```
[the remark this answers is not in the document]
baywet: Evidence would suggest otherwise
https://github.com/microsoft/OpenAPI.NET/blob/df1b405e537d0f3c2a829b6a80fef82959d0eab5/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L286

src/Microsoft.OpenApi/Models/OpenApiCallback.cs
gregsdennis: That's not my goal with this PR. Let's keep it focused to integrating my lib.

baywet: I understand, but removing default values goes against that other goal. Could you example on why you removed default values?

gregsdennis: I explained why I removed them in the comment where you asked why I removed them.

baywet: I'm not sure I'm understanding, are you referring to this ?
> NRT isn't enabled for some reason, so I left it.

gregsdennis: Ah, no, you asked in slack, and I replied there.

Also, Darrel asked the same above.

baywet: I eventually figured :D
Thanks for the additional context. Could we imagine using the setting of system.text.json to avoid serializing empty nodes (empty objects or empty properties) instead?

gregsdennis: I'm open to that if you can find such an option. I didn't look very hard.

baywet: I'm guessing you'd do that with [DefaultIgnoreCondition](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.defaultignorecondition?view=net-8.0) and set it to `WhenWritingDefault` (on the options, otherwise you can use the enum in JsonIgnoreAttribute)

gregsdennis: That only works when the property value is null. It doesn't work for empty collections.

baywet: alright, thanks for the context here. Changing those default values probably has some side effect on the copy constructors and such, but unit tests should reveal that. We'll enable NRT as a separate effort.

baywet: created a dedicated issue https://github.com/microsoft/OpenAPI.NET/issues/1202

src/Microsoft.OpenApi.Readers/YamlConverter.cs
gregsdennis: Your suggestion allocates an extra array.

src/Microsoft.OpenApi/Models/OpenApiCallback.cs
gregsdennis: > Changing those default values probably has some side effect on the copy constructors and such, but unit tests should reveal that.

This PR has one focus:  use JsonSchema.Net and STJ to handle serialization.  We may (probably will) need to add custom converters at some point, but this PR, in its current state, is proof of concept.  For right now, I'm more worried about "can it be done and, if so, how?" and less worried about dev experience.

src/Microsoft.OpenApi.Readers/YamlConverter.cs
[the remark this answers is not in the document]
gregsdennis: Sure, okay.  I'm not worried about these nuances right now.  Let's please focus on making it work.

src/Microsoft.OpenApi.Readers/YamlConverter.cs
baywet: right, sorry I meant this

```suggestion
            return new JsonArray(yaml.Select(static x => x.ToJsonNode()));
```

gregsdennis:
> We using this deserialization pattern extensively in our SDK generation and are able to parse unrecognized properties into a dictionary without any problem. - @darrelmiller

My concern was less about unknown keywords and more about keywords defined in custom vocabularies, which, if you're supporting 2020-12, you will need to be able to support somehow, probably through some extension mechanism where people can define their own keyword implementations.

PR #1195, opened 2023-03-28
OpenApiOperation constructor preserve null

### LINKED ISSUE #1192: OpenApiOperation copy constructor - RequestBody null not preserved
**Describe the bug**
OpenApiOperation copy constructor: `public OpenApiOperation(OpenApiOperation operation)` changes a `null` in the `RequestBody` into an object instance.

**To Reproduce**
```C#
var operation = new OpenApiOperation();
//operation.RequestBody = null is the default
var copiedOperation = new OpenApiOperation(operation);
```
`copiedOperation.RequestBody` is an `new OpenApiRequestBody`.

**Expected behavior**
`copiedOperation.RequestBody` should also be null, as on the source object. 

https://github.com/microsoft/OpenAPI.NET/blob/db02b957533b0fe9e0d880d580ad4c99f2bb0671/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L117-L132

I could provide a PR to fix this.

dldl-cmd:
Fixes #1192

 - Fix OpenApiOperation RequestBody and Tags null is not preserved
 - Add tests for null preserving null on copy constructor
 - Add tests for copy constructor serialization result to be equal
 - Fix OpenApiParameter object changes during serialization, so that the serialization result is on consecutive run the same.


src/Microsoft.OpenApi/Models/OpenApiOperation.cs
```diff
@@ -116,13 +116,13 @@ public class OpenApiOperation : IOpenApiSerializable, IOpenApiExtensible
         /// </summary>
         public OpenApiOperation(OpenApiOperation operation)
         {
-            Tags = new List<OpenApiTag>(operation?.Tags);
+            Tags = operation.Tags != null ? new List<OpenApiTag>(operation?.Tags) : null;
```
MaggieKimani1, line 119: Tags = operation?.Tags != null ? new List<OpenApiTag>(operation?.Tags) : null;

MaggieKimani1, line 119:
```suggestion
            Tags = operation?.Tags != null ? new List<OpenApiTag>(operation?.Tags) : null;
```

MaggieKimani1 (changes requested):

baywet:
Thanks for the contribution. @MaggieKimani1 can you merge this PR when you have a couple of minutes please so it goes out with the next preview?

Issue #1089, opened 2022-11-22
In OpenAPI.NET v2 can we get rid of IOpenApiAny


darrelmiller:
I believe our use of OpenApiAny for examples was a mistake.  We should simply use strings.
The only other places we use OpenApiAny is in OpenApiSchema.Default and in Extensions.
When we move to a third party JSONSchema library, it will not use IOpenApiAny for default.
Extensions we could just treat as strings unless a strongly typed extension handler was registered.

Issue #841, opened 2022-04-26
Use System.Text.Json for parsing JSON instead of YamlSharp


darrelmiller:
System.Text.Json is significantly faster than YamlSharp for parsing JSON and it doesn't have an issue with the key length limit that 
YamlSharp has. #786 

This will require detecting if the description being read is JSON or YAML which we can do from either the file extension or the content-type header.

We need to investigate what will be required to implement a new set of parse nodes that wrap a JsonDocument.

PR #1211, opened 2023-04-11
Use System.Text.JsonNodes for parsing

### LINKED ISSUE #786: Parsing error when path key exceeds 1023 characters
Using v1.2.3 (also tried with 1.3.1-preview), I am getting a parsing error if the OA3 json has a path key that exceeds 1023 characters.  For example, take the PetStore reference OA3 json and change the `/pets` path key to be extremely long. Attempt to read/parse the file, and get a parsing error:
```
Message: (Lin: 15, Col: 1032, Chr: 1302) - (Lin: 15, Col: 1033, Chr: 1303): While parsing a flow mapping,  did not find expected ',' or '}'.
Pointer: #line=15
```
Is this a known limitation? I don't see anything in the [OA3 spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#pathsObject) about a limit on key sizes.  Is there any workaround?
Yes, the key seems excessively long, but it is generated by a customer's application which my product does not control.

Edit: Also, the error is fatal in that no other path entries get parsed - the Paths property on the oaDoc is empty.

**Sample Code:**
```
         var filepath = @"c:\temp\petstore-3.0-with-a-really-long-path.json";
         OpenApiStreamReader reader = new OpenApiStreamReader ();
         using (Stream file = new FileStream (filepath, FileMode.Open))
         {
            var oaDoc = reader.Read (file, out OpenApiDiagnostic diags);
            // expected: json to be parsed successfully
            // observed: error found 
            if (diags.Errors.Count > 0)
            {
               var error = diags.Errors.First ();
               Console.WriteLine ($"Found error(s), msg: {error.Message} ptr: {error.Pointer}");
            }
         }
```
**Sample (long) Key:**
/pets-with-a-really-long-path-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123456789-0123

MaggieKimani1:
Fixes #786, #841 , #1089


src/Microsoft.OpenApi.Readers/OpenApiTextReaderReader.cs
```diff
@@ -1,13 +1,17 @@
...
 using System.IO;
 using System.Linq;
+using System.Text.Json;
+using System.Text.Json.Nodes;
 using System.Threading.Tasks;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Readers.Interface;
 using SharpYaml;
+//using SharpYaml;
```
baywet, line 14:
```suggestion
```

src/Microsoft.OpenApi.Readers/Exceptions/OpenApiReaderException.cs
```diff
@@ -38,13 +39,13 @@ public class OpenApiReaderException : OpenApiException
         /// </summary>
         /// <param name="message">Plain text error message for this exception.</param>
         /// <param name="node">Parsing node where error occured</param>
-        public OpenApiReaderException(string message, YamlNode node) : base(message)
+        public OpenApiReaderException(string message, JsonNode node) : base(message)
         {
             // This only includes line because using a char range causes tests to break due to CR/LF & LF differences
             // See https://tools.ietf.org/html/rfc5147 for syntax
-            Pointer = $"#line={node.Start.Line}";
+            //Pointer = $"#line={node.Start.Line}";
```
baywet, line 46: commented out code

src/Microsoft.OpenApi.Readers/ParseNodes/JsonPointerExtensions.cs
```diff
@@ -12,32 +13,32 @@ namespace Microsoft.OpenApi.Readers.ParseNodes
...
                 foreach (var token in currentPointer.Tokens)
                 {
-                    var sequence = pointer as YamlSequenceNode;
+                    var array = pointer as JsonArray;
 
-                    if (sequence != null)
+                    if (array != null)
                     {
-                        pointer = sequence.Children[Convert.ToInt32(token)];
+                        pointer = array[Convert.ToInt32(token)];
```
baywet, line 34: maybe use a try convert and hoist the conversion to the parent condition instead, it'll be safer

src/Microsoft.OpenApi.Readers/ParseNodes/ListNode.cs
```diff
@@ -5,48 +5,44 @@
...
+            _nodeList = jsonArray;
         }
 
         public override List<T> CreateList<T>(Func<MapNode, T> map)
         {
             if (_nodeList == null)
             {
-                throw new OpenApiReaderException(
-                    $"Expected list at line {_nodeList.Start.Line} while parsing {typeof(T).Name}", _nodeList);
+                //throw new OpenApiReaderException($"Expected list at line {_nodeList.Start.Line} while parsing {typeof(T).Name}", _nodeList);
```
baywet, line 26: commented out code

src/Microsoft.OpenApi.Readers/ParseNodes/MapNode.cs
```diff
@@ -59,32 +52,27 @@ internal class MapNode : ParseNode, IEnumerable<PropertyNode>
...
                     T value;
                     try
                     {
                         Context.StartObject(key);
-                        value = n.Value as YamlMappingNode == null
-                          ? default(T)
-                          : map(new MapNode(Context, n.Value as YamlMappingNode));
+                        value = n.Value as JsonObject == null
+                          ? default
+                          : map(new MapNode(Context, n.Value as JsonObject));
```
baywet, line 66:
```suggestion
                        value = n.Value is JsonObject jsonObject
                          ? map(new MapNode(Context, jsonObject))
                          : default;
```

src/Microsoft.OpenApi.Readers/YamlHelper.cs
```diff
@@ -1,34 +1,38 @@
...
         {
-            var scalarNode = node as YamlScalarNode;
-            if (scalarNode == null)
+
+            var scalarNode = node as JsonValue;
+            if (node == null)
             {
-                throw new OpenApiException($"Expected scalar at line {node.Start.Line}");
-            }
+                //throw new OpenApiException($"Expected scalar at line {node.Start.Line}");
```
baywet, line 22: commented out code,
pattern matching

src/Microsoft.OpenApi/Extensions/JsonNodeExtension.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Text.Json.Nodes;
+using Microsoft.OpenApi.Writers;
+
+namespace Microsoft.OpenApi.Extensions
+{
+    internal static class JsonNodeExtension
+    {
+        //private static void Write(this JsonNode, IOpenApiWriter writer) => writer.WriteValue(this);
```
baywet, line 11: remove file all together?

src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
```diff
@@ -191,7 +190,7 @@ internal OpenApiBodyParameter ConvertToBodyParameter()
             };
             if (bodyParameter.Extensions.ContainsKey(OpenApiConstants.BodyName))
             {
-                bodyParameter.Name = (Extensions[OpenApiConstants.BodyName] as OpenApiString)?.Value ?? "body";
+                bodyParameter.Name = (Extensions[OpenApiConstants.BodyName].ToString()) ?? "body";
```
baywet, line 193: avoid null coalesce with strings since it doesn't take care of empty strings. use type assertion + string.IsNullOrEmpty instead

src/Microsoft.OpenApi/Writers/OpenApiWriterAnyExtensions.cs
```diff
@@ -30,54 +32,62 @@ public static void WriteExtensions(this IOpenApiWriter writer, IDictionary<strin
...
-                    item.Value.Write(writer, specVersion);
+                    
+                    if (item.Value == null)
+                    {                        
+                        writer.WriteNull();
+                    }
+                    else
+                    {
+                        writer.WriteAny(item.Value);
+                        //item.Value.Write(writer, specVersion);
```
baywet, line 43: commented out code

baywet (changes requested):
This is great work! I especially like the fact that we're getting rid of so much code and taking advantage of generics here. Left a couple of comments to help wrap this up.

src/Microsoft.OpenApi.Readers/ParseNodes/JsonPointerExtensions.cs
MaggieKimani1: Do you mind providing an example for this?

baywet: assuming the token value is a string you can update the condition above to 
```CSharp
if (array != null && int32.TryParse(token, out var tokenValue))
```

and then use the new tokenValue in the indexer

darrelmiller:
This is looking good.  I ran a bunch of manual tests and it looks to be handling examples and schemas well. If we can fix the remaining failing tests, I think we are good to merge this.

We cannot yet remove the Yaml parser, even for JSON because we have a problem with the way we handle strings that are "NaN".  I believe the Yaml parser converts that into an unquoted NaN and writes that into the JSON payload.  And then the Json parser will not read it.  I'm not quite sure how to fix this problem yet.

MaggieKimani1:
> Why is the CI/CD step passing when there are failing tests?

It's being skipped actually. I think it was disabled to run on PRs - not sure.

MaggieKimani1:
> This is looking good. I ran a bunch of manual tests and it looks to be handling examples and schemas well. If we can fix the remaining failing tests, I think we are good to merge this.
> 
> We cannot yet remove the Yaml parser, even for JSON because we have a problem with the way we handle strings that are "NaN". I believe the Yaml parser converts that into an unquoted NaN and writes that into the JSON payload. And then the Json parser will not read it. I'm not quite sure how to fix this problem yet.

Initially I thought it was a matter of "it doesn't work on my machine" as the tests were failing intermittently. However, I see that its caused by an external dependency library that helps generate the diffs for the XUnit.Verify tool when rendering snapshots. I found a couple of issues opened on their repo targeting the same problem e.g. https://github.com/approvals/ApprovalTests.Net/issues/638 so I'm hoping a Visual Studio reinstall/package downgrade might help as well as its not a code issue.

Please point me to the Yaml parser you're referring to as per changes in this PR.

darrelmiller:
I think we just need to reset the baseline tests.  If we delete the stored test result and allow the verifier to re-create the tests results then that should fix it.  However, in one of the tests it appears like we used base64 encode binary but we are not doing that before.  Changing that behavior would be a breaking change.

Regarding the yaml parser, no don't change what you have.  We just need to do some work to make sure when we convert YamlNodes to JsonNodes that a "NaN" value stays as a string in the JsonNode.  It is possible that it might just work if we use this new branch to transform Graph Yaml into a Graph JSON document.  I just haven't tried that yet.

darrelmiller (approved):

Issue #1212, opened 2023-04-11
Stack Overflow with new cloning changes


baywet:
1.6.4-preview2 (just released earlier) produces stack overflows. Running dotnet test in [this PR](https://github.com/microsoft/kiota/pull/2534) triggers the behavior. The DictionaryCloneHelper is probably missing some exist conditions.

```
at Microsoft.OpenApi.Models.OpenApiSchema..ctor(Microsoft.OpenApi.Models.OpenApiSchema)
   at Microsoft.OpenApi.Models.OpenApiMediaType..ctor(Microsoft.OpenApi.Models.OpenApiMediaType)
   at DynamicClass.InvokeStub_OpenApiMediaType..ctor(System.Object, System.Object, IntPtr*)
   at System.Reflection.ConstructorInvoker.Invoke(System.Object, IntPtr*, System.Reflection.BindingFlags)
   at System.Reflection.RuntimeConstructorInfo.Invoke(System.Reflection.BindingFlags, System.Reflection.Binder, System.Object[], System.Globalization.CultureInfo)
   at System.RuntimeType.CreateInstanceImpl(System.Reflection.BindingFlags, System.Reflection.Binder, System.Object[], System.Globalization.CultureInfo)
   at System.Activator.CreateInstance(System.Type, System.Object[])
   at Microsoft.OpenApi.Helpers.DictionaryCloneHelper.Clone[[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](System.Collections.Generic.IDictionary`2<System.__Canon,System.__Canon>)
   at Microsoft.OpenApi.Models.OpenApiResponse..ctor(Microsoft.OpenApi.Models.OpenApiResponse)
   at DynamicClass.InvokeStub_OpenApiResponse..ctor(System.Object, System.Object, IntPtr*)
   at System.Reflection.ConstructorInvoker.Invoke(System.Object, IntPtr*, System.Reflection.BindingFlags)
   at System.Reflection.RuntimeConstructorInfo.Invoke(System.Reflection.BindingFlags, System.Reflection.Binder, System.Object[], System.Globalization.CultureInfo)
[...]

Issue #1193, opened 2023-03-28
Copy constructor doesn't create new `OpenApiPathItem` objects when creating new instance of `OpenApiPaths`


irvinesunday:
**Describe the bug**
Copy constructors doesn't create a new instance of `OpenApiPathItem` and its constituent properties on `new OpenApiDocument(document)`.

**To Reproduce**
I have created a breaking test to demonstrate this:
https://github.com/microsoft/OpenAPI.NET/blob/d5a5f8a8890825824f31187355850945416aa7ea/test/Microsoft.OpenApi.Tests/Models/OpenApiDocumentTests.cs#L1341-L1358
This is in the branch: `is/copy-ctrs`

**Expected behavior**
In the test, I change the operationId of the cloned OpenAPI document. I don't expect the `operationId` of the original OpenAPI document to also be updated.

darrelmiller:
When cloning OpenApiPaths we are using the standard Dictionary constructor which does not do a clone of the values of the key value pairs.
https://github.com/microsoft/OpenAPI.NET/blob/vnext/src/Microsoft.OpenApi/Models/OpenApiExtensibleDictionary.cs#L31

MaggieKimani1:
Reopening this issue as its not fixed.

baywet:
This is not a bug, it's by design. Dictionary creates shallow copies, removing the derived constructor would be a breaking change at this point. And the rest of the cloning infrastructure is using offering shallow copies as well.

I've amended the doc comments in #2947 to address that.

And if we get an issue/popular request to implement deep copies across the object model, we should do that at a later stage.

Let me know if you have any additional comments or questions.

PR #1219, opened 2023-04-17
Revert dictionary clone helper logic 

MaggieKimani1:
Reverts the DictionaryCloneHelper logic for cloning dictionaries in order to unblock Kiota. This fix was a result of [this issue](https://github.com/microsoft/OpenAPI.NET/issues/1193) which introduced [this bug](https://github.com/microsoft/OpenAPI.NET/issues/1212) that is blocking Kiota. Since DeX API has a Cloning workaround, I've reverted this logic for us to unblock Kiota as we take more time to investigate on how to prevent the StackOverflow exception generated as a result of circular object dependencies when cloning an OpenApiDocument.

Example: 
```
System.Text.Json.JsonException : A possible object cycle was detected. This can either be due to a cycle or if the object depth is larger than the maximum allowed depth of 64. Consider using ReferenceHandler.Preserve on JsonSerializerOptions to support cycles.
Path: $.Operations.Responses.Content.Schema.AllOf.Reference.HostDocument.Paths.Operations.Responses.Content.Schema.AllOf.Reference.HostDocument.Paths.Operations.Responses.Content.Schema.AllOf.Reference.HostDocument.Paths.Operations.Responses.Content.Schema.AllOf.Reference.HostDocument.Paths.Operations.Responses.Content.Schema.AllOf.Properties.Type.
```


src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
```diff
@@ -33,6 +33,9 @@
     <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
         <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
     </PropertyGroup>
+    <ItemGroup>
```
baywet, line 36: I'm not sure how this change is relevant to this scope?

src/Microsoft.OpenApi/Any/OpenApiAnyCloneHelper.cs
```diff
@@ -2,6 +2,9 @@
 // Licensed under the MIT license.
 
 using System.Reflection;
+using Microsoft.OpenApi.Helpers;
+using System.Text.Json.Serialization;
+using System.Text.Json;
```
baywet, line 7: revert? (git checkout src/Microsoft.OpenApi/Any/OpenApiAnyCloneHelper.cs main)

src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
```diff
@@ -33,6 +33,9 @@
     <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
```
baywet, line 33: do we need to bump version in this PR or in another one?

baywet (changes requested):

baywet (approved):

Issue #1140, opened 2023-01-24
Validation rules should have a unique key to drive it by configuration


baywet:
**Is your feature request related to a problem? Please describe.**
OpenApi.net currently provides a set of validation rules, and the ability for library consumers to define their own. This is a great first step in a linting experience. However driving configuration of those rules is nearly impossible today.

**Describe the solution you'd like**
Most linters (eslint, etc...) name their validation rules with a unique key to allow enablement/disablement via configuration. As well as configuration of the rules themselves.
In addition to that change, the ValidationRuleSet class should provide methods to CRUD validation rules by key to allow easy manipulation of them at runtime.
It should also provide a clear method that clears all the current configured validation rules.
This will need to be implemented in v2 as it's most likely a breaking change.

**Describe alternatives you've considered**
Matching by typename (not target type) would be quite hard because a lot of the validation rules are lambdas and don't have a trivial type name.

**Additional context**
Kiota rules are all named today https://microsoft.github.io/kiota/using.html#--disable-validation-rules---dvr
Ran into the issue while trying to get the dvr switch in kiota to disable OpenAPI.net default rules when the all value is provided.
https://github.com/microsoft/kiota/pull/2182

darrelmiller:
> a lot of the validation rules are lambdas

Yeah, I'm not sure why we did this.  But considering we did. I think I agree with your assertion.

irvinesunday:
Resolved by: https://github.com/microsoft/OpenAPI.NET/pull/1256

PR #1256, opened 2023-05-29
[Breaking] Modify validation rules to add CRUD operations

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET/issues/1140

This PR:
- Modifies validation rules to add Create, Read, Update and Delete operations to allow for a configurable experience by clients.
- Updates relevant tests.
- This is a breaking change.

### DIFF
--- a/src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
+++ b/src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Linq;
 using System.Reflection;
-using System.Collections;
 using System.Collections.Generic;
 using Microsoft.OpenApi.Exceptions;
 using Microsoft.OpenApi.Properties;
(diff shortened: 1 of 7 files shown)


baywet:
@irvinesunday this is conflicting. Would you mind resolving conflicts before we take a look at it please?

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -86,55 +118,158 @@ public ValidationRuleSet(ValidationRuleSet ruleSet)
...
+        }
+
+        /// <summary>
+        /// Removes a rule from the list of rules.
+        /// </summary>
+        /// <param name="rule">The rule to be removed.</param>
+        /// <returns>true if the rule is successfully removed; otherwise, false.</returns>
+        public bool Remove(ValidationRule rule)
+        {
+            return _rules.Values.SelectMany(x => x).ToList().Remove(rule);
```
baywet, line 230:
```suggestion
            return _rules.Values.Remove(rule);
```
ToList means you're removing from a new list instance, effectively having no effect on the current data.

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -86,55 +118,158 @@ public ValidationRuleSet(ValidationRuleSet ruleSet)
...
+            if (!_rules.ContainsKey(key))
+            {
+                return false;
+            }
+            else
+            {
+                _rules[key].Remove(oldRule);
+                _rules[key].Add(newRule);
+                return true;
+            }
```
baywet, line 194:
```suggestion
            if (_rules.TryGetValue(key, out var currentRules))
            {
                currentRules.Add(newRule);
                return currentRules.Remove(oldRule);
            }
            return false;
```

Note: this implementation doesn't conserve order, but the previous one didn't either, I'm not sure whether it's a requirement?

baywet (changes requested):

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
irvinesunday: I don't believe the order should matter really.

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -86,55 +118,154 @@ public ValidationRuleSet(ValidationRuleSet ruleSet)
...
+        }
+
+        /// <summary>
+        /// Removes a rule from the list of rules.
+        /// </summary>
+        /// <param name="rule">The rule to be removed.</param>
+        /// <returns>true if the rule is successfully removed; otherwise, false.</returns>
+        public bool Remove(ValidationRule rule)
+        {
+            return _rules.Values.Remove((IList<ValidationRule>)rule);
```
baywet, line 226: I'm not sure I understand the reasoning here? AFAIK Values is a new projection of the values in the dictionary (new collection) and modifying this collection won't impact the source dictionary.

irvinesunday: You're right. I have resolved this.

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -217,21 +217,29 @@ public bool Remove(string key, ValidationRule rule)
...
-            return _rules.Values.Remove((IList<ValidationRule>)rule);
+            foreach(var item in _rulesDictionary.Values)
+            {
+                if (item.Remove(rule))
+                {
+                    return true;
+                }
+            }
+
+            return false;
```
baywet:
```suggestion
            return _rulesDictionary.Values.FirstOrDefault(x => x.Remove(rule)) is null ? false : true;
```

baywet (changes requested):

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
irvinesunday: Simplified it further as...
```CSharp
return _rulesDictionary.Values.FirstOrDefault(x => x.Remove(rule)) is not null;
```

baywet (approved):
Thanks for making the changes

PR #1278, opened 2023-06-19
JsonSchema.NET Integration

### LINKED ISSUE #795: Add OpenAPI 3.1 Support
Includes:

- [x] #1046	
- [x] #1039	
- [x] #1042
- [x] #1056 	
- [x] #1077 
- [x] #1138 	
- [x] #1275
- [x] #1274
- [x] #1273
- [x] #1272
- [x] #1271

### LINKED ISSUE #1271: ExclusiveMinimum and ExclusiveMaximum change from type boolean to type number

### LINKED ISSUE #1272: Enable upcasting from V2/V3 to 3.1

### LINKED ISSUE #1273: Map nullable to null type

### LINKED ISSUE #1274: Add support for $schema, $id, $comment, $vocabulary, $dynamicRef, $dynamicAnchor, $recursiveAnchor, $recursiveRef

### LINKED ISSUE #1275: Support the core and validation keywords from the latest JSON schema specification (JSON Schema Specification Draft 2020-12)

MaggieKimani1:
- Resolves integration of the JsonSchema.NET library for complete JSON schema support as part of implementing OpenAPI 3.1.
- Fixes:
  - fixes #795
  - fixes #1271,
  - fixes #1272, 
  - fixes #1273, 
  - fixes #1274,
  - fixes #1275


src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -154,33 +168,37 @@ public void SerializeAsV3(IOpenApiWriter writer)
...
             // If the reference exists but points to other objects, the object is serialized to just that reference.
 
             // schemas
-            writer.WriteOptionalMap(
-                OpenApiConstants.Schemas,
-                Schemas,
-                (w, key, component) =>
+            if (Schemas31 != null && Schemas31.Any())
+            {
+                if (writer is OpenApiYamlWriter)
```
darrelmiller, line 184: I think we need to update the OpenApiJsonWriter to handle writing JsonSchema objects rather than lifting the code into the model objects.

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -154,33 +168,37 @@ public void SerializeAsV3(IOpenApiWriter writer)
...
+
+                    var yamlSchema = serializer.Serialize(yamlNode);
+
+                    writer.WritePropertyName(OpenApiConstants.Schemas);
+                    writer.WriteRaw(yamlSchema);
+                }
+                else
+                {
+                    writer.WritePropertyName(OpenApiConstants.Schemas);
+                    writer.WriteRaw(JsonSerializer.Serialize(Schemas31));
```
darrelmiller, line 199: Let's add a WriteJsonSchema method to the writer interface.

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -336,19 +354,14 @@ private void RenderComponents(IOpenApiWriter writer)
...
+            if (loops.TryGetValue(typeof(JsonSchema), out List<object> schemas))
             {
 
-                writer.WriteOptionalMap(
-                   OpenApiConstants.Schemas,
-                   Schemas,
-                   static (w, key, component) => {
-                       component.SerializeAsV31WithoutReference(w);
-                   });
+                writer.WriteRaw(JsonSerializer.Serialize(schemas));
```
darrelmiller, line 360: If we make the previously suggested change, this change can be reverted. This won't work because it will be missing the "schemas" property.

src/Microsoft.OpenApi/Models/OpenApiDiscriminator.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.OpenApi.Models
     /// <summary>
     /// Discriminator object.
     /// </summary>
-    public class OpenApiDiscriminator : IOpenApiSerializable
+    public class OpenApiDiscriminator : IOpenApiSerializable, IOpenApiExtensible
```
darrelmiller, line 13: Seems weird that this change is in this PR. But maybe this is 3.1 feature.

MaggieKimani1: This should have been a separate PR, that was an oversight on my end.

test/Microsoft.OpenApi.Tests/Models/OpenApiDocumentTests.SerializeAdvancedDocumentAsV2JsonWorks_produceTerseOutput=False.verified.txt
```diff
@@ -50,15 +50,66 @@
         "responses": {
           "200": {
             "description": "pet response",
-            "schema": {"type":"array","items":{"type":"object","required":["id","name"],"properties":{"id":{"type":"integer","format":"int64"},"name":{"type":"string"},"tag":{"type":"string"}}}}
+            "schema": {
```
baywet, line 53: we have an extraneous line here, I believe this is due to the implementation

src/Microsoft.OpenApi/Writers/OpenApiYamlWriter.cs
```diff
@@ -241,7 +243,18 @@ public override void WriteJsonSchema(JsonSchema schema)
                                 .Build();
 
             var yamlSchema = serializer.Serialize(yamlNode);
-            WriteRaw(yamlSchema);
+
+            //remove trailing newlines
+            yamlSchema = yamlSchema.Trim();
```
baywet, line 248: same comment about using a potential serializer option and delegating the concern to the serializer instead of mixing concerns

baywet (changes requested):

src/Microsoft.OpenApi.Readers/SchemaTypeConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using Json.Schema;
+
+namespace Microsoft.OpenApi.Readers
+{
+    internal static class SchemaTypeConverter
+    {
+        internal static SchemaValueType ConvertToSchemaValueType(string value)
+        {
+            return value switch
```
baywet, line 13: are we sure the value is always going to be lowercased or should we normalize it?

src/Microsoft.OpenApi.Readers/V2/OpenApiParameterDeserializer.cs
```diff
@@ -17,7 +19,8 @@ namespace Microsoft.OpenApi.Readers.V2
     /// </summary>
     internal static partial class OpenApiV2Deserializer
     {
-        private static readonly FixedFieldMap<OpenApiParameter> _parameterFixedFields =
+        private static JsonSchemaBuilder _parameterJsonSchemaBuilder;
+        private static FixedFieldMap<OpenApiParameter> _parameterFixedFields =
```
baywet, line 23:
```suggestion
        private static readonly FixedFieldMap<OpenApiParameter> _parameterFixedFields =
```

src/Microsoft.OpenApi.Readers/V3/OpenApiDocumentDeserializer.cs
```diff
@@ -21,10 +21,8 @@ internal static partial class OpenApiV3Deserializer
                 } /* Version is valid field but we already parsed it */
             },
             {"info", (o, n) => o.Info = LoadInfo(n)},
```
baywet, line 23: are those deletions intended?

baywet (changes requested):
Happy to see the progress on this!
There a a number of repeating patterns I started commenting but eventually gave up on because of the occurences:
- startswith is used without a string comparison
- static fields could be readonly
- type inference can be used in the condition directly

MaggieKimani1:
> Happy to see the progress on this! There a a number of repeating patterns I started commenting but eventually gave up on because of the occurences:
> 
> * startswith is used without a string comparison
> * static fields could be readonly
> * type inference can be used in the condition directly

Thanks for pointing these out. Resolving..

src/Microsoft.OpenApi.Readers/V3/OpenApiDocumentDeserializer.cs
MaggieKimani1: Yes. the properties are in 3.1-specific so I moved them to the V3.1 deserializer

src/Microsoft.OpenApi.Readers/SchemaTypeConverter.cs
MaggieKimani1: We can normalize it

baywet: thanks for making the change, you don't need to affect the variable, you can do the normalization straight in the switch

baywet:
@maggieKimani1 there are a bunch of comments from codeQL, would you mind fixing them before I do another review please?

src/Microsoft.OpenApi.Readers/ParseNodes/AnyListFieldMapParameter.cs
```diff
@@ -14,28 +14,28 @@ internal class AnyListFieldMapParameter<T>
         /// Constructor
         /// </summary>
         public AnyListFieldMapParameter(
-            Func<T, IList<OpenApiAny>> propertyGetter,
-            Action<T, IList<OpenApiAny>> propertySetter,
-            Func<T, OpenApiSchema> schemaGetter)
+            Func<T, IList<JsonNode>> propertyGetter,
+            Action<T, IList<JsonNode>> propertySetter,
+            Func<T, JsonSchema> SchemaGetter = null)
```
darrelmiller, line 19:
```suggestion
            Func<T, JsonSchema> schemaGetter = null)
```

darrelmiller: Don't commit this or it will break the rest of the file. Just do the rename.

src/Microsoft.OpenApi.Readers/V31/OpenApiSecurityRequirementDeserializer.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            var mapNode = node.CheckMapNode("security");
+            string description = null;
+            string summary = null;
+
+            var securityRequirement = new OpenApiSecurityRequirement();
+
+            foreach (var property in mapNode)
+            {
+                if (property.Name.Equals("description") || property.Name.Equals("summary"))
```
darrelmiller, line 25: I don't believe this is correct.  SecurityRequirement references are not technically Reference Objects in OAS so you can't have description and summaries along side them.

src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
```diff
@@ -33,8 +33,15 @@
     <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
         <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
     </PropertyGroup>
+    <ItemGroup>
+      <PackageReference Include="JsonSchema.Net" Version="4.1.5" />
+      <PackageReference Include="JsonSchema.Net.OpenApi" Version="1.1.0" />
+      <PackageReference Include="SharpYaml" Version="2.1.0" />
```
darrelmiller, line 39: We should not be pulling in SharpYaml here.  This must be a mistake.

src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
```diff
@@ -33,8 +33,15 @@
...
+    </ItemGroup>
     <ItemGroup>
       <PackageReference Include="System.Text.Json" Version="7.0.2" />
+      <PackageReference Include="Yaml2JsonNode" Version="1.1.1" />
+      <PackageReference Include="YamlDotNet" Version="13.1.0" />
```
darrelmiller, line 44: Apart from the fact that these two references cannot be in this project, we can't have a reference to both SharpYaml and YamlDotNet. They do the same thing.  If Yaml2JsonNode has take a dependency on YamlDotNet, then we are going to have to rewrite that code.

src/Microsoft.OpenApi/Models/OpenApiConstants.cs
```diff
@@ -595,6 +595,16 @@ public static class OpenApiConstants
         /// </summary>
         public static readonly Uri defaultUrl = new Uri("http://localhost/");
 
+        /// <summary>
+        /// Field: V3 JsonSchema Reference Uri
+        /// </summary>
+        public const string v3ReferenceUri = "https://everything.json/components/schemas/";
```
darrelmiller, line 601: I feel like we should have a more appropriate default base URL for here rather than something that is based on the library we are taking a dependency on.  Not a big deal, but something to think about.

src/Microsoft.OpenApi/Models/OpenApiParameter.cs
```diff
@@ -103,9 +106,13 @@ public virtual bool Explode
         public virtual bool AllowReserved { get; set; }
 
         /// <summary>
-        /// The schema defining the type used for the parameter.
+        /// The schema defining the type used for the request body.
```
darrelmiller, line 109: This comment change is wrong.

src/Microsoft.OpenApi/Models/References/OpenApiHeaderReference.cs
```diff
@@ -72,7 +73,7 @@ public override string Description
         public override bool AllowEmptyValue { get => Target.AllowEmptyValue; set => Target.AllowEmptyValue = value; }
 
         /// <inheritdoc/>
-        public override OpenApiSchema Schema { get => Target.Schema; set => Target.Schema = value; }
+        public override JsonSchema Schema { get => _schema; set => _schema = value; }
```
darrelmiller, line 76: I think this should still delegate to Target.Schema.  We should probably make _schema in the base class private and not protected.

src/Microsoft.OpenApi/Services/OpenApiReferenceResolver.cs
```diff
@@ -4,7 +4,9 @@
 using System;
```
darrelmiller, line 4: This class should be deleted. We don't need it anymore.

src/Microsoft.OpenApi/Validations/ValidationRule.cs
```diff
@@ -29,7 +29,7 @@ public abstract class ValidationRule
     /// Class containing validation rule logic for <see cref="IOpenApiElement"/>.
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    public class ValidationRule<T> : ValidationRule where T : IOpenApiElement
+    public class ValidationRule<T> : ValidationRule
```
darrelmiller, line 32: Ugh. This is because JsonSchema doesn't implement this interface.  Hmm, that seems like it is going to cause us problems.

src/Microsoft.OpenApi/Writers/OpenApiWriterBase.cs
```diff
@@ -4,9 +4,16 @@
...
+using System.Text.Json;
+using Json.Schema;
+using Json.Schema.OpenApi;
+using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Exceptions;
+using Microsoft.OpenApi.Extensions;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Properties;
+using Microsoft.OpenApi.Services;
+using YamlDotNet.Serialization.ObjectGraphVisitors;
```
darrelmiller, line 16: We can't take this dependency. We are going to have to write the serialization of JsonSchema manually.

src/Microsoft.OpenApi/Writers/OpenApiWriterExtensions.cs
```diff
@@ -181,25 +180,6 @@ public static void WriteProperty<T>(this IOpenApiWriter writer, string name, T v
...
         }
 
-        /// <summary>
-        /// Write the optional of collection string.
-        /// </summary>
-        /// <param name="writer">The Open API writer.</param>
-        /// <param name="name">The property name.</param>
-        /// <param name="elements">The collection values.</param>
-        /// <param name="action">The collection string writer action.</param>
-        public static void WriteOptionalCollection(
```
darrelmiller, line 191: I am curious as to how we were able to remove this.  Isn't Tags an optional collection of strings?

darrelmiller (changes requested):
It's looking good.  I just added a few comments that we should address.  Specifically those related to dependencies on the YAML libraries.

src/Microsoft.OpenApi.Readers/V31/OpenApiSecurityRequirementDeserializer.cs
darrelmiller: Same for Tags.  We cannot override descriptions for these references.

src/Microsoft.OpenApi/Writers/OpenApiWriterExtensions.cs
MaggieKimani1: Actually Tags uses the generic `WriteOptionalCollection<T>` method to be written out. I think the thought process at the time was we could reuse the generic method to write out all types. Let me add it back.

MaggieKimani1: Tags are defined as an `IList<OpenApiTag>` hence they use the generic `WriteOptionalCollection<T>` method during serialization. However, security schemes and enums are collections of type string so I can add the method back.

src/Microsoft.OpenApi/Services/OpenApiReferenceResolver.cs
MaggieKimani1: We've refactored it and are using it to resolve JSON schema references when visiting a JSON schema instance

darrelmiller: Defer resolution to future PR.

darrelmiller (approved):

PR #1284, opened 2023-06-27
Fix hardcoded non-portable path separators

mikekistler:
This PR fixes a bunch of hard-coded "\" separators in the tests that caused many failures when running on MacOS (maybe also on Linux?).

There is still one test failing on my MacOS system, OpenApiWorkspacesShouldNormalizeDocumentLocations. It looks like the value of `Environment.CurrentDirectory` is returning a different value on my system -- it adds "/net7.0" where the test appears not to expect this.


andrueastman (approved):

baywet (approved):
Thanks for the contribution!

Issue #1103, opened 2022-12-06
Implement $ref for not schema objects as Proxy objects


darrelmiller:
Instead of using the ReferenceObject as a placeholder and then doing a second pass to resolve the reference objects to real objects, we should create derived objects e.g. "ParameterReference" that will delegate to a target object if the property in the ParameterReference is not set.  This will enable setting Summary and Description properties alongside the $ref.

See here for example https://github.com/microsoft/OpenAPI.NET/blob/dm/ProxyReference/src/Microsoft.OpenApi/Models/OpenApiSchemaReference.cs  however, we will not do this on OpenAPISchema objects as they handle $ref in a native JSON Schema way.  We will do this for all other types.

PR #1290, opened 2023-07-10
Creates derived reference objects for on-demand reference resolution 

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET/issues/1103

This PR:
- Creates reference objects that can be used to delegate reference resolution on-demand. This will also help in overriding the summary and description property values of the referenceable objects.
- Adds the relevant unit tests to validate the above.


src/Microsoft.OpenApi/Models/References/OpenApiExampleReference.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                Id = referenceId,
+                HostDocument = hostDocument,
+                Type = ReferenceType.Example,
+                ExternalResource = externalResource
+            };
+        }
+
+        /// <inheritdoc/>
+        public override string Description { get => _description ?? Target.Description; set => _description = value; }
```
baywet, line 63: strings should be tested with isnullorempty instead of null coalesce to guard against empty strings when NRT is enabled.

baywet (changes requested):

baywet:
Also @irvinesunday you should have a look at the bugs reported by sonarcloud

src/Microsoft.OpenApi/Models/References/OpenApiHeaderReference.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                Id = referenceId,
+                HostDocument = hostDocument,
+                Type = ReferenceType.Header,
+                ExternalResource = externalResource
+            };
+        }
+
+        /// <inheritdoc/>
+        public override string Description { get => _description ?? Target.Description; set => _description = value; }
```
baywet, line 59: still contains null coalesce for strings

irvinesunday: I had missed quite a handful. I have resolved them.

irvinesunday:
@baywet are there other suggestions you had for this PR?

baywet (approved):

PR #1293, opened 2023-07-18
parse server's url with variables when convert openapi spec 3.0 to 2.0

TinaLiu02:
Applied the suggested changes and the test results as below.
OpenAPI Workbench: The URL in server object supports Server Variables in 3.0, now it has been converted to 2.0 correctly. Please refer below screenshot.
image

Unit Test Results:
image


baywet (changes requested):
Thanks for the contribution, here is a suggestion

TinaLiu02:
Thanks. Applied the suggested changes.

baywet (approved):
Thanks for making the changes! @MaggieKimani1 and @millicentachieng to review and merge

MaggieKimani1 (approved):

PR #1316, opened 2023-08-21
Fix missing response schema bug

### LINKED ISSUE #1315: Incomplete OpenApiResponse
**Describe the bug**
OpenApiResponse object doesn't include the schema that is included in the swagger file.

        "responses": {
          "200": {
            "description": "OK",
            "schema": {
              "format": "binary",
              "description": "The content of the file.",
              "type": "string",
              "x-ms-summary": "File Content"
            }
          },
          "default": {
            "description": "Operation Failed."
          }
        }, ...

**To Reproduce**
``` C#
string swaggerFile = @"https://raw.githubusercontent.com/microsoft/Power-Fx/main/src/tests/Microsoft.PowerFx.Connectors.Tests/Swagger/AzureBlobStorage.json";
using HttpClient client = new HttpClient();
HttpResponseMessage webResponse = await client.GetAsync(swaggerFile).ConfigureAwait(false);
string swagger = await webResponse.Content.ReadAsStringAsync().ConfigureAwait(false);	
OpenApiDocument doc = new OpenApiStringReader().Read(swagger, out OpenApiDiagnostic diag);
OpenApiOperation op = doc.Paths["/{connectionId}/datasets/default/GetFileContentByPath"].Operations[OperationType.Get];
OpenApiResponse response = op.Responses["200"];	
```
Look at the response object and compare it with the swagger file
All fields/props are either null or empty.

**Expected behavior**
Have a way to access the schema, via Content or any other means.

**Screenshots/Code Snippets**
See code snippet

**Additional context**
It's likely related to "Produces" item which is not defined, neither locally, nor globally.
Per Swagger/OpenAPI spec, in Response object, schema SHOULD be accompanied by a relevant produces mime-type.
... but it's not mandatory.

MaggieKimani1:
Fixes #1315

### DIFF
--- a/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
+++ b/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <TargetFrameworks>netstandard2.0</TargetFrameworks>
+        <LangVersion>9.0</LangVersion>
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
         <PackageIconUrl>http://go.microsoft.com/fwlink/?LinkID=288890</PackageIconUrl>
         <PackageProjectUrl>https://github.com/Microsoft/OpenAPI.NET</PackageProjectUrl>
--- a/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
+++ b/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <TargetFrameworks>netstandard2.0</TargetFrameworks>
-        <LangVersion>9.0</LangVersion>
+        <LangVersion>latest</LangVersion>
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
         <PackageIconUrl>http://go.microsoft.com/fwlink/?LinkID=288890</PackageIconUrl>
         <PackageProjectUrl>https://github.com/Microsoft/OpenAPI.NET</PackageProjectUrl>
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
@@ -92,12 +92,12 @@ namespace Microsoft.OpenApi.Readers.V2
                 foreach (var produce in produces)
                 {
 
-                    if (response.Content.ContainsKey(produce) && response.Content[produce] != null)
+                    if (response.Content.TryGetValue(produce, out var produceValue)) 
                     {
                         if (schema != null)
                         {
-                            response.Content[produce].Schema = schema;
-                            ProcessAnyFields(mapNode, response.Content[produce], _mediaTypeAnyFields);
+                            produceValue.Schema = schema;
+                            ProcessAnyFields(mapNode, produceValue, _mediaTypeAnyFields);
                         }
                     }
                     else
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
@@ -73,34 +73,47 @@ namespace Microsoft.OpenApi.Readers.V2
             }
 
             var produces = context.GetFromTempStorage<List<string>>(TempStorageKeys.OperationProduces)
-                ?? new List<string> { "application/octet-stream" };
+                ?? context.GetFromTempStorage<List<string>>(TempStorageKeys.GlobalProduces);
 
-            var schema = context.GetFromTempStorage<OpenApiSchema>(TempStorageKeys.ResponseSchema, response);
-
-            foreach (var produce in produces)
+            if (produces != null)
             {
+                var schema = context.GetFromTempStorage<OpenApiSchema>(TempStorageKeys.ResponseSchema, response);
 
-                if (response.Content.ContainsKey(produce) && response.Content[produce] != null)
-                {
-                    if (schema != null)
-                    {
-                        response.Content[produce].Schema = schema;
-                        ProcessAnyFields(mapNode, response.Content[produce], _mediaTypeAnyFields);
-                    }
-                }
-                else
+                if (produces.Count == 0 && schema != null)
                 {
                     var mediaType = new OpenApiMediaType
                     {
                         Schema = schema
                     };
 
-                    response.Content.Add(produce, mediaType);
+                    response.Content.Add(string.Empty, mediaType);
                 }
-            }
 
-            context.SetTempStorage(TempStorageKeys.ResponseSchema, null, response);
-            context.SetTempStorage(TempStorageKeys.ResponseProducesSet, true, response);
+                foreach (var produce in produces)
+                {
+
+                    if (response.Content.ContainsKey(produce) && response.Content[produce] != null)
+                    {
+                        if (schema != null)
+                        {
+                            response.Content[produce].Schema = schema;
+                            ProcessAnyFields(mapNode, response.Content[produce], _mediaTypeAnyFields);
+                        }
+                    }
+                    else
+                    {
+                        var mediaType = new OpenApiMediaType
+                        {
+                            Schema = schema
+                        };
+
+                        response.Content.Add(produce, mediaType);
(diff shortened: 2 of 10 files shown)


baywet (dismissed):

darrelmiller:
Do we need to make the same fix for empty Consumes array?

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
```diff
@@ -73,47 +73,34 @@ private static void ProcessProduces(MapNode mapNode, OpenApiResponse response, P
...
 
-            if (produces != null)
+            var schema = context.GetFromTempStorage<OpenApiSchema>(TempStorageKeys.ResponseSchema, response);
+
+            foreach (var produce in produces)
             {
-                var schema = context.GetFromTempStorage<OpenApiSchema>(TempStorageKeys.ResponseSchema, response);
 
-                if (produces.Count == 0 && schema != null)
+                if (response.Content.ContainsKey(produce) && response.Content[produce] != null)
```
baywet: TryGetValue instead?

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
```diff
@@ -73,47 +73,34 @@ private static void ProcessProduces(MapNode mapNode, OpenApiResponse response, P
             }
 
             var produces = context.GetFromTempStorage<List<string>>(TempStorageKeys.OperationProduces)
```
baywet, line 75: are we sure the returned value from the first part is going to be null and never an empty list/collection?

MaggieKimani1: Well that depends. If produces is missing from the file then it will be null, else its going to return an empty list if the value of produces is an empty array

MaggieKimani1: Oh okay I get your question. 
As per my understanding, I don't think there would be any downstream consequence because an empty value in the first part means we'll still set the value of the response schema in the Media type object.
Maybe @darrelmiller  has more thoughts on this?

darrelmiller: There are consequences to this because if the global produces defines `application/json` as the default response, then if produces at the operation level is an empty array, we are going to ignore the global produces and return `application/octet-stream`.  It is arguable as to what the correct behaviour should be in this case.  Does "empty produces" mean "assume default content type" or does it mean "equivalent to null".  

Sadly, I think the reason for the empty array is due to naïve serializers and empty array is effectively null.  This means that we should respect the global produces. I think you need to consider an empty OperationProduces as if it was a null.

darrelmiller (approved):

src/Microsoft.OpenApi.Readers/V2/OpenApiV2Deserializer.cs
```diff
@@ -29,10 +29,11 @@ internal static partial class OpenApiV2Deserializer
                 return;
             }
 
-            foreach (var propertyNode in mapNode)
+            var allFields = fixedFieldMap.Keys.Union(mapNode.Select(x => x.Name));
```
baywet, line 32:
```suggestion
            var allFields = fixedFieldMap.Keys.Union(mapNode.Select(static x => x.Name));
```

src/Microsoft.OpenApi.Readers/V2/OpenApiV2Deserializer.cs
```diff
@@ -29,10 +29,11 @@ internal static partial class OpenApiV2Deserializer
...
+            foreach (var propertyNode in allFields)
             {
-                propertyNode.ParseField(domainObject, fixedFieldMap, patternFieldMap);
-                requiredFields?.Remove(propertyNode.Name);
+                mapNode[propertyNode]?.ParseField(domainObject, fixedFieldMap, patternFieldMap);
```
baywet, line 35: try get instead with a condition?

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
baywet: with the latest implementation (using null coallesce) and empty but non null return from one of the first two members would not produce application/octet-stream, that probably needs to be fixed

baywet (changes requested):

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
MaggieKimani1: We're treating an empty produces value as null so the empty value won't be set at all here https://github.com/microsoft/OpenAPI.NET/blob/5b080619be23f2b9e2e3995f53393f2263e5a4e3/src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs#L70-L71.

src/Microsoft.OpenApi.Readers/V2/OpenApiV2Deserializer.cs
MaggieKimani1: MapNode implements an IEnumerable <T> hence doesn't expose a TryGetValue() method

baywet: ah, I can see the implementation already does a TryGetValue https://github.com/microsoft/OpenAPI.NET/blob/5b080619be23f2b9e2e3995f53393f2263e5a4e3/src/Microsoft.OpenApi.Readers/ParseNodes/MapNode.cs#L46

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
baywet: per Darrel's comment I believe we need to consider empty arrays as null, so an implementation like this one would probably be more suitable.

```csharp
var produces = context.GetFromTempStorage<List<string>>(TempStorageKeys.OperationProduces) is {} opProdcudes && opProduces.Any() ? opProduces :
                    context.GetFromTempStorage<List<string>>(TempStorageKeys.GlobalProduces) is {} globProduces && globProduces.Any() ? globProduces:
                    new List<string> { "application/octet-stream" };
```

src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <TargetFrameworks>netstandard2.0</TargetFrameworks>
+        <LangVersion>9.0</LangVersion>
```
baywet, line 4: why are we restricting the lang version here?

MaggieKimani1: The `static` anonymous function is unavailable in C# 7.3 which is what the project was using.

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
MaggieKimani1: We do treat an empty produces array as null when setting it to our temporary storage [here](https://github.com/microsoft/OpenAPI.NET/blob/5b080619be23f2b9e2e3995f53393f2263e5a4e3/src/Microsoft.OpenApi.Readers/V2/OpenApiOperationDeserializer.cs#L70-L71) as per [this comment](https://github.com/microsoft/OpenAPI.NET/pull/1316/files#r1312122317). We only set the value if the array is non-empty.
So at this point of retrieval from the temp storage, we won't need to check for an empty array since in that case the value returned will always be null hence default to octet-stream if both the first and second values are null.

src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
baywet: why not set this to latest instead?

src/Microsoft.OpenApi.Readers/V2/OpenApiResponseDeserializer.cs
baywet: ah ok, I was reading that stack the other way around for some reason. Thanks for clarifying. Resolving.

src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
MaggieKimani1: Resolving..

baywet (approved):

PR #1333, opened 2023-09-06
moves all microsoft extensions to a single place

baywet:
related https://github.com/microsoft/kiota/issues/3066


andrueastman (approved):
👍🏼

PR #1343, opened 2023-09-14
Fixes null reference exception. Closes #1342

### LINKED ISSUE #1342: Hidi: Filtering OpenAPI spec by Postman collection fails
**Describe the bug**

Filtering OpenAPI spec by a Postman collection fails with the following error:

```text
Could not transform the document, reason: Object reference not set to an instance of an object.
```

**To Reproduce**

- Download Graph openapi.yaml
- Create a Postman collection with some endpoint
- `hidi transform -d openapi.yaml -c collection.json --co`

**Expected behavior**

Filtered collection

**Screenshots/Code Snippets**
image

**Additional context**
Error is caused by a null reference here:

https://github.com/microsoft/OpenAPI.NET/blob/2984d30ebcdb82b332af644a8d0e806748ade5a5/src/Microsoft.OpenApi.Hidi/OpenApiService.cs#L449

This stems from initializing the requestUrls as null here: 

https://github.com/microsoft/OpenAPI.NET/blob/2984d30ebcdb82b332af644a8d0e806748ade5a5/src/Microsoft.OpenApi.Hidi/OpenApiService.cs#L147

To fix this, we'd need to instantiate requestUrls as an empty Dictionary. If you're okay with this proposal I'd be happy to submit a PR.

waldekmastykarz:
Fixes null reference exception. Closes #1342

### DIFF
--- a/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
+++ b/src/Microsoft.OpenApi.Hidi/OpenApiService.cs
@@ -151,7 +151,7 @@ namespace Microsoft.OpenApi.Hidi
             }
             else if (postmanCollection != null)
             {
-                requestUrls = EnumerateJsonDocument(postmanCollection.RootElement, requestUrls);
+                requestUrls = EnumerateJsonDocument(postmanCollection.RootElement, new());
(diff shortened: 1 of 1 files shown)


baywet (approved):
Thanks for the fix!

PR #1344, opened 2023-09-14
enables NRT, analysis all and treat warning as errors on hidi

baywet:
- - enables NRT for main hidi project
- - enables all mode analysis

### DIFF
--- a/.azure-pipelines/ci-build.yml
+++ b/.azure-pipelines/ci-build.yml
@@ -216,7 +216,7 @@ stages:
         displayName: publish Hidi as executable
         inputs:
           command: 'publish'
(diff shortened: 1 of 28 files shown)


andrueastman (commented):
Looks great @baywet. 

There is a trimming warning that is causing the release build to fail due to https://github.com/dotnet/runtime/discussions/59230. I believe we should probably suppress/fix before the merge.

https://github.com/microsoft/OpenAPI.NET/blob/b6cf6198fef5a5d0a298694eb07bc6242eac639b/src/Microsoft.OpenApi.Hidi/Utilities/SettingsUtilities.cs#L28

baywet:
@andrueastman, thanks. I went the other way around by disabling trimming since fixing it would represent significant work until the source generators are available, and suppressing it means broken functionality.

andrueastman (approved):

PR #1373, opened 2023-10-02
use negation pattern

SimonCropp:


baywet (approved):

baywet (changes requested):
@simoncropp this is now conflicting due to other merges. Would you mind rebasing please?

baywet (approved):

PR #1372, opened 2023-10-02
inline some out variables

SimonCropp:

PR #1374, opened 2023-10-02
use return from TryGetValue

SimonCropp:


baywet:
Closing this one since this is the recommendation I made in #1372 and we'll end up with a conflict

PR #1380, opened 2023-10-02
remove some duplicate dictionary lookups

SimonCropp:


src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -59,12 +59,17 @@ public bool HasOperations(string label)
         {
             Utils.CheckArgumentNullOrEmpty(label, nameof(label));
 
-            if (!(PathItems?.ContainsKey(label) ?? false))
+            if (PathItems == null)
             {
                 return false;
             }
```
baywet, line 66:
```suggestion
```

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -59,12 +59,17 @@ public bool HasOperations(string label)
...
 
-            return PathItems[label].Operations?.Any() ?? false;
+            if (PathItems.TryGetValue(label, out var item))
+            {
+                return item.Operations?.Any() ?? false;
+            }
+
+            return false;
```
baywet, line 72:
```suggestion
            return PathItems is not null && PathItems.TryGetValue(label, out var item) && item.Operations is not null && item.Operations.Any();
```

baywet (changes requested):

baywet (approved):

PR #1382, opened 2023-10-02
remove redundant casts

SimonCropp:


src/Microsoft.OpenApi.Readers/ParseNodes/OpenApiAnyConverter.cs
```diff
@@ -50,16 +50,16 @@ public static IOpenApiAny GetSpecificOpenApiAny(IOpenApiAny openApiAny, OpenApiS
                 return newObject;
             }
 
-            if (!(openApiAny is OpenApiString))
+            if (!(openApiAny is OpenApiString apiString))
```
baywet, line 53:
```suggestion
            if (openApiAny is not OpenApiString apiString)
```

src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
```diff
@@ -75,16 +75,14 @@ public static bool IsEmailAddress(this string input)
                 }
 
                 // If value is not a string and also not an object, there is a data mismatch.
-                if (!(value is OpenApiObject))
+                if (!(value is OpenApiObject anyObject))
```
baywet, line 78:
```suggestion
                if (value is not OpenApiObject anyObject)
```

src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
```diff
@@ -115,16 +113,14 @@ public static bool IsEmailAddress(this string input)
                 }
 
                 // If value is not a string and also not an array, there is a data mismatch.
-                if (!(value is OpenApiArray))
+                if (!(value is OpenApiArray anyArray))
```
baywet, line 116:
```suggestion
                if (value is not OpenApiArray anyArray)
```

baywet (changes requested):

baywet (approved):
Thanks for the contribution!

PR #1388, opened 2023-10-03
cleanup exception handling

SimonCropp:
* leverage CallerArgumentExpressionAttribute
 * there were two exception handling helpers. `Error` and `Utils`. Consolidated on `Utils`. Happy to rename if u want
 * the string format and CultureInfo code in Error was redundant, since the only place it was used in was not doing any culture sensitive formats.


src/Microsoft.OpenApi/CallerArgumentExpressionAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+[ExcludeFromCodeCoverage]
+[DebuggerNonUserCode]
+[AttributeUsage(AttributeTargets.Parameter)]
+sealed class CallerArgumentExpressionAttribute :
+    Attribute
+{
+    public CallerArgumentExpressionAttribute(string parameterName) =>
+        ParameterName = parameterName;
+
+    public string ParameterName { get; }
```
baywet, line 15: isn't the value here "value" in case of netstandard 2.0?

src/Microsoft.OpenApi/CallerArgumentExpressionAttribute.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Diagnostics;
```
baywet, line 1: should the whole file be wrapped in conditional compilation directives? (#if !NETCOREAPP3_1_OR_GREATER)
[reference](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerargumentexpressionattribute?view=net-7.0)

baywet (changes requested):

src/Microsoft.OpenApi/CallerArgumentExpressionAttribute.cs
SimonCropp: i have just kept the same convention as the actual implementation https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerargumentexpressionattribute.-ctor?view=net-7.0#system-runtime-compilerservices-callerargumentexpressionattribute-ctor(system-string)

src/Microsoft.OpenApi/CallerArgumentExpressionAttribute.cs
SimonCropp: given u target netstandard2, it is not really valuable.

If it was my project i would leverage a polyfill. eg

 * https://github.com/SimonCropp/Polyfill
 * https://github.com/Sergio0694/PolySharp
 * https://github.com/Tyrrrz/PolyShim

Then u get CallerArgumentExpressionAttribute for free as well as a significant number of other newer apis+c# features

SimonCropp:
conflict resolved

src/Microsoft.OpenApi/CallerArgumentExpressionAttribute.cs
baywet: My reasoning being: if we start enabling multi-targeting to benefit from newer features, this won't conflict with what the framework provides if conditional compilation directives are added.

src/Microsoft.OpenApi/CallerArgumentExpressionAttribute.cs
baywet: I guess my confusion is how does this custom attribute get the actual parameter name since it doesn't have the support of the framework and can't take advantage of compiler capabilities? (due to the version being targeted)

SimonCropp: > can't take advantage of compiler capabilities

thats not correct. compiler capabilities are determined by the c# version being targeted. some times u need the marker attributes for some features

src/Microsoft.OpenApi/CallerArgumentExpressionAttribute.cs
SimonCropp: but if u use a polyfill you dont need any compilation directives. or at least significantly fewer compilation directives

src/Microsoft.OpenApi/CallerArgumentExpressionAttribute.cs
baywet: oh this is only compile time sugar! I thought this relied on some IL feature. Thanks for clarifying.

baywet (approved):

baywet:
@simoncropp this one is conflicting again, would you mind rebasing please?

SimonCropp:
@baywet done

PR #1416, opened 2023-10-08
use some pattern matching

SimonCropp:


src/Microsoft.OpenApi.Readers/Services/OpenApiRemoteReferenceCollector.cs
```diff
@@ -39,14 +39,11 @@ public override void Visit(IOpenApiReferenceable referenceable)
         /// </summary>
         private void AddReference(OpenApiReference reference)
         {
-            if (reference != null)
+            if (reference is {IsExternal: true})
             {
-                if (reference.IsExternal)
+                if (!_references.ContainsKey(reference.ExternalResource))
```
baywet, line 44: collapse this if into the previous one

baywet (changes requested):

src/Microsoft.OpenApi.Readers/Services/OpenApiRemoteReferenceCollector.cs
```diff
@@ -39,15 +39,10 @@ public override void Visit(IOpenApiReferenceable referenceable)
...
+                !_references.ContainsKey(reference.ExternalResource))
             {
-                if (reference.IsExternal)
-                {
-                    if (!_references.ContainsKey(reference.ExternalResource))
-                    {
-                        _references.Add(reference.ExternalResource, reference);
-                    }
-                }
+                _references.Add(reference.ExternalResource, reference);
```
baywet, line 45: why not use TryAdd here instead of looking up the key ourselves?

baywet (changes requested):

src/Microsoft.OpenApi.Readers/Services/OpenApiRemoteReferenceCollector.cs
```diff
@@ -39,15 +39,10 @@ public override void Visit(IOpenApiReferenceable referenceable)
         /// </summary>
         private void AddReference(OpenApiReference reference)
         {
-            if (reference != null)
+            if (reference is {IsExternal: true} &&
+                !_references.ContainsKey(reference.ExternalResource))
```
baywet, line 43:
```suggestion
            if (reference is {IsExternal: true})
```

src/Microsoft.OpenApi.Readers/Services/OpenApiRemoteReferenceCollector.cs
baywet:
```suggestion
                _references.TryAdd(reference.ExternalResource, reference);
```

baywet (approved):

baywet (approved):

PR #1423, opened 2023-10-09
Use lookup collections

larsbj1988:
Changed collection types to either Dictionary or HashSet from arrays where lookups are used.
Keeping single collection definition over yaml control characters.

### DIFF
--- a/src/Microsoft.OpenApi/Writers/SpecialCharacterStringExtensions.cs
+++ b/src/Microsoft.OpenApi/Writers/SpecialCharacterStringExtensions.cs
@@ -65,7 +65,7 @@ namespace Microsoft.OpenApi.Writers
 
         // Double-quoted strings are needed for these non-printable control characters.
         // http://www.yaml.org/spec/1.2/spec.html#style/flow/double-quoted
-        private static readonly Dictionary<char, string> _yamlControlCharacterReplacements = new()
+        private static readonly Dictionary<char, string> _yamlControlCharacterCharReplacements = new()
         {
             {'\0', "\\0"},
             {'\x01', "\\x01"},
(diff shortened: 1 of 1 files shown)


larsbj1988:
> @microsoft-github-policy-service agree

src/Microsoft.OpenApi/Writers/SpecialCharacterStringExtensions.cs
```diff
@@ -134,39 +135,11 @@ internal static string GetYamlCompatibleString(this string input)
...
-                input = input.Replace("\x1a", "\\x1a");
-                input = input.Replace("\x1b", "\\x1b");
-                input = input.Replace("\x1c", "\\x1c");
-                input = input.Replace("\x1d", "\\x1d");
-                input = input.Replace("\x1e", "\\x1e");
-                input = input.Replace("\x1f", "\\x1f");
-
+                foreach (var replacement in _yamlControlCharacterReplacements)
+                {
+                    input = input.Replace(replacement.Key.ToString(), replacement.Value);
```
baywet, line 140: could we use a second dictionary or something similar to avoid the string alloc caused by the Key.ToString()?

baywet (changes requested):
Thanks for the contribution, since this is a hot path, we should probably try to avoid allocs as much as possible.

src/Microsoft.OpenApi/Writers/SpecialCharacterStringExtensions.cs
larsbj1988: Thank you for your suggestion. I added an additional commit :)

andrueastman (approved):

baywet (approved):
Thanks for making the changes!

PR #1443, opened 2023-10-27
Bump up Hidi version

irvinesunday:


MaggieKimani1:
Maybe bump the lib versions as well so that the recent changes go out in one release?

MaggieKimani1 (approved):

PR #1446, opened 2023-10-31
Use Load/LoadAsync pattern to load OpenApi documents

### LINKED ISSUE #1434: Change the way OpenAPI documents are loaded in v2
Currently we have a set of OpenApiReaders to load OpenAPI documents.  We have OpenApiStreamReader, OpenApiStringReader, OpenApiTextReader and OpenApiYamlDocumentReader. Each one of these a smaller number of methods. e.g. Read, ReadAsync, ReadFragment.

To use this object you must create the reader and then call a Read method.  However, you generally only call a single Read method because it loads in the entire source content.  Normally this pattern of using a Reader object is for when a client application wants to read in portions of a document and perform actions between those reads.  That is not how we process documents.

This issue proposes that we move to the Load/LoadAsync/Parse pattern https://learn.microsoft.com/en-us/dotnet/api/system.xml.linq.xdocument.load?view=net-7.0 which is more consistent with our usage.

```[tasklist]
### Tasks
- [x] Define an interface for the format providers i.e YAML and JSON
- [x] Create a registry for managing the format providers
- [x] Implement the interface for each format provider i.e OpenApiJsonReader(lives in core lib) and OpenApiYamlReader(in readers lib)
- [x] Move loading code to the core lib
- [x] Create an OpenApiModelFactory class that abstracts the loading logic
- [x] Add static load methods on the OpenApiDocument class that call the factory
- [ ] https://github.com/microsoft/OpenAPI.NET/issues/1447
```

MaggieKimani1:
Fixes #1434

### DIFF
--- a/src/Microsoft.OpenApi.Readers/Document.cs
+++ b/src/Microsoft.OpenApi.Readers/Document.cs
@@ -1,0 +1,88 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Readers
+{
+    /// <summary>
+    /// Loads an OpenApiDocument instance through Load/LoadAsync/Parse pattern
+    /// </summary>
+    internal static class Document
+    {
+        /// <summary>
+        /// Loads an OpenApiDocument from a file stream
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <param name="diagnostic"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static OpenApiDocument Load(Stream stream, out OpenApiDiagnostic diagnostic, OpenApiReaderSettings settings = null)
+        {
+            return new OpenApiStreamReader(settings).Read(stream, out diagnostic);
+        }
+
+        /// <summary>
+        /// Loads an OpenApiDocument from a TextReader
+        /// </summary>
+        /// <param name="reader"></param>
+        /// <param name="diagnostic"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static OpenApiDocument Load(TextReader reader, out OpenApiDiagnostic diagnostic, OpenApiReaderSettings settings = null)
+        {
+            return new OpenApiTextReaderReader(settings).Read(reader, out diagnostic);
+        }
+
+        /// <summary>
+        /// Loads an OpenApiDocument from a string input
+        /// </summary>
+        /// <param name="input"></param>
+        /// <param name="diagnostic"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static OpenApiDocument Load(string input, out OpenApiDiagnostic diagnostic, OpenApiReaderSettings settings = null)
+        {
+            return new OpenApiStringReader(settings).Read(input, out diagnostic);
+        }
+
+        /// <summary>
+        /// Loads an OpenApiDocument from a string input
+        /// </summary>
+        /// <param name="input"></param>
+        /// <param name="diagnostic"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static OpenApiDocument Parse(string input, out OpenApiDiagnostic diagnostic, OpenApiReaderSettings settings = null)
+        {
+            return new OpenApiStringReader(settings).Read(input, out diagnostic);
+        }
+
+        /// <summary>
+        /// Loads an OpenApiDocument asynchronously from a TextReader
+        /// </summary>
+        /// <param name="reader"></param>
+        /// <param name="cancellationToken"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static Task<ReadResult> LoadAsync(TextReader reader, CancellationToken cancellationToken = default, OpenApiReaderSettings settings = null)
+        {
+            return new OpenApiTextReaderReader(settings).ReadAsync(reader, cancellationToken);
+        }
+
+        /// <summary>
+        /// Loads an OpenApiDocument from a file stream
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <param name="cancellationToken"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static async Task<ReadResult> LoadAsync(Stream stream, CancellationToken cancellationToken = default, OpenApiReaderSettings settings = null)
+        {
+            return await new OpenApiStreamReader(settings).ReadAsync(stream, cancellationToken);
+        }
+    }
+}
--- a/src/Microsoft.OpenApi.Readers/Document.cs
+++ b/src/Microsoft.OpenApi.Readers/Document.cs
@@ -11,7 +11,7 @@ namespace Microsoft.OpenApi.Readers
     /// <summary>
     /// Loads an OpenApiDocument instance through Load/LoadAsync/Parse pattern
     /// </summary>
-    internal static class Document
+    public static class Document
     {
         /// <summary>
         /// Loads an OpenApiDocument from a file stream
@@ -68,7 +68,7 @@ namespace Microsoft.OpenApi.Readers
         /// <param name="cancellationToken"></param>
         /// <param name="settings"></param>
         /// <returns></returns>
-        public static Task<ReadResult> LoadAsync(TextReader reader, CancellationToken cancellationToken = default, OpenApiReaderSettings settings = null)
+        public static Task<ReadResult> LoadAsync(TextReader reader, OpenApiReaderSettings settings = null, CancellationToken cancellationToken = default)
         {
             return new OpenApiTextReaderReader(settings).ReadAsync(reader, cancellationToken);
         }
@@ -80,7 +80,7 @@ namespace Microsoft.OpenApi.Readers
         /// <param name="cancellationToken"></param>
         /// <param name="settings"></param>
         /// <returns></returns>
-        public static async Task<ReadResult> LoadAsync(Stream stream, CancellationToken cancellationToken = default, OpenApiReaderSettings settings = null)
+        public static async Task<ReadResult> LoadAsync(Stream stream, OpenApiReaderSettings settings = null, CancellationToken cancellationToken = default)
         {
             return await new OpenApiStreamReader(settings).ReadAsync(stream, cancellationToken);
         }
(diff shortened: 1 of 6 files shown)


src/Microsoft.OpenApi.Readers/Document.cs
```diff
@@ -0,0 +1,88 @@
...
+    internal static class Document
+    {
+        /// <summary>
+        /// Loads an OpenApiDocument from a file stream
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <param name="diagnostic"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static OpenApiDocument Load(Stream stream, out OpenApiDiagnostic diagnostic, OpenApiReaderSettings settings = null)
```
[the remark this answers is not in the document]
MaggieKimani1: Sorry this is an oversight. Resolving..

src/Microsoft.OpenApi.Readers/Document.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Readers
+{
+    /// <summary>
+    /// Loads an OpenApiDocument instance through Load/LoadAsync/Parse pattern
+    /// </summary>
+    public static class Document
```
baywet, line 14: I'm not sure why those methods aren't defined on OpenAPIDocument directly?

MaggieKimani1: With the current implementation, since the `OpenApiDocument` class resides in the `Microsoft.OpenApi` project, defining the methods within the class would need us to take a dependency on the `Microsoft.OpenApi.Readers` library(for access to the StreamReader, TextReader etc) which already depends on the `Microsoft.OpenApi` lib. This will end up causing a circular dependency between the two.

src/Microsoft.OpenApi.Readers/Document.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        /// <summary>
+        /// Loads an OpenApiDocument from a file stream
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <param name="cancellationToken"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static async Task<ReadResult> LoadAsync(Stream stream, OpenApiReaderSettings settings = null, CancellationToken cancellationToken = default)
```
baywet, line 83: why is there any difference with the other loadAsync in terms of async/await?

src/Microsoft.OpenApi.Readers/Document.cs
baywet: right, thanks for clarifying.

src/Microsoft.OpenApi.Readers/Document.cs
```diff
@@ -0,0 +1,88 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.OpenApi.Readers
```
darrelmiller, line 9: Move this to the Microsoft.OpenAPI namespace so that a user doesn't have to pull in a second namespace just to be able to load a document.

src/Microsoft.OpenApi.Readers/Document.cs
```diff
@@ -0,0 +1,88 @@
...
+    public static class Document
+    {
+        /// <summary>
+        /// Loads an OpenApiDocument from a file stream
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <param name="diagnostic"></param>
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static OpenApiDocument Load(Stream stream, out OpenApiDiagnostic diagnostic, OpenApiReaderSettings settings = null)
```
darrelmiller, line 23: These should be extension methods so that they appear on the OpenApiDocument class.

src/Microsoft.OpenApi.Readers/Document.cs
darrelmiller: What is the best practice when an method only calls one other method that is async?  Should the inner method be awaited or should we just return the Task?

baywet: I've seen arguments both ways, but the "perf recommendation" is to avoid instantiating a new task for no reason and just skip async/await to return the task.

src/Microsoft.OpenApi.Readers/Document.cs
baywet: we don't have an instance of OpenAPIDocument yet, and as far as I know static extension methods are not a thing. Maybe the design needs to be reviewed here?

MaggieKimani1: > These should be extension methods so that they appear on the OpenApiDocument class.

So the format should ideally look like this: 
```c#
var doc = new OpenApiDocument().Load(stream)
```

baywet: in which case we wouldn't be returning the document from the method? Works for me but I'm not sure the internals of OpenAPIDocument are accessible in an extension method outside of the assembly.

VisualBean: This would create an odd pattern of
``` csharp
var document = new OpenApiDocument();

document.Load();
```

Where document would now be null.


Perhaps marking OpenApiDocument as partial could do the trick as well (adding the static load methods in the reader lib)

baywet: thanks for jumping in the discussion. The partial solution would only work if the classes were part of the same assembly and namespace. But these methods need to be in the readers package since they have dependency on readers. Whereas the OpenAPIDocument class is in the main package.

VisualBean: Forgot about the assembly part.

MaggieKimani1: > in which case we wouldn't be returning the document from the method? Works for me but I'm not sure the internals of OpenAPIDocument are accessible in an extension method outside of the assembly.

We can return a document in order to avoid null references when we return a void as @VisualBean has indicated [here](https://github.com/microsoft/OpenAPI.NET/pull/1446#discussion_r1377830123).

baywet: This pattern is strange though. Why define those methods as extension methods if we're returning a new instance? How is the consumer supposed to know which instance to hold on to?

VisualBean: Perhaps simply making static methods on the readers will provide an adequate solution.

The main issue (compared to XmlDocument) is that reading is separate from the models.

The main problem Darrel seems to reference is that the readers aren't really 'readers', they simply parse the entirety of a string/stream/whatever and make a document (or fragment), so newing them seems wrong, as they don't keep state (other than options).

src/Microsoft.OpenApi.Readers/OpenApiDocumentExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers;
+
+namespace Microsoft.OpenApi
+{
+    /// <summary>
+    /// Loads an OpenApiDocument instance through Load/LoadAsync/Parse pattern
+    /// </summary>
+    public static class OpenApiDocumentExtensions
```
irvinesunday, line 15: Maybe rename this class to `OpenApiDocumentReaderExtensions`

MaggieKimani1: The current class implementation extends the OpenApiDocument object, not the document readers hence the name. The only issue is that we can't add it to the models lib due to the circular dependency issue.

src/Microsoft.OpenApi.Readers/OpenApiDocumentExtensions.cs
```diff
@@ -0,0 +1,113 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. 
+
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers;
+
+namespace Microsoft.OpenApi
```
irvinesunday, line 10: Should it not be in the `Microsoft.OpenApi.Readers` namespace ?

MaggieKimani1: Refer to [this comment](https://github.com/microsoft/OpenAPI.NET/pull/1446#discussion_r1377615854)

darrelmiller:
## Goals
- We would like to move the loading code into OpenApi.Net core library so that people who only want to load JSON don't need a dependency on the third party YAML library
- We would like to avoid developers needing to discover a secondary "factory/reader" class and simply be able to use a static member on the model to load/parse content

## Constraints
- We cannot create a circular dependency between the core library and reader library
- We must support loading any OpenAPI model object directly, not just a complete document. (fragments)
- We don't want to duplicate all of the loading/parsing code in every model object's static methods.

## Proposal

In the core library we create a static factory for loading/parsing any OpenAPI model object.

T OpenApiModelFactory.LoadAsync<T>(url)
T OpenApiModelFactory.LoadAsync<T>(stream, format)
T OpenApiModelFactory.LoadAsync<T>(textReader, format)
...
T OpenApiModelFactory.Parse<T>(string)

We can add simple static helper methods to the OpenAPIDocument class to avoid the "discovery problem". 
OpenApiDocument OpenApiDocument.LoadAsync(stream)  -> Calls factory
OpenApiDocument OpenApiDocument.Parse(string)  -> Calls factory

We need to be able to add support for other formats in the factory:

OpenApiModelFactory.RegisterReader(IOpenApiReader)  <- Making this up. Look at how Kiota does it.

OpenApiJsonReader : IOpenApiReader   <- Lives in the core library
OpenApiYamlReader : IOpenApiReader  <- Lives in the reader library

OpenApiModelFactory.RegisterReader(new OpenApiJsonReader)  <= Do this by default in a static ctor.

// User chooses to enable YAML parsing by adding reference to Microsoft.OpenApiReaders
OpenApiModelFactory.RegisterReader(new OpenApiYamlReader) ;

baywet:
@darrelmiller

> OpenApiJsonReader : IOpenApiReader <- Lives in the core library 

Is this representative of things now or a desired change? Why would we play favorite child with a specific format?

> OpenApiDocument OpenApiDocument.LoadAsync(stream) -> Calls factory

Assuming the factory lives in readers, and the LoadAsync method in core, this is not going to be possible due to circular references

> User chooses to enable YAML parsing by adding reference to Microsoft.OpenApiReaders

Didn't we say we'd enable auto-registration through static constructors? (I've never done that in the past, I'm not sure this is possible)

MaggieKimani1:
Superseded by #1553

PR #1492, opened 2023-12-04
Fix culture dependency in ParseDecimalWithFallbackOnOverflow_Overflows_ReturnsFallback test

dldl-cmd:
Fixes a problem when the tests are failing because the test input is generated with a culture dependend input. 

This happens on for example german systems because the dot is replaced with a comma and therefore a FormatException is throw.


MaggieKimani1 (approved):

MaggieKimani1:
Thanks for the contribution!

PR #1491, opened 2023-12-04
Fix unresolved references in OpenApiWalker

dldl-cmd:
Fixes #1490

PR #1504, opened 2023-12-18
bugfix/conflicting path

baywet:
- - fixes a bug where paths with multiple parameters would be considered conflicting
- - bumps version number for release


darrelmiller:
Disabled until we validate that the PR https://github.com/microsoft/OpenAPI.NET/pull/1491 doesn't cause a problem

baywet:
@maggiekimani1 I removed the version bump commit here so we can merge this PR ahead of #1525 and without having to wait on #1491
We'll bump and release whenever we're ready this way.

MaggieKimani1 (approved):

Issue #1506, opened 2023-12-19
Components that are reference objects do not work


darrelmiller:
Here is an example of a security object.  Reading and writing this example fails to produce a valid OpenAPI. 
![image](https://github.com/microsoft/OpenAPI.NET/assets/447694/8cece9c2-9bdf-4973-be70-01ec4c3e1806)

```yaml
openapi: 3.0
info:
  version: 2.0.0
  title: hello world
paths:
  /:
    get:
      responses:
        '200':
          description: OK

security:
  - ReferenceObject: []
components:
  securitySchemes:
    ReferenceObject:
      $ref: 'external.yaml#/components/securitySchemes/RealObject'    
    RealObject:
      type: http
      scheme: basic

```
A schema component that is reference object also fails for different reasons. In the output OpoenAPI, the transitive reference is lost and the inline reference points directly to the "realObject" instead of the "referenceObject".
![image](https://github.com/microsoft/OpenAPI.NET/assets/447694/1bb88a1d-8281-4b12-bcbb-e0c532ae4f93)
```yaml
openapi: 3.0
info:
  version: 2.0.0
  title: hello world
paths:
  /:
    get:
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ReferenceObject'
components:
  schemas:
    ReferenceObject:
      $ref: '#/components/schemas/RealObject'
    RealObject:
      type: object
      properties:
        name: 
          type: string

```

Because of these bugs we need to revert PR #1491

dldl-cmd:
For the first issue, with the invalid written securityScheme this can be reproduced, by the following code, which indicates that it happens while writting the document.

```C#
var openApiDocument = new OpenApiDocument
{
    Components = new OpenApiComponents
    {
        SecuritySchemes = new Dictionary<string, OpenApiSecurityScheme>
        {
            ["ReferenceObject"] = new OpenApiSecurityScheme
            {
                Reference = new OpenApiReference
                {
                    Id = "RealObject",
                    Type = ReferenceType.Tag,
                    ExternalResource = "external.yaml"
                }
            },
            ["RealObject"] = new OpenApiSecurityScheme
            {
                Type = SecuritySchemeType.Http,
                Scheme = "scheme"
            }
        }
    }
};

var memoryStream = new MemoryStream();
var streamWriter = new FormattingStreamWriter(memoryStream, CultureInfo.InvariantCulture);
[...]

Issue #1490, opened 2023-12-04
External reference to SecuritySchemes not loading documents


dldl-cmd:
When reading an OpenApi document not all documents get loaded into the workspace. When a reference to a security scheme in another file exists this file does not get loaded into the workspace and there the reference cannot be resolved.

**OpenApi File & Program example To Reproduce**

`ExampleOpenApi.yaml`
```yaml
openapi: '3.0.2'

info:
    title: Example API
    description: Example for OpenApiWalker reference problem
    version: 1.0.0

servers:
    - url: 'https://localhost'

paths:
    /example:
        get:
            summary: "Test"
            responses:
              '200':
                description: Example response
                content:
                    application/json:
                        schema:
                            $ref: CommonSchemas.yaml#/components/schemas/ExampleSchema

security:
    - TestSecurityScheme: []

components:
    securitySchemes:
        TestSecurityScheme:
            $ref: CommonSecuritySchemas.yaml#/components/securitySchemes/TestSecurityScheme
```

`CommonSchemas.yaml`
```yaml
openapi: '3.0.2'

info:
    title: Common definitions
    description: Common definitions for OpenApiWalker reference problem
    version: 1.0.0

paths:

components:
    schemas:
        ExampleSchema:
            type: object
            properties:
              PropA:
                type: string
              PropB:
                type: string
```

`CommonSecuritySchemas.yaml`
```yaml
openapi: '3.0.2'

info:
    title: Common definitions
    description: Common definitions for OpenApiWalker reference problem
    version: 1.0.0

paths:

components:
    securitySchemes:
        TestSecurityScheme:
            type: oauth2
            flows:
                clientCredentials:
                    tokenUrl: https://localhost/connect/token
                    scopes:
                        Read: Allows to read data
                        Write: Allows to write data
                        Admin: Allows to change access to data
```

Example program:

```C#
using Microsoft.OpenApi.Models;
using Microsoft.OpenApi.Readers;

await using var openApiContent = new FileStream("ExampleOpenApi.yaml", FileMode.Open, FileAccess.Read);

var readerSettings = new OpenApiReaderSettings
{
    LoadExternalRefs = true,
    BaseUrl = new Uri(AppContext.BaseDirectory, UriKind.Absolute)
};

var reader = new OpenApiStreamReader(readerSettings);
var readResult = await reader.ReadAsync(openApiContent);

var workspace = readResult.OpenApiDocument.Workspace;

[...]

PR #1507, opened 2023-12-19
Revert "Fix unresolved references in OpenApiWalker"

darrelmiller:
Reverts microsoft/OpenAPI.NET#1491 due to issues identified in #1506

### DIFF
--- a/src/Microsoft.OpenApi/Services/OpenApiWalker.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiWalker.cs
@@ -1091,8 +1091,7 @@ namespace Microsoft.OpenApi.Services
         /// </summary>
         private bool ProcessAsReference(IOpenApiReferenceable referenceable, bool isComponent = false)
         {
-            var isReference = referenceable.Reference != null && 
-                              (!isComponent || referenceable.UnresolvedReference);
+            var isReference = referenceable.Reference != null && !isComponent;
             if (isReference)
             {
                 Walk(referenceable);
--- a/test/Microsoft.OpenApi.Tests/Walkers/WalkerLocationTests.cs
+++ b/test/Microsoft.OpenApi.Tests/Walkers/WalkerLocationTests.cs
@@ -233,18 +233,6 @@ namespace Microsoft.OpenApi.Tests.Walkers
                     Headers = new Dictionary<string, OpenApiHeader>
                     {
                         ["test-header"] = testHeader
-                    },
-                    SecuritySchemes = new Dictionary<string, OpenApiSecurityScheme>
-                    {
-                        ["test-secScheme"] = new OpenApiSecurityScheme
-                        {
-                            Reference = new OpenApiReference
-                            {
-                                Id = "reference-to-scheme",
-                                Type = ReferenceType.SecurityScheme
-                            },
-                            UnresolvedReference = true
-                        }
                     }
                 }
             };
@@ -257,7 +245,6 @@ namespace Microsoft.OpenApi.Tests.Walkers
                 "referenceAt: #/paths/~1/get/responses/200/content/application~1json/schema",
                 "referenceAt: #/paths/~1/get/responses/200/headers/test-header",
                 "referenceAt: #/components/schemas/derived/anyOf/0",
-                "referenceAt: #/components/securitySchemes/test-secScheme",
                 "referenceAt: #/components/headers/test-header/schema"
             });
         }


MaggieKimani1 (approved):

baywet:
I'd like to see the unit tests outlined in #1506 added as part of the revert if possible. This way if the author of the PR wants to come back with an updated PR, it'll prevent any regression.

dldl-cmd:
I tried to write a test for the found issue, but it seems, that it also does not work when the fix was not applied. I tried the following based of commit d1cf89f9293196080268ae49ed3c01f65c642816. Therefore I think reverting the fix will not be beneficial. 

**securitySchemeReference.yaml**
```yaml
openapi: 3.0
info:
  version: 2.0.0
  title: hello world
paths:
  /:
    get:
      responses:
        '200':
          description: OK

security:
  - ReferenceObject: []
components:
  securitySchemes:
    ReferenceObject:
      $ref: 'external.yaml#/components/securitySchemes/RealObject'    
    RealObject:
      type: http
      scheme: basic
```

**OpenApiReadingWritingTests.cs**
```C#
using System.Globalization;
using System.IO;
using System.Text;
using FluentAssertions;
using Xunit;

namespace Microsoft.OpenApi.Readers.Tests.V3Tests
{
    [Collection("DefaultSettings")]
    public class OpenApiReadingWritingTests
    {
        private const string SampleFolderPath = "V3Tests/Samples/OpenApiReadingWriting/";

        [Fact]
        public void ReadingAndWritingShouldResultInTheSameDocument()
        {
            using var stream = Resources.GetStream(Path.Combine(SampleFolderPath, "securitySchemeReference.yaml"));
            var streamReader = new StreamReader(stream, Encoding.UTF8);
            var sampleDocument = streamReader.ReadToEnd();
            stream.Seek(0, SeekOrigin.Begin);
            var openApiDoc = new OpenApiStreamReader().Read(stream, out var diagnostic);


            var memoryStream = new MemoryStream();
            var streamWriter = new FormattingStreamWriter(memoryStream, CultureInfo.InvariantCulture);
            var writer = new OpenApiYamlWriter(streamWriter);

            openApiDoc.SerializeAsV3(writer);

            writer.Flush();

            var writtenDocument = Encoding.UTF8.GetString(memoryStream.ToArray());

            // Assert
            diagnostic.Should().BeEquivalentTo(new OpenApiDiagnostic
            {
                SpecificationVersion = OpenApiSpecVersion.OpenApi3_0
            });

            sampleDocument.Should().BeEquivalentTo(writtenDocument);
        }
    }
}
```

Results in a file with the following content:
```yaml
openapi: 3.0.1
info:
  title: hello world
  version: 2.0.0
paths:
  /:
    get:
      responses:
        '200':
          description: OK
components:
  securitySchemes:
    ReferenceObject:
    external.yaml#/components/securitySchemes/RealObject:
    RealObject:
      type: http
      scheme: basic
security:
  - external.yaml#/components/securitySchemes/RealObject: [ ]
```

MaggieKimani1:
I have also tested the concerns highlighted on #1506 and found that they're unrelated to the fix done by #1491.
Ultimately the issue being addressed is should we support defining external references at the root of the component?
For instance using the example below:

File 1: `ExampleSchema.yaml:`
``` yaml
openapi: '3.0.2'
info:
    title: Example API
    description: Example for OpenApiWalker reference problem
    version: 1.0.0
servers:
    - url: 'https://localhost'
paths:
    /example:
        get:
            summary: "Test"
            responses:
              '200':
                description: Example response
                content:
                    application/json:
                        schema:
                            $ref: '#/components/schemas/ExampleSchema'
components:
    schemas:
        ExampleSchema:
           $ref: CommonSchema.yaml#/components/schemas/ExampleSchema
```

File 2. `CommonSchema.yaml`

``` yaml
openapi: '3.0.2'
info:
    title: Common definitions
    description: Common definitions for OpenApiWalker reference problem
    version: 1.0.0
paths: {}
components:
    schemas:
        ExampleSchema:
            type: object
            properties:
              PropA:
                type: string
              PropB:
                type: string
```
The `CommonSchema.yaml` file doesn't get loaded into the `OpenApiWorkSpace` due to the same issue highlighted on this comment [https://github.com/microsoft/OpenAPI.NET/issues/1490#issuecomment-1838428614](url)
We should be able to support visiting remote references that exist directly in a component hence the fix #1491 shouldn't be reverted.

darrelmiller:
I do not believe it is possible to properly support reference objects at the root of a component with the current design.  I believe this PR reintroduces the condition that allows the following to get rendered properly.

```yaml
openapi: '3.0.2'
info:
    title: Example API
    description: Example for OpenApiWalker reference problem
    version: 1.0.0
components:
  schemas: 
   schemaA:
     $ref: "#/components/schemas/schemaB"
   schemaB:    
     type: string
```

I am not suggesting this is particularly useful, but this is a scenario that we used to support.

darrelmiller:
After review, we decided that the bug that this PR introduces where we can't accurately write out a document that has a double reference is a lesser bug than not resolving a reference at the root of the component.  Therefore we will keep this PR.

PR #1524, opened 2024-01-11
Fixes v2 examples serialization in a response object

MaggieKimani1:
- Retains examples present in a media type object by adding serialization logic of the examples map to `x-examples` during down-casting from v3 to v2
- Adds logic to cater for the round trip i.e.  serializing `x-examples` in a v2 document to examples in the resulting v3 document during upcasting
- Fixes #1430 

### Snippets
``` yaml
openapi: 3.0.1
info:
  title: Pet Store API
  version: 1.0.0
paths:
  /pets:
    get:
      summary: Get all pets
      responses:
        '200':
          description: Successful response
          content:
            application/json:
              schema:
                type: array
                items:
                  type: object
                  properties:
                    name:
                      type: string
                    age:
                      type: integer
              examples:
                example1:
                  summary: Example - List of Pets
                  value:
                    - name: "Buddy"
                      age: 2
                    - name: "Whiskers"
                      age: 1
               example2:
                  summary: Example - Playful Cat
                  value:
                    name: "Whiskers"
                    age: 1
```
Results to:
``` yaml
swagger: '2.0'
info:
  title: Pet Store API
  version: 1.0.0
paths:
  /pets:
    get:
      summary: Get all pets
      produces:
        - application/json
      responses:
        '200':
          description: Successful response
          schema:
            type: array
            items:
              type: object
              properties:
                name:
                  type: string
                age:
                  type: integer
          x-examples:
            example1:
              summary: Example - List of Pets
              value:
                - name: "Buddy"
                  age: 2
                - name: "Whiskers"
                  age: 1
           example2:
              summary: Example - Playful Cat
              value:
                name: "Whiskers"
                age: 1
```
And the reverse is also true.

PR #1538, opened 2024-01-23
Fix: Examples serialization in a response object during v2->v3 upcasting or vice versa

MaggieKimani1:
This is a rebase of #1524 on our `vnext` branch since its not a breaking change

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
@@ -150,7 +150,7 @@ namespace Microsoft.OpenApi.Models
                 Required = Required,
                 Extensions = Extensions.ToDictionary(static k => k.Key, static v => v.Value)  // Clone extensions so we can remove the x-bodyName extensions from the output V2 model.
             };
-            if (bodyParameter.Extensions.TryGetValue(OpenApiConstants.BodyName, out var bodyParameterName))
+            if (bodyParameter.Extensions.ContainsKey(OpenApiConstants.BodyName))
             {
                 bodyParameter.Name = (Extensions[OpenApiConstants.BodyName] as OpenApiString)?.Value ?? "body";
                 bodyParameter.Extensions.Remove(OpenApiConstants.BodyName);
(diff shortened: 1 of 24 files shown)


src/Microsoft.OpenApi/Writers/IOpenApiWriter.cs
```diff
@@ -72,5 +74,13 @@ public interface IOpenApiWriter
...
         /// </summary>
         void Flush();
+
+        /// <summary>
+        /// Writes out existing examples in a mediatype object
+        /// </summary>
+        /// <param name="writer"></param>
+        /// <param name="example"></param>
+        /// <param name="version"></param>
+        void WriteV2Examples(IOpenApiWriter writer, OpenApiExample example, OpenApiSpecVersion version);
```
baywet, line 84: this change is source breaking. I'm not sure what's been our policy around versioning in this lib, but if we follow semver (which seems like we do), this would mandate a major bump.

src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
```diff
@@ -153,10 +146,11 @@ internal OpenApiBodyParameter ConvertToBodyParameter()
                 // To allow round-tripping we use an extension to hold the name
                 Name = "body",
                 Schema = Content.Values.FirstOrDefault()?.Schema ?? new OpenApiSchema(),
+                Examples = Content.Values.FirstOrDefault()?.Examples,
                 Required = Required,
                 Extensions = Extensions.ToDictionary(static k => k.Key, static v => v.Value)  // Clone extensions so we can remove the x-bodyName extensions from the output V2 model.
             };
-            if (bodyParameter.Extensions.ContainsKey(OpenApiConstants.BodyName))
```
baywet, line 159: I'm not sure why that change was made? the bodyParameterName is not in use

baywet (changes requested):

src/Microsoft.OpenApi/Writers/IOpenApiWriter.cs
darrelmiller: This method also seems really out of place in this interface.  If I am creating a new writer, why would I have to create a special method for writing v2 examples.

darrelmiller: The original request from the team who need this primarily care about the fact that they can't read these examples. They don't have a use case for round-tripping them.  If we really need a breaking change then we can push the support for round tripping them to v2 and just ship the reading for V1.x

MaggieKimani1: I can refactor this by abstracting the method from the interface so that its no longer a breaking change. The round-tripping does not introduce a breaking change so I think we can retain that functionality.

baywet (approved):

darrelmiller (approved):

PR #1545, opened 2024-01-31
Migrate projects targeting net7 to net8

### LINKED ISSUE #1533: Migrate projects to .NET 8

MaggieKimani1:
Fixes #1533

### DIFF
--- a/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
+++ b/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFramework>net8.0</TargetFramework>
+        <TargetFramework>netstandard2.0</TargetFramework>
         <LangVersion>latest</LangVersion>
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
         <Version>1.6.11</Version>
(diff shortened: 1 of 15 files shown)


src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFramework>netstandard2.0</TargetFramework>
+        <TargetFramework>net8.0</TargetFramework>
```
baywet, line 3: this effectively means we're dropping support for netfx, and most of net core except for the latest.
I'd like to make sure we've evaluated the consequences for our customers with the PM before we make changes to the code for the libraries (changing hidi and tests are fine) ? (I don't want us to have to backtrack the work later on when people complain about too high of a target)

baywet (changes requested):

src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
MaggieKimani1: I can revert the libs to continue supporting netstandard2.0 so that we don't drop support for customers using it and only upgrade hidi and the tests.

baywet: I think the upgrade is great for all other projects besides OpenAPI and readers. For those we should keep netstandard2.0 if we're not expecting huge benefits from new APIs, or make sure we make a case for it, understand how much of the audience we'd be "loosing" and probably target something lower like net6.

baywet (changes requested):
the pipelines needs to be updated
https://github.com/microsoft/OpenAPI.NET/blob/35045ae5a0d590ec1eb9bc50a638de3c3e36cba1/.azure-pipelines/ci-build.yml#L38
https://github.com/microsoft/OpenAPI.NET/blob/35045ae5a0d590ec1eb9bc50a638de3c3e36cba1/.github/workflows/ci-cd.yml#L19
etc...

baywet (approved):
thanks for making the changes

PR #1553, opened 2024-02-07
Use loading pattern

### LINKED ISSUE #1434: Change the way OpenAPI documents are loaded in v2

### LINKED ISSUE #1447: Short-circuit JSON documents to avoid YAML parser
Now that ParseNode is a wrapper around JSON Node and not YAML node we can improve performance for JSON documents by avoiding the YAML parser.  We can only do this when we know that the input document is JSON.

To determine if we can parse the JSON directly we use the following heuristics:
- If the file path ends in ".json" we assume it is JSON content
- If the http URL ends in ".json" we assume it is JSON content
- If the HTTP request returns a response content type that is either "application/json" or ends in "+json" then we assume the content is JSON.

All string comparisons can be case insensitive.

MaggieKimani1:
This PR is an implementation of the loading design proposed as part of this conversation: https://github.com/microsoft/OpenAPI.NET/pull/1446#issuecomment-1798795124 which seeks to resolve the issue highlighted in https://github.com/microsoft/OpenAPI.NET/issues/1434

It involves:
- Creating a static factory for loading/parsing any OpenAPI model object that implements the Load/LoadAsync/Parse pattern as per this implementation https://learn.microsoft.com/en-us/dotnet/api/system.xml.linq.xdocument.load?view=net-7.0
- Introducing static Load/LoadAsync/Parse methods in the OpenAPI document class to eliminate the discovery issue
- Adding a registry class that allows users to register different OpenAPI format providers and their implementations
- Moves the loading code into the core library so that people who only want to load JSON don't need a dependency on the third party YAML library
- Short-circuits parsing of JSON documents by avoiding the YAML parser.

Fixes #1434
Fixes #1447


src/Microsoft.OpenApi/Models/OpenApiModelFactory.cs
```diff
@@ -0,0 +1,221 @@
...
+            format ??= OpenApiConstants.Json;
+            return OpenApiReaderRegistry.GetReader(format).Read<T>(input, version, out diagnostic, settings);
+        }
+
+
+        private static string GetContentType(string url)
+        {
+            var response = _httpClient.GetAsync(url).GetAwaiter().GetResult();
+            var contentType = response.Content.Headers.ContentType.MediaType;
+            if (contentType.EndsWith(OpenApiConstants.Json, StringComparison.OrdinalIgnoreCase))
```
andrueastman, line 172: You probably want logic similar to https://github.com/microsoft/kiota-abstractions-dotnet/blob/4bff08d4ff90b24c705f71ba4d3baf89a85f31c0/src/serialization/ParseNodeFactoryRegistry.cs#L49

Using `endsWith` may bring issues in scenarios where subtypes are present.
https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#content-type

src/Microsoft.OpenApi/Models/OpenApiModelFactory.cs
```diff
@@ -0,0 +1,221 @@
...
+                return OpenApiConstants.Yaml;
+            }
+            return null;
+        }
+
+        private static string GetFormat(string url)
+        {
+            if (!string.IsNullOrEmpty(url))
+            {
+                if (url.StartsWith("http", StringComparison.OrdinalIgnoreCase) || url.StartsWith("https", StringComparison.OrdinalIgnoreCase))
```
andrueastman, line 187: Out of curiosity are yaml/json the only formats that could exist? If not, it may make sense to simplify this to only pull the suffix and use for lookup in the registry(in the event that the caller already registered that format in the registry)

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
```diff
@@ -0,0 +1,494 @@
...
+            // Create workspace for all documents to live in.
+            var openApiWorkSpace = new OpenApiWorkspace();
+
+            // Load this root document into the workspace
+            var streamLoader = new DefaultStreamLoader(settings.BaseUrl);
+            var workspaceLoader = new OpenApiWorkspaceLoader(openApiWorkSpace, settings.CustomExternalLoader ?? streamLoader, settings);
+            await workspaceLoader.LoadAsync(new OpenApiReference() { ExternalResource = "/" }, document, OpenApiConstants.Json, null, cancellationToken);
+        }
+
+        private Stream GetStream(string url)
```
andrueastman, line 456: Similar comment to other reader on code duplication.

src/Microsoft.OpenApi/Models/OpenApiModelFactory.cs
MaggieKimani1: Yeah using the suffix for a registry lookup sounds more effective. Let me update this

src/Microsoft.OpenApi/Interfaces/IOpenApiReader.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Reader;
+
+namespace Microsoft.OpenApi.Interfaces
+{
+    /// <summary>
+    /// Interface for Open API readers.
+    /// </summary>
+    public interface IOpenApiReader
```
darrelmiller, line 15: Why do we still have this interface?  I thought the idea was to remove the "reader" pattern and replace with Load

src/Microsoft.OpenApi/Interfaces/IOpenApiReader.cs
```diff
@@ -0,0 +1,121 @@
...
+    {
+        /// <summary>
+        /// Reads the input URL and parses it into an Open API document.
+        /// </summary>
+        /// <param name="url">The input to read from.</param>
+        /// <param name="diagnostic">The diagnostic entity containing information from the reading process.</param>
+        /// <param name="settings"> The OpenApi reader settings.</param>
+        /// <returns></returns>
+        OpenApiDocument Read(string url, out OpenApiDiagnostic diagnostic, OpenApiReaderSettings settings = null);
```
darrelmiller, line 24: This interface doesn't need all of these options.  The user will never call these.  We only need the most primitive one. e.g. TextReader

src/Microsoft.OpenApi/Reader/OpenApiReaderRegistry.cs
```diff
@@ -0,0 +1,48 @@
...
+        public static readonly IOpenApiReader DefaultReader = new OpenApiJsonReader();
+
+        /// <summary>
+        /// Registers an IOpenApiReader for a given OpenAPI format.
+        /// </summary>
+        /// <param name="format">The OpenApi file format.</param>
+        /// <param name="reader">The reader instance.</param>
+        public static void RegisterReader(string format, IOpenApiReader reader)
+        {
+            _readers[format] = reader;
```
VisualBean, line 29: consider normalizing the format.
ToLower() or something.

MaggieKimani1: This won't be necessary since I've initialized the _readers dictionary with a case-insensitive string comparer hence all entries are treated as case-insensitive so we don't need to normalize the format

darrelmiller (approved):
I think this PR looks good.  I want have a conversation about having a Load method that doesn't require you to say what the format is, but we can defer that to conversation to after this PR has been merged..

PR #1581, opened 2024-03-06
Fix OpenApiStreamReader.ReadAsync doesn't respect "leave open" setting

### LINKED ISSUE #1579: OpenApiStreamReader.Read/ReadAsync doesn't honour "leaveOpen"Setting
https://github.com/microsoft/OpenAPI.NET/blob/e65b484b512a84351054979bf2901b2462984582/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs#L76C36-L76C48

The value from the leave open setting should be passed to the [StreamReaderConstructor](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=net-8.0#system-io-streamreader-ctor(system-io-stream-system-text-encoding-system-boolean-system-int32-system-boolean)) so it doesn't dispose the stream when asked to keep it open.

Repro:

```CSharp
var stream = new MemoryStream();
// copy the value from a file stream, string etc.. in the stream
var settings = new OpenApiReaderSettings
{
    LeaveStreamOpen = true,
};
var reader = new OpenApiStreamReader(settings);
var result = await reader.ReadAsync(input, cancellationToken)
stream.Seek(0, SeekOrigin.Begin); // results in a object disposed exception when it shouldn't
```

When this is fixed, we can clean `WorskpaceManagementService.MigrateFromLockFileAsync` in kiota.

MaggieKimani1:
Fixes #1579

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiStreamReader.cs
@@ -73,7 +73,7 @@ namespace Microsoft.OpenApi.Readers
                 bufferedStream.Position = 0;
             }
 
-            using var reader = new StreamReader(bufferedStream);
+            using var reader = new StreamReader(bufferedStream, default, true, -1, leaveOpen: _settings.LeaveStreamOpen);
             return await new OpenApiTextReaderReader(_settings).ReadAsync(reader, cancellationToken);
         }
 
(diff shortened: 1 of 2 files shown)


andrueastman (approved):

baywet (commented):
The change also needs to be implemented lines 44 and 89 for consistency.
Thanks for addressing this super fast!

MaggieKimani1:
> The change also needs to be implemented lines 44 and 89 for consistency. Thanks for addressing this super fast!

Here's the PR: https://github.com/microsoft/OpenAPI.NET/pull/1582

PR #1582, opened 2024-03-06
Update the OpenApiStreamReader's Read and ReadFragment method to be consistent with ReadAsync

MaggieKimani1:
This PR builds onto #1581


MaggieKimani1:
@baywet do we need to do a release for this fix to go out and unblock Kiota?

baywet (approved):

baywet:
A release would be nice to clean-up some of the patch work I had to do. But it's not urgent. Just let me know when it's available

PR #1584, opened 2024-03-06
Support trailing slashes in paths

### LINKED ISSUE #1580: Support trailing slashes in OpenAPI paths
_Coming from <https://github.com/microsoft/kiota/issues/4291>_

We're working with an API which endpoints end with trailing slashes, for example `/api/v1/app/{app_id}/msg/`.
See the OpenAPI document at <https://github.com/svix/svix-webhooks/blob/8d32e47e0484f5d0839bce364d8700d2c7457937/openapi.json#L7779>.

We're using Kiota to generate a .NET client for this API, and the generated client doesn't contain the trailing slashes (see the issue at the top for more details).

While we agree it's not common to have trailing slashes, apparently it's supported according to the relevant RFCs.

OpenAPI.NET should respect the OpenAPI path when building out the `OpenApiUrlTreeNode` tree.

N/A.

**Additional context**
I've forked the repo and pushed a branch which shows how the trailing slash is removed, see <https://github.com/microsoft/OpenAPI.NET/compare/vnext...mderriey:OpenAPI.NET:mderriey/support-trailing-slashes?expand=1>.

I haven't opened a PR yet because it might be too much noise given it's not actionable as-is.

mderriey:
fixes #1580

This is a very naïve attempt at supporting trailing slashes in paths.

Please let me know if there are other cases that would need to be covered by the tests, or if this implementation would break in specific scenarios.

### DIFF
--- a/src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
@@ -150,7 +150,7 @@ namespace Microsoft.OpenApi.Services
             }
 
             var segments = path.Split('/');
-            if (path.EndsWith("/"))
+            if (path.EndsWith("/", StringComparison.OrdinalIgnoreCase))
             {
                 // Remove the last element, which is empty, and append the trailing slash to the new last element
                 // This is to support URLs with trailing slashes
--- a/src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
@@ -150,6 +150,13 @@ namespace Microsoft.OpenApi.Services
             }
 
             var segments = path.Split('/');
+            if (path.EndsWith("/"))
+            {
+                // Remove the last element, which is empty, and append the trailing slash to the new last element
+                // This is to support URLs with trailing slashes
+                Array.Resize(ref segments, segments.Length - 1);
+                segments[segments.Length - 1] += "/";
+            }
 
             return Attach(segments: segments,
                           pathItem: pathItem,
--- a/test/Microsoft.OpenApi.Tests/Services/OpenApiUrlTreeNodeTests.cs
+++ b/test/Microsoft.OpenApi.Tests/Services/OpenApiUrlTreeNodeTests.cs
@@ -495,8 +495,9 @@ namespace Microsoft.OpenApi.Tests.Services
                 secondToLeafNode = secondToLeafNode.Children[childName];
             }
 
-            Assert.True(secondToLeafNode.Children.TryGetValue(expectedLeafNodeName, out var lastNode));
-            Assert.Equal(expectedLeafNodePath, lastNode.Path);
+            Assert.True(secondToLeafNode.Children.TryGetValue(expectedLeafNodeName, out var leafNode));
+            Assert.Equal(expectedLeafNodePath, leafNode.Path);
+            Assert.Empty(leafNode.Children);
         }
     }
 }
--- a/test/Microsoft.OpenApi.Tests/Services/OpenApiUrlTreeNodeTests.cs
+++ b/test/Microsoft.OpenApi.Tests/Services/OpenApiUrlTreeNodeTests.cs
@@ -466,5 +466,23 @@ namespace Microsoft.OpenApi.Tests.Services
 
             await Verifier.Verify(diagram);
         }
+
+        [Fact]
+        public void SupportsTrailingSlashesInPath()
+        {
+            var openApiDocument = new OpenApiDocument
+            {
+                Paths = new()
+                {
+                    ["/cars/{car-id}/build/"] = new()
+                }
+            };
+
+            var label1 = "trailing-slash";
+            var rootNode = OpenApiUrlTreeNode.Create(openApiDocument, label1);
+            var buildNode = rootNode.Children["cars"].Children["{car-id}"].Children["build"];
+
+            // Should buildNode have a path of "build/" or should it have a child with an empty string key?
+        }
     }
 }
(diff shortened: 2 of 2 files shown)


src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -150,6 +150,13 @@ public void Attach(OpenApiDocument doc, string label)
             }
 
             var segments = path.Split('/');
+            if (path.EndsWith("/"))
```
baywet, line 153:
```suggestion
            if (path.EndsWith('/'))
```

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -150,6 +150,13 @@ public void Attach(OpenApiDocument doc, string label)
...
+            {
+                // Remove the last element, which is empty, and append the trailing slash to the new last element
```
baywet, line 155: I'm a bit worried that this will result in additional "empty" nodes that were not here previously and would potentially. Can you add unit tests to control for that please?

baywet (changes requested):
Thanks for the contributions, a couple of recommendations.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
mderriey: I've just pushed a commit to augment the tests and make sure that the leaf nodes, which have a path with a trailing slash, have no children.

Let me know if there's other cases we should cover.

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
mderriey: .NET Standard 2.0 doesn't support the `EndsWith(Char)` overload, see <https://learn.microsoft.com/dotnet/api/system.string.endswith?view=netstandard-2.0>.

baywet: thanks, I forgot we had such a low compatibility bar.... Can you pass StringComparison.OrdinalIgnoreCase then please? Won't make a perf difference, but it will avoid triggering static analysis warnings if we add some more in the future.

mderriey: Done, thanks.

test/Microsoft.OpenApi.Tests/Services/OpenApiUrlTreeNodeTests.cs
```diff
@@ -466,5 +466,38 @@ public async Task VerifyDiagramFromSampleOpenAPI()
 
             await Verifier.Verify(diagram);
         }
+
+        public static TheoryData<string, string[], string, string> SupportsTrailingSlashesInPathData => new TheoryData<string, string[], string, string>
+        {
+            // Path, children up to second to leaf, last expected leaf node name, expected leaf node path
+            { "/cars/{car-id}/build/", ["cars", "{car-id}"], "build/", @"\cars\{car-id}\build/" },
+            { "/cars/", [], "cars/", @"\cars/" },
```
baywet, line 474:
```suggestion
            { "/cars/{car-id}/build/", ["cars", "{car-id}"], @"build\", @"\cars\{car-id}\build\" },
            { "/cars/", [], "cars/", @"\cars\" },
```

wouldn't this make more sense?

src/Microsoft.OpenApi/Services/OpenApiUrlTreeNode.cs
```diff
@@ -150,6 +150,13 @@ public void Attach(OpenApiDocument doc, string label)
             }
 
             var segments = path.Split('/');
+            if (path.EndsWith("/", StringComparison.OrdinalIgnoreCase))
+            {
+                // Remove the last element, which is empty, and append the trailing slash to the new last element
+                // This is to support URLs with trailing slashes
+                Array.Resize(ref segments, segments.Length - 1);
+                segments[segments.Length - 1] += "/";
```
baywet, line 158:
```suggestion
                segments[segments.Length - 1] += @"\";
```

baywet (changes requested):
thank you for making the update, I think we're getting closer to a solution here. Left a couple of comments

test/Microsoft.OpenApi.Tests/Services/OpenApiUrlTreeNodeTests.cs
mderriey: That's for you to say.

My interpretation was that `\` was used as as a hierarchy separator, and given the last `/` is not a separator per-se, we might want to keep it as-is to indicate that the last segment is `<whatever>/`.

If you want the changes in, though, that's fine, I'll push them.

baywet: as long as it doesn't derail the logic/node structure, it'd make sense to be consistent.
@darrelmiller I can't remember why backslashes were originally used here instead of forward slashes, would you mind providing some context please?

darrelmiller: I can't find out why we used backslash as a separator. It doesn't make sense to me. I tried looking at all the relevant PRs. Maybe @irvinesunday  remembers?

baywet: alright, thanks for your patience with us @mderriey it seems we suffer from collective amnesia.
In the interest of unblocking this PR, and reducing potential side effects, let's perform the least changes possible, and align everything on backslashes.

Issue #615, opened 2021-07-26
Feature: Add support for overlays


baywet:
the [overlay specification](https://github.com/OAI/Overlay-Specification/blob/main/versions/1.0.0.md) is getting close to its final version. It'd be nice to implement the specification in the library so client applications can take advantage of it for thing like generic errors and such.
The solution should be implemented in a separate package this one takes a dependency on, so people can use this functionality outside of openAPI. But the code can be collocated here.

darrelmiller:
This issue https://github.com/microsoftgraph/msgraph-metadata/issues/124 is dependent on overlays.

darrelmiller:
I don't feel strongly about whether we should create a new project or not.  However, I suspect we will re-use the OpenApiAny types and the ParseNode classes in the reader, so I think it would be easier if it is all part of the same library.  In theory the an Overlay could be used to apply to any JSON/YAML document, however, I think we should scope our efforts to just updating OpenAPI documents.

I believe because of this issue https://github.com/OAI/Overlay-Specification/discussions/8 we are probably going to have to implement our own JSONPath querying.

The update property could be implemented as type IOpenApiAny.

darrelmiller:
Upon further discussion we believe that accepting strongly typed OpenAPI objects is not going to work.

Instead we should limit the update property to only accept OpenApiArray, OpenApiObject and OpenApiPrimitive.

```yaml
overlay: 1.0.0
actions:
 - target: "$.*.get.parameters[name=='top']"
   update: 
     required: true
```

```
var overlay = new OverlayDocument();
var action = new OverlayAction() {
   Target = "...",
   Update = new OpenApiObject() {
      [["required"],[true]]  // With the correct c# syntax  :-)
   }
}
```

ouvreboite:
Given that overlays are based on jsonpath, would it make sense to apply them « before » the deserialization made in OpenAPI.net ?

Or would it makes sense to apply them after (meaning somehow translating the jsonpath to an actual traversal of the C# OpenApiDocument graph)?

If the former, I suppose the overlay processing could be done in a dedicated library, with no actual tie to OpenAPI.net). It’s just a preprocessing step on the raw JSON/YAML document.

[...]

PR #1598, opened 2024-03-18
Overlay support spec doc

njaci1:
add a folder for specs and add the first spec doc for overlay support.


Specs/Overlay Support Spec.md
```diff
@@ -0,0 +1,106 @@
+# Feature Specification: Overlays Support
+
+## Objective
+Support use of Overlays for enabling developers to enhance existing OpenAPI description files without changing the original file.
+
+## Problem Statement
+Existing OpenAPI documents used for AI plugin creation might lack necessary properties or require modifications for them to provide a high quality AI plugin. Direct editing of the original OpenAPI document is often undesirable or impractical.
```
sebastienlevert, line 7: Not AI plugin, but I'd even say referenced OpenAPI descriptions that participate in the GPTs and Plugins ecosystem.

Specs/Overlay Support Spec.md
```diff
@@ -0,0 +1,106 @@
...
+        - name: id
+          in: path
+          required: true
+          description: User ID
+      responses:
+        '200':
+          description: Successful response
+```
+
+**Overlay File (overlay.yaml)**
```
sebastienlevert, line 53: This overlay doesn't use JSON Path to define the structure. Are we saying that this is our MVP for now? Only structured overlays vs. targeted overlays would be supported?

https://github.com/OAI/Overlay-Specification/blob/main/versions/1.0.0.md#targeted-overlays

I'm honestly OK at this point for either, but make it clear would be better and provide a more appropriate debate!

darrelmiller: You can't really separate the two. Targeted overlays use a merge patch at the target. Structured overlays are just the trivial case of targeted.

darrelmiller:
We should add some discussion of the developer experience of applying an overlay. Is this a separate OverlayService class? Is it an extension method on OpenAPIDocument? Does it change the existing document or create a new patched document? Is it in a different project? If not, how do we handle the YAML dependency?

Specs/Overlay Support Spec.md
njaci1: Yeah, you have phrased it better. I'll rephrase in the spec to capture this.

Specs/Overlay Support Spec.md
njaci1: @sebastienlevert  is your question whether we will support either structured or targeted rather than both? We will support both cases.

Specs/Overlay Support Spec.md
adhiambovivian: Why restrict the use-case of overlays to AI plug-ins when OpenAPI.NET library caters to a diverse range of users?

Specs/Overlay Support Spec.md
```diff
@@ -0,0 +1,106 @@
+# Feature Specification: Overlays Support
+
+## Objective
```
adhiambovivian, line 3: Shouldn't the spec include customer application and target audience?

MaggieKimani1:
Maybe use this issue to add onto some of the use cases or scenarios where overlays would come in handy https://github.com/microsoft/OpenAPI.NET/issues/615

Specs/Overlay Support Spec.md
maisarissi: Sure @adhiambovivian . For copilot scenarios we might would want to provide instructions to the orchestrator on how to use the certain function (/createSong as an exmaple), add localization, provide information on how to present the response to the user (like using adaptive cards or other rendering) etc.

Specs/Overlay Support Spec.md
```diff
@@ -0,0 +1,106 @@
...
+## Problem Statement
+Existing OpenAPI documents used for AI plugin creation might lack necessary properties or require modifications for them to provide a high quality AI plugin. Direct editing of the original OpenAPI document is often undesirable or impractical.
+
+## Solution Overview
+Overlays provide a flexible mechanism for extending OpenAPI documents without directly modifying the original files. This allows necessary properties to be added, existing values modified, and the OpenAPI description tailored for effective AI plugin development.
+
+## Functional Requirements
+1. **Overlay File Format Support**
+   - Support overlay files in accordance with the Overlay Specification (v1.0.0), available at [Overlay Specification](https://github.com/OAI/Overlay-Specification/blob/3f398c6/versions/1.0.0.md).
+   - Support both YAML and JSON formats for overlay files.
```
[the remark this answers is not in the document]
njaci1: Yes we will support both. I have included this in the spec.

Specs/Overlay Support Spec.md
njaci1: I have revised the objective statement to try capture the target audience.

njaci1:
> We should add some discussion of the developer experience of applying an overlay. Is this a separate OverlayService class? Is it an extension method on OpenAPIDocument? Does it change the existing document or create a new patched document? Is it in a different project? If not, how do we handle the YAML dependency?

@darrelmiller I have added a section on 'User Journey' to start the discussion. Have a look at it and continue the discussion.

Specs/Overlay Support Spec.md
```diff
@@ -0,0 +1,120 @@
...
+       
+  2. **User Journey**
+     1. Start by loading an existing OpenAPI document or creating one from scratch to create an instance of `OpenApiDocument`.
+        ```csharp
+        var openApiDocument = LoadOpenApiDocument(documentPath);
+
+     3. Use the OverlayService to apply the overlay file to the OpenAPI document. The OverlayService parses the overlay file, applies the changes directly to the OpenApiDocument instance, and performs in-place modifications.
+         ```csharp
+        var overlayService = new OverlayService();
+        overlayService.ApplyOverlay(openApiDocument, overlayPath);
```
darrelmiller, line 32:
```suggestion
        var overlay = await Overlay.LoadAsync(stream);
        var jsonElement = await JsonDocument.ParseAsync(stream).RootElement;
        var newJsonElement = overlay.Apply(jsonElement);
```

darrelmiller: End to end process would be:
- Use OpenApiDocument.Load to load JSON/Yaml file with Overlay referenced in readerSettings
- Load Yaml and translate to JSONNodes or just load JSONNodes
- Load Overlay as JsonNodes.
- Apply Overlay to OpenAPI JsonNodes
- Pass "overlayed" JsonNodes to rest of OpenAPI Parser.

baywet:

PR #1603, opened 2024-03-25
Use proxy references to resolve references

### LINKED ISSUE #1571: Use proxy reference objects in place of `OpenApiReference` objects when models have references
Related to: https://github.com/microsoft/OpenAPI.NET/issues/1555

```[tasklist]
### Tasks
- [ ] Use proxy references during deserialization of models where models have references
- [ ] Validate models with proxy references as references and not as model classes.
- [ ] Properly serialize models which have proxy references objects as their `OpenApiReference`

### PR DESCRIPTION
**Pull request by MaggieKimani1**
This PR:
- Proposes the use of proxy reference objects to resolve both local and external references present in an OpenAPI document. A proxy reference class inherits the properties and methods of the class it proxies and is used to lazily retrieve the reference property values of its target.
- The proxy reference models use a `host OpenApi document` instance passed in their constructor in order to find the target the $refs point to and dereferences them lazily.
- Initially, we had a 2-step process where we load up a document in memory then have a second pass where we walk through the document and use an `OpenApiReferenceResolver` visitor to locate and resolve references. With this approach, the second pass includes walking through the document instance and using a `HostDocumentResolver` visitor that visits all proxy reference objects and sets the value of the host document.

**To note:**
- If one tries to parse a document fragment, they'll need to provide a host document value otherwise reference resolution fails.

Fixes #1571

### CODE REVIEW

File: src/Microsoft.OpenApi/Services/OpenApiWalker.cs [hunk 1]
```diff
@@ -391,11 +394,17 @@ internal void Walk(OpenApiContact contact)
         /// </summary>
         internal void Walk(OpenApiCallback callback, bool isComponent = false)
         {
-            if (callback == null || ProcessAsReference(callback, isComponent))
+            if (callback == null)
```
darrelmiller, line 397: Let's chat about this change. I'm very concerned about changing OpenApiWalker behavior.  I'm not saying this is wrong, I just need to have it explained slowly to me :-)

src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
```diff
@@ -52,6 +53,15 @@ public static bool IsEmailAddress(this string input)
                 return;
             }
 
+            // Resolve the Json schema in memory before validating the data types.
+            var reference = schema.GetRef();
+            if (reference != null)
+            {
+                var referencePath = string.Concat("https://registry", reference.OriginalString.Split('#').Last());
+                var resolvedSchema = (JsonSchema)SchemaRegistry.Global.Get(new Uri(referencePath));
```
darrelmiller, line 61: I'm assuming this reference to the global registry goes away with Irvine's changes?

src/Microsoft.OpenApi/Writers/OpenApiWriterBase.cs
```diff
@@ -446,6 +446,10 @@ public void WriteJsonSchema(JsonSchema schema, OpenApiSpecVersion version)
                     {
                         FindJsonSchemaRefs.ResolveJsonSchema(schema);
                     }
+                    else if (Settings.InlineLocalReferences)
```
darrelmiller, line 449: I'm still not convinced that we should be doing anything about inlining in the writer.  All inlining should be done by a "referenceResolver" that mutates the current document.

darrelmiller (approved):

src/Microsoft.OpenApi/Validations/Rules/RuleHelpers.cs
MaggieKimani1: Yes this goes away in Irvine's PR

Issue #805, opened 2022-03-27
Inlining in 1.3 is only partially useful


darrelmiller:
Inline local works, but in the process you lose the name of the component and therefore passing this code generator will end up generating a different type for every usage.  It should be useful for scenarios like Power Platform though.

Inline external will pull in components from external files.  However, if those components have local references in the external file, those local references don't get brought into the inlined file.  Merging of components during writing seems like a bad approach.  Having the writer render a reference either as a `$ref` or as the complete object seemed fairly harmless for a writer as it did not change the semantics of the document significantly.  However, merging components is a much bigger task. e.g. What if there is a naming conflict?

In v2 of OpenAPI.NET reference objects and target components are going to be distinct concepts and therefore changing which will be rendered on the fly is going to be too much responsibility for the writer.  I suspect the alternative will be to use a visitor to inline references. This will establish the precedent of inlining being a mutation to the document.  At that point we can do things like merge the components of external files.

In the meanwhile, if you want to inline external files successfully, you will need to inlineLocal and inlineExternal.

darrelmiller:
@MaggieKimani1 @irvinesunday @baywet 

As a follow up to our conversation on Tuesday, I have a proposal.

- Creating an OpenAPIDocument via a public ctor will always create an OpenAPIWorkspace.  That document will be considered an "entry" document.
- All model objects will have an optional ctor parameter called hostDocument of type OpenAPIDocument
- If the model object is instantiated without a hostDocument, then reference resolution will fail. That's OK.
- The OpenAPIWorkspace will own an instance of a  "ComponentRegistry" for any component that exists in any document in the OpenAPIWorkspace.  
- Having a ComponentRegistry per OpenAPIWorkspace allows us to avoid using the global SchemaRegistry and keeps JsonSchema.Net's registry as an implementation detail.
- When parsing an entry document.  All components must be registered in the ComponentRegistry with a URI as the identifier. i.e. We should treat all components like JsonSchema.Net treats Schemas.  This also allows us to treat tags and security schemes in a way that is consistent with how $id works for Schemas.
- After parsing the entry document, if reader settings indicate that external references should be loaded, we will walk the entry document looking for external references.  All documents that host external references will be loaded into the same OpenAPIWorkspace and their components will be registered in the same component registry.
- Non-schema Components will be registered with a URI that is unique to the hostdocument and the type of component to avoid clashes.
[...]

PR #1609, opened 2024-04-02
Creates component registries in the `OpenApiWorkspace` for `$ref` resolution

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET/issues/805

This PR:
- Updates the `OpenApiWorkspace.cs` class by adding components registries that store `JsonSchemas` (or fragments), `IOpenApiReferenceable` objects and stream artifacts. These are indexed with unique key identifiers (URIs) that are a component of the unique document id that the components exist or as provided (in the case of stream artifacts or `JsonSchema` fragments). These unique key identifiers are also used to fetch the components from the components registry in the `OpenApiWorkspace`. The format of the URI identifier will be of the format: 
ex: http://openapi.net/{documentId}/components/schema/{name}
- OpenApiWorkspaces are instantiated in the constructors of OpenApiDocuments. The document's `BaseUri` property is also instantiated in the constructor and is composed of the base path http://openapi.net/ and the unique document Id, which is a GUID.
- All reference resolution (local and external) now happens in the workspace of the entry level OpenApiDocument. External documents have their `ExternalResource` identifier registered together with the unique document id of their host document in a `_documentsIdRegister`. This is a look-up table that external ref resolution uses to look up the unique document ids of external documents, given an `ExternalResource` identifier in an `OpenApiReference` object. This document id extracted from the lookup table will be used alongside the base path and the reference object id to construct the unique URI key that will be used to fetch the resolved component from the workspace component registry.


src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
```diff
@@ -62,7 +39,7 @@ public OpenApiWorkspace(Uri baseUrl)
         /// </summary>
         public OpenApiWorkspace()
         {
-            BaseUrl = new("file://" + Environment.CurrentDirectory + $"{Path.DirectorySeparatorChar}" );
+            BaseUrl = new Uri(OpenApiConstants.BaseRegistryUri);
```
darrelmiller, line 42: I'm fighting with the feeling that "it can't be this simple", but if every document gets it's own GUID in the baseURI then we really don't need to use a file path.  I _think_ this is really good.

darrelmiller (approved):
I can't say I have been able to mentally compile every line of this, but I really like the way this is looking.  We need to get this merged with Maggie's changes and then do some extensive testing on it.

darrelmiller (approved):

PR #1626, opened 2024-04-15
Validate Json schema using JsonSchema.NET

### LINKED ISSUE #1620: Use JsonSchema.NET to perform JSON schema validation

MaggieKimani1:
Refactors the current custom JSON schema validation logic in favor of JsonSchema.NET's `schema.Evaluate()` that performs validation of a Json node instance against an existing schema.
Fixes #1620


darrelmiller (approved):

src/Microsoft.OpenApi/Extensions/JsonSchemaBuilderExtensions.cs
```diff
@@ -206,6 +206,7 @@ public void Evaluate(EvaluationContext context)
     /// The nullable keyword
     /// </summary>
     [SchemaKeyword(Name)]
+    [SchemaSpecVersion(SpecVersion.Draft202012)]
     public class NullableKeyword : IJsonSchemaKeyword
```
darrelmiller, line 210: Nullable is not a supported keyword in OpenAPI 3.1  It is implemented by assigning an array to the type property and including null as one of the values in the array.

Issue #1632, opened 2024-04-17
Resolve `JsonSchemas` references with `$id` value pointers


irvinesunday:
Given this OpenAPI description with a `JsonSchema` reference with a pointer to its `$id` value, we need to resolve this successfully:

```CSharp
openapi: 3.1.0
info:
  title: Testbed
  version: 1.0.0
paths:
  /Input:
    get:
      parameters:
        - $ref: '#/components/parameters/size'
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Input'
    post:
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: 'foo://bar/Input.thing'
      responses:
        '200':
          description: OK
components:
  parameters:
    size:
      name: size
      in: query
      required: true
      schema:
        type: integer
        format: int32
      patternProperties:
        ^[a-z]{3}$:
          type: string
  schemas:
    Input:
      $id: 'foo://bar/Input.thing'
      type: object
      properties:
        name:
          type: string
        age:
          type: integer
```

PR #1633, opened 2024-04-17
Register and retrieve `JsonSchema` references with `$id` value pointers 

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET/issues/1632

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
@@ -569,7 +569,7 @@ namespace Microsoft.OpenApi.Models
             }
 
             string uriLocation;
-            string relativePath = "/components/" + reference.Type.GetDisplayName() + "/" + reference.Id;
+            string relativePath = OpenApiConstants.ComponentsSegment + reference.Type.GetDisplayName() + "/" + reference.Id;
 
             uriLocation = useExternal
                 ? Workspace.GetDocumentId(reference.ExternalResource)?.OriginalString + relativePath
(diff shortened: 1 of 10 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -485,22 +485,22 @@ public IOpenApiReferenceable ResolveReference(OpenApiReference reference)
...
+        {
+            const char pound = '#';
             string uriLocation;
-            string id = referenceUri.OriginalString.Split('/')?.Last();
-            string relativePath = "/components/" + ReferenceType.Schema.GetDisplayName() + "/" + id;
-            
-            if (referenceUri.OriginalString.StartsWith("#"))
+            int poundIndex = referenceUri.OriginalString.IndexOf(pound);
+
+            if (poundIndex > 0)
```
andrueastman, line 493: Just confirming, this looks to be functially equivalent to the previous state of  `if (referenceUri.OriginalString.StartsWith("#"))`.
Any reason for the change as `poundIndex` isn't used elsewhere?

irvinesunday: In the current implementation, I'm specifically checking that:
1. The `referenceUri` string value has a # present, and
2. The # is not the 1st character in the string value

In the previous example we were checking whether the # character is the 1st character in the  `referenceUri` string value.

andrueastman: Thanks for clarifying this!

andrueastman (approved):

PR #1636, opened 2024-04-22
Initial changes for serializing JSON Schema

darrelmiller:


darrelmiller:
@MaggieKimani1 I merged the release/2.0.0 branch into here and there are now more failing tests.  I have been looking at the extensionskeyword.  I am starting to think that we won't be able to use an extension keyword.  I suspect we will need to use a UnrecognizedKeyword class for 'x-' properties because the ExtensionKeyword does not serialize.  I'm not sure why though.

darrelmiller:
I'm still on the fence as to whether it is easier to fix the tests in this PR or to re-apply the changes in this PR to the main feature branch.  Whatever is easier.

MaggieKimani1:
> I'm still on the fence as to whether it is easier to fix the tests in this PR or to re-apply the changes in this PR to the main feature branch. Whatever is easier.

I can try cherry-pick the changes and apply them to the v2 branch just to see if the tests will still fail.

src/Microsoft.OpenApi.Hidi/Formatters/PowerShellFormatter.cs
```diff
@@ -350,12 +350,12 @@ private static JsonSchema CopySchema(JsonSchema schema, JsonSchema newSchema)
...
-            }
-            if (schema.GetOpenApiExternalDocs() == null && newSchema.GetOpenApiExternalDocs() is { } externalDocs)
-            {
-                schemaBuilder.OpenApiExternalDocs(externalDocs);
+                schemaBuilder.Discriminator(discriminator.PropertyName,discriminator.Mapping, discriminator.Extensions);
             }
+            //if (schema.GetExternalDocs() == null && newSchema.GetExternalDocs() is { } externalDocs)
+            //{
+            //    schemaBuilder.ExternalDocs(new Uri(externalDocs), externalDocs.Description);
+            //}
```
MaggieKimani1, line 358: @darrelmiller why did we get rid of ExternalDocs here?

MaggieKimani1:
> > I'm still on the fence as to whether it is easier to fix the tests in this PR or to re-apply the changes in this PR to the main feature branch. Whatever is easier.
> 
> I can try cherry-pick the changes and apply them to the v2 branch just to see if there'll be much of a difference and if the tests will still fail.

**Update**:
Most of these changes are crucial especially the serialization ones so I've ended up lifting almost all of them. Now we have 99 tests failing on the V2 branch vs around 40 on this branch 😢
Maybe we can work on cleaning up this branch instead?

src/Microsoft.OpenApi.Hidi/Formatters/PowerShellFormatter.cs
darrelmiller: I don't know.  This seems like a mistake to me.

Issue #1141, opened 2023-01-26
hidi doesn't return a failed exit code on validation error


baywet:
**Describe the bug**
hidi validate -d /some/invalid/path will result in a critical error in the console, however the exit code will be 0

**To Reproduce**
hidi validate -d /some/invalid/path will result in a critical error in the console, however the exit code will be 0

**Expected behavior**
exit code to be -1

MaggieKimani1:
Related #900

PR #1686, opened 2024-06-07
Return -1 exit code when the document is not valid

### LINKED ISSUE #900: Hidi returns exit code 0 even when failing on mac
Hello here is the log form my terminal:
```bash
bash-3.2$ hidi validate -d core/openapi.yml -ll trace
trce: Microsoft.OpenApi.Hidi.OpenApiService[0]
      => Reading input stream
      5ms: Read file core/openapi.yml
crit: Microsoft.OpenApi.Hidi.OpenApiService[0]
      OpenAPI specification version '' is not supported.
bash-3.2$ echo $?
0
bash-3.2$ hidi --version
1.0.0-preview4+b04aecd4a4955f71bc8f1a21ac74b8ac0976cb2a
```

The file I'm validating is just wrong for the purpose of demonstration, it starts like this:

```yaml
openapiWRONG: 3.0.3
...
```
Do you have an idea how I can get the correct exit code? It seems like it does not even print to the stderr but everything is printed to stdout.

KuraiAndras:
The previous related PR #1145 did return an exit code, but it did not use the result of the validation. This PR fixes this behavior.

This properly fixes #900 and #1141.

### DIFF
--- a/src/Microsoft.OpenApi.Hidi/Handlers/ValidateCommandHandler.cs
+++ b/src/Microsoft.OpenApi.Hidi/Handlers/ValidateCommandHandler.cs
@@ -33,8 +33,8 @@ namespace Microsoft.OpenApi.Hidi.Handlers
             try
             {
                 if (hidiOptions.OpenApi is null) throw new InvalidOperationException("OpenApi file is required");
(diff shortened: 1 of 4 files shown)


KuraiAndras:

MaggieKimani1 (approved):

PR #1698, opened 2024-06-24
Fixes null extension bug

### LINKED ISSUE #1695: extension parsing mechanism should not fail if the extension returns null
Imagine the following scenario:
1. I defined an extension with a key
2. A description contains an extension with the same key, but with a different structure
3. My extension code returns null, in the hope that generic parsing (OpenAPIAny) will kick in instead.
4. I get a null exception

This is caused by this block (and others similar to it)

https://github.com/microsoft/OpenAPI.NET/blob/c70b9a2653e2ff901a70bcaf1382aa1db3684469/src/Microsoft.OpenApi.Readers/V2/OpenApiV2Deserializer.cs#L114

Which should instead be

```csharp
            if (node.Context.ExtensionParsers.TryGetValue(name, out var parser) && parser(
                    OpenApiAnyConverter.GetSpecificOpenApiAny(node.CreateAny()),
                    OpenApiSpecVersion.OpenApi2_0) is {} result)
            {
                return result;
            }
            else
            {
                return OpenApiAnyConverter.GetSpecificOpenApiAny(node.CreateAny());
            }
```

Which ultimately triggers https://github.com/microsoft/OpenAPI.NET/blob/c70b9a2653e2ff901a70bcaf1382aa1db3684469/src/Microsoft.OpenApi/Extensions/OpenApiExtensibleExtensions.cs#L34

Once those implementations are updated, all the Microsoft defined extensions should also be updated to return null instead of throwing.
https://github.com/microsoft/OpenAPI.NET/blob/c70b9a2653e2ff901a70bcaf1382aa1db3684469/src/Microsoft.OpenApi/MicrosoftExtensions/OpenApiPagingExtension.cs#L76

MaggieKimani1:
Fixes #1695

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiV2Deserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiV2Deserializer.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
 using System.Collections.Generic;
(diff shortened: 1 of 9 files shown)


andrueastman (commented):
Any chance you can add tests/assertions for the changes for the relevant extensions in the tests?
This would probably be an inverse of the test at https://github.com/microsoft/OpenAPI.NET/blob/bb9e5b51a8860575a1de1a2ca923c12973ea30ae/test/Microsoft.OpenApi.Tests/MicrosoftExtensions/OpenApiReservedParameterExtensionTests.cs#L13

baywet (approved):
besides Andrew's comments, LGTM

andrueastman (approved):

baywet (approved):

Issue #1202, opened 2023-04-04
Enable null reference types on next major version


baywet:
[null reference type support](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) leverages the compiler to help us and  consumers avoid null reference exceptions.
Effectively this would be a breaking change and what would be required would be to:
- enable the feature in the csproj.
- annotate all properties and method parameters/return types with ? when a null value makes sense.
- check that we're not using null coalesce for strings, and instead string.IsNullOrEmpty in the code base
- solve other issues reported by the compiler

EDIT: the work for v2 is happening on `release/2.0.0`, for anybody looking to contribute here, please target that branch.

Bosch-Eli-Black:
Yes, please! 🙂  Our consuming project already leverages NRT's, and currently it's tricky to know which methods from the OpenAPI.NET NuGet might return `null`, since they're not annotated.

HavenDV:
I see there's a bit of an issue with this, perhaps it's worth implementing this as a gradual rollout for individual files using #nullable enable instead of trying to do everything at once?

HavenDV:
https://github.com/microsoft/OpenAPI.NET/pull/1705
I've created a small PR to slowly move this issue towards a solution, I'm hoping for quick feedback/merge so I can continue this
For now this is just marking up the code as is with no behavior changes/test edits.

PR #1705, opened 2024-06-26
feat: Added Nullability markup for some classes

HavenDV:
Related to #1202 

I think before that, #1202 should be split into two parts - marking the current code as is without changing the behavior
And after that, for the next big update, REQUIRED things should be replaced so that they are not null even in an empty document

This PR does not change behavior in places where null guards were provided or null was expected.
In places where there were NREs, ? were added. or other trivial fixes to avoid NRE
I'll continue with nullable markup if the change is accepted quickly enough

### DIFF
--- a/src/Microsoft.OpenApi.Hidi/Formatters/PowerShellFormatter.cs
+++ b/src/Microsoft.OpenApi.Hidi/Formatters/PowerShellFormatter.cs
@@ -52,7 +52,8 @@ namespace Microsoft.OpenApi.Hidi.Formatters
 
         public override void Visit(OpenApiPathItem pathItem)
         {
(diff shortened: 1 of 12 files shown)


baywet (commented):
Hi @HavenDV
Thanks for starting this! The work for the v2 is happening on `release/2.0.0`. Any chance you could rebase your branch (conflicts expected) and update the target of your PR please?

baywet:
This would be source breaking as anybody with an existing application with a dependency on the library would suddenly get compilation warning/errors (depending on the project configuration) instructing them to check for nullability at numerous places

HavenDV:
@baywet Done - https://github.com/microsoft/OpenAPI.NET/pull/1710

PR #1710, opened 2024-06-27
feat: Added Nullability markup for some classes

HavenDV:
Related to #1202 

I think before that, #1202 should be split into two parts - marking the current code as is without changing the behavior
And after that, for the next big update, REQUIRED things should be replaced so that they are not null even in an empty document

This PR does not change behavior in places where null guards were provided or null was expected.
In places where there were NREs, ? were added. or other trivial fixes to avoid NRE
I'll continue with nullable markup if the change is accepted quickly enough


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -16,6 +16,8 @@
 using Microsoft.OpenApi.Services;
 using Microsoft.OpenApi.Writers;
 
+#nullable enable
```
baywet, line 19: instead of relying on directives, could we enable that at the project level to get a consistent experience, and better end to end linting from the compiler?

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
```diff
@@ -19,60 +20,60 @@ public class OpenApiComponents : IOpenApiSerializable, IOpenApiExtensible
         /// <summary>
         /// An object to hold reusable <see cref="JsonSchema"/> Objects.
         /// </summary>
-        public IDictionary<string, JsonSchema> Schemas { get; set; } = new Dictionary<string, JsonSchema>();
+        public IDictionary<string, JsonSchema>? Schemas { get; set; } = new Dictionary<string, JsonSchema>();
```
baywet, line 23: for the properties which have a default value, should we instead avoid making them nullable? Doing so will require null propagation/checks all over (for consumers as well) and degrade the experience.

For the ones that don't, would it be worth adding a default value?

baywet (changes requested):

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
HavenDV: Then I will have to add nullable disable for all files and gradually turn it off

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
HavenDV: I tried, but if you look at the constructors, they still allow null values ​​to be set. And since I'm not changing the behavior, I made it null.
Also it breaks tests in some scenarios because there is a difference between null and empty value in yaml/json files

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
baywet: yes, I understand this represents much more work.

HavenDV: My main goal is to do this gradually, it will also be much easier to review/align with team changes

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
baywet: Do you mean the copy constructors? is that because of the null propagration?

HavenDV: Yes. I think we need to change the copy constructors so that they do not allow null right away.
But still not sure which of these properties can be null because as I wrote above here the difference is in json/yaml
Perhaps only those marked REQUIRED in the comments should be not null?

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
HavenDV: > Then I will have to add nullable disable for all files and gradually turn it off

Let's do this in next PR so it will be easier to do fast review

baywet: If you want to do it gradually maybe we can setup a branch feature/nullability (from release/2) here, and you can make multiple PRs to it. What I'm worried about is that we end up in a state where nullability is implemented halfway through, and can't publish public previews until the work is fully implemented because of side effects.

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
baywet: yes, I think that makes more sense, both from a standard implementation perspective and from a developer experience perspective.

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
HavenDV: I understand, but I can’t guarantee it. it will also imply constant conflicts that will have to be resolved, which will further slow down the process.
As far as I know, maui is marked in the same way, gradually. For me, it is better to have partial markings than not to have them at all.

baywet: Noted. I'm most likely not going to be the one doing the reviews anyway as it's outside of my core responsibilities, and I only followed up because I had originally created the issue.

I'd like @millicentachieng @coseguera and @MaggieKimani1 input on this and how they want to process this large batch of pull requests since they are the one most likely to review and merge your work at this point.

src/Microsoft.OpenApi/Models/OpenApiComponents.cs
MaggieKimani1: I believe @irvinesunday had already started some work on enabling null reference types https://github.com/microsoft/OpenAPI.NET/issues/1202

HavenDV: Judging by the pr it looks abandoned, there's a lot more work to be done here

darrelmiller:
We reviewed these changes and I think we understand the incremental step you are trying to take.  This is consistent with the direction that we want to go for the v2 release.  Thank you for making these changes.  Be aware that there are likely going to still be a fair amount of code churn in this branch, so the kind of changes that still need to be made may cause merge conflicts.

darrelmiller (approved):

HavenDV:
It seems we need to update System.Text.Json to 8.0.4 because 
```
/home/runner/work/OpenAPI.NET/OpenAPI.NET/src/Microsoft.OpenApi/Microsoft.OpenApi.csproj : error NU1903: Warning As Error: Package 'System.Text.Json' 7.0.2 has a known high severity vulnerability, https://github.com/advisories/GHSA-hh2w-p6rv-4g7w [/home/runner/work/OpenAPI.NET/OpenAPI.NET/Microsoft.OpenApi.sln]
/home/runner/work/OpenAPI.NET/OpenAPI.NET/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj : error NU1903: Warning As Error: Package 'System.Text.Json' 7.0.2 has a known high severity vulnerability, https://github.com/advisories/GHSA-hh2w-p6rv-4g7w [/home/runner/work/OpenAPI.NET/OpenAPI.NET/Microsoft.OpenApi.sln]
```
It doesn't allow to pass CICD. After the update, some of the tests fail, so this needs to be done in a separate PR and the tests fixed
Or I can add NoWarn for this temporarily.

baywet:
@HavenDV good call. 
Can you add the suppression in a similar way as this PR please? https://github.com/microsoft/kiota-dotnet/pull/292

And make sure we're using version ranges for library projects. 

Thanks a lot!

baywet:
> Quote reply
> Reference in ne

Actually this has already been handled in #1729

HavenDV:
Yes, I saw this so I decided not to risk making parallel changes here.
I'll have some free time today/weekend, I might take this further.

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -116,6 +118,7 @@
             Tags = document?.Tags != null ? new List<OpenApiTag>(document.Tags) : null;
             ExternalDocs = document?.ExternalDocs != null ? new(document?.ExternalDocs) : null;
             Extensions = document?.Extensions != null ? new Dictionary<string, IOpenApiExtension>(document.Extensions) : null;
+            BaseUri = document?.BaseUri != null ? document.BaseUri : new(OpenApiConstants.BaseRegistryUri + Guid.NewGuid().ToString());
```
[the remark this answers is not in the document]
baywet:
```suggestion
            BaseUri = document?.BaseUri != null ? document.BaseUri : new(OpenApiConstants.BaseRegistryUri + Guid.NewGuid());
```

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -97,13 +99,13 @@
         public OpenApiDocument() 
         {
             Workspace = new OpenApiWorkspace();
-            BaseUri = new(OpenApiConstants.BaseRegistryUri + Guid.NewGuid().ToString());            
+            BaseUri = new(OpenApiConstants.BaseRegistryUri + Guid.NewGuid().ToString());
```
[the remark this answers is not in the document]
baywet:
```suggestion
            BaseUri = new(OpenApiConstants.BaseRegistryUri + Guid.NewGuid());
```

src/Microsoft.OpenApi/Writers/OpenApiWriterExtensions.cs
```diff
@@ -137,7 +138,7 @@ public static void WriteProperty<T>(this IOpenApiWriter writer, string name, T v
         public static void WriteOptionalObject<T>(
             this IOpenApiWriter writer,
             string name,
-            T value,
+            T? value,
```
baywet, line 141: shouldn't this be replicated for the collection method?

HavenDV: Yes, but for now I've used #nullable enable/#nullable restore block to limit the amount of changes here because it's not required in this PR

MaggieKimani1 (approved):

HavenDV:
@baywet I'm not sure, but maybe it's blocked by "baywet requested changes" status. Is there anything else that needs fixing?

baywet (approved):
Sorry, I forgot to reset my review

darrelmiller:
@MaggieKimani1   Could we get this PR merged before we get further conflicts?

Issue #1715, opened 2024-07-03
Avoid use of unbounded reflection to resolve Enum display names


captainsafia:
In a few places in the OpenAPI.NET codebase, the `GetDisplayName` method is used to resolve a prettified display name associated with an enum value from the `[Display]` attribute.

https://github.com/microsoft/OpenAPI.NET/blob/3d971dea96e1d531f395d808743b0d1e863c1b23/src/Microsoft.OpenApi/Extensions/EnumExtensions.cs#L40

The use of unbounded reflection in this codepath makes it difficult to enable trimming for applications that use Microsoft.OpenApi in any capacity.

https://github.com/microsoft/OpenAPI.NET/blob/3d971dea96e1d531f395d808743b0d1e863c1b23/src/Microsoft.OpenApi/Extensions/EnumExtensions.cs#L24

Given the limited set of enums used in the application, I'd recommend an alternative reflection-free approach for resolving display names associated with enums. Something like a Enum <> string cache can be used to resolve this.

Assuming folks are OK with this approach, I'd be happy to submit a PR for review.

cc: @MaggieKimani1 @baywet

baywet:
Thanks for bringing this up. Do you think we could simply add the AOT attributes (`DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)` and whatnot ) and call it a day? This way the trimming would know not to trim the things we need here and we wouldn't have to implement a registry mechanism.

captainsafia:
Yep, that's an option. It does get a little bit interested since we're targeting `netstandard2.0`. The `DynamicallyAccessedMembers` aren't included for that target so you'll have to source include them yourself. There's guidance on how to do this in [this blog post](https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/#approach-2-define-the-attributes-internally).

So our two approaches would be:

- Remove reflection altogether and rely on a cached lookup.
- Keep reflection, explicitly define the `[DynamicallyAccessedMembers]` attributes in the source for OpenAPI.NET, and apply them to the relevant codepaths.

The first seemed cleaner to me, especially given that there is a finite set enums and I don't expect them to change much (as long as we are targeting OpenAPI v3.x)

captainsafia:
While digging into implementing this, I realized that the `CloneFromCopyConstructor` implementation in the library also presents a challenge for us making it fully trim compatible.

I _believe_ that the annotations referenced above will help with this but I have to give it a try.

[...]

Issue #1114, opened 2022-12-29
Make simple usage of `Microsoft.OpenAPI` trimmable


nb-rubenoost:
I'm trying to use this library in a console app. When trying to publish using trimming, I encountered the fact that this library is not trimmable for even the simplest of cases.

## **Is your feature request related to a problem? Please describe.**

The library is not trimmable when only the following functionality is used:

- Creating the model by hand through: `new OpenApiDocument { ... }`
- Serializing the model using `OpenApiJsonWriter` or `OpenApiYamlWriter`

Since these 2 steps don't feel like they are using reflection, I would expect this to be trimmable, which it is not. AFAIK, this is due to `EnumExtensions.GetDisplayName()`:
```
/_/src/Microsoft.OpenApi/Extensions/EnumExtensions.cs(27,13): Trim analysis warning IL2075: Microsoft.OpenApi.Extension
s.EnumExtensions.GetAttributeOfType<T>(Enum): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicCo
nstructors', 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.PublicFields', 'Dynamicall
yAccessedMemberTypes.PublicNestedTypes', 'DynamicallyAccessedMemberTypes.PublicProperties', 'DynamicallyAccessedMemberT
ypes.PublicEvents' in call to 'System.Type.GetMember(String)'. The return value of method 'System.Object.GetType()' doe
s not have matching annotations. The source value must declare at least the same requirements as those declared on the
target location it is assigned to.
```

## **Describe the solution you'd like**

In short, make `Microsoft.OpenAPI` trimmable for the simple scenario as described before (creating the model and serializing it).

The code that triggers this is:

```csharp
public static string GetDisplayName(this Enum enumValue)
{
    var attribute = enumValue.GetAttributeOfType<DisplayAttribute>();
    return attribute == null ? enumValue.ToString() : attribute.Name;
}

public static T GetAttributeOfType<T>(this Enum enumValue) where T : Attribute
{
    var type = enumValue.GetType();
    var memInfo = type.GetMember(enumValue.ToString()).First();
    var attributes = memInfo.GetCustomAttributes<T>(false);
    return attributes.FirstOrDefault();
}
```

This is only used to associate a string with an enum value. My proposed solution is:

[...]

PR #1717, opened 2024-07-04
Make OpenAPI.NET library trim-compatible

captainsafia:
This PR removes the use of reflection in `GetDisplayName` for enums and adds `[DynamicallyAccessMembers]` attributes to `CloneFromCopyConstructor`.

Prior to this change, using the `Microsoft.OpenApi` package in an application with trimming enabled would present the following warnings:

```
    /_/src/Microsoft.OpenApi/Extensions/EnumExtensions.cs(27): Trim analysis warning IL2075: Microsoft.OpenApi.Extensions.EnumExtensions.GetAttributeOfType<T>(Enum): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors', 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.PublicFields', 'DynamicallyAccessedMemberTypes.PublicNestedTypes', 'DynamicallyAccessedMemberTypes.PublicProperties', 'DynamicallyAccessedMemberTypes.PublicEvents' in call to 'System.Type.GetMember(String)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
    /_/src/Microsoft.OpenApi/Any/OpenApiAnyCloneHelper.cs(23): Trim analysis warning IL2075: Microsoft.OpenApi.Any.OpenApiAnyCloneHelper.CloneFromCopyConstructor(IOpenApiAny): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'System.Type.GetConstructors()'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```

Closes https://github.com/microsoft/OpenAPI.NET/issues/1715
Closes https://github.com/microsoft/OpenAPI.NET/issues/1114


src/Microsoft.OpenApi/Attributes/TrimmingAttributes.cs
```diff
@@ -0,0 +1,145 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+#nullable enable
+
+namespace System.Diagnostics.CodeAnalysis
```
captainsafia, line 6: The attributes that the trimmer uses to determine what should/shouldn't be trimmed aren't defined for the netstandard target. We're using the approach documented [here](https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/#approach-2-define-the-attributes-internally) to leverage them for our netstandard2.0 targeting packages.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
```diff
@@ -42,5 +43,107 @@ public static string GetDisplayName(this Enum enumValue)
...
             return attribute == null ? enumValue.ToString() : attribute.Name;
         }
+
+        /// <summary>
+        /// Gets the enum display name for <see typeparamref="T"/> without the use of reflection. 
+        /// </summary>
+        /// <typeparam name="T">The type of the enum value.</typeparam>
+        /// <param name="enumValue">The enum value.</param>
+        /// <returns>The display string to use.</returns>
+        public static string GetDisplayName<T>(this T enumValue) where T : Enum
```
captainsafia, line 53: I made this API public in keeping with the previous one, but I'd prefer to keep it internal. Is there a strong need for package users to be able to invoke this method?

src/Microsoft.OpenApi/Models/OpenApiReference.cs
```diff
@@ -83,7 +83,7 @@ public string ReferenceV3
                     return Id;
                 }
 
-                return "#/components/" + Type.GetDisplayName() + "/" + Id;
+                return "#/components/" + Type.Value.GetDisplayName() + "/" + Id;
```
captainsafia, line 86: Here and below, we unwrap the value from the nullable type to support the compiler in being able to resolve the `ReferenceType` in our generic `GetDisplayName` method.

src/Microsoft.OpenApi/Models/OpenApiReference.cs
```diff
@@ -201,7 +201,7 @@
                     return ExternalResource + "#" + Id;
                 }
 
-                return ExternalResource + "#/components/" + Type.GetDisplayName() + "/"+ Id;
+                return ExternalResource + "#/components/" + Type.Value.GetDisplayName() + "/"+ Id;
```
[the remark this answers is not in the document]
captainsafia: This warning is interesting. The code comments on the property say:

```
/// <remarks>This must be present if <see cref="ExternalResource"/> is not present.</remarks>
public ReferenceType? Type { get; set; }
```

But they don't say anything about the inverse: if `ExternalResource` is present, can `Type` be null? I assume in that case, we'd want to avoid this codepath altogether and just return `ExternalResource`?

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
```diff
@@ -42,5 +43,88 @@ public static string GetDisplayName(this Enum enumValue)
...
+        internal static string GetDisplayName(this ParameterStyle parameterStyle) => parameterStyle switch
+            {
+                ParameterStyle.Matrix => "matrix",
+                ParameterStyle.Label => "label",
+                ParameterStyle.Form => "form",
+                ParameterStyle.Simple => "simple",
+                ParameterStyle.SpaceDelimited => "spaceDelimited",
+                ParameterStyle.PipeDelimited => "pipeDelimited",
+                ParameterStyle.DeepObject => "deepObject",
+                _ => parameterStyle.ToString()
```
baywet, line 61: I'd argue we should not have default cases. This way anybody adding new members will get a compile time error to come fix things here.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
```diff
@@ -42,5 +43,88 @@ public static string GetDisplayName(this Enum enumValue)
...
+        /// </summary>
+        /// <param name="securitySchemeType">The enum value.</param>
+        /// <returns>The display string to use.</returns>
+        internal static string GetDisplayName(this SecuritySchemeType securitySchemeType) => securitySchemeType switch
+            {
+                SecuritySchemeType.ApiKey => "apiKey",
+                SecuritySchemeType.Http => "http",
+                SecuritySchemeType.OAuth2 => "oauth2",
+                SecuritySchemeType.OpenIdConnect => "openIdConnect",
+                _ => securitySchemeType.ToString()
```
baywet, line 127: don't you need equivalent parsing methods for the parsing scenarios?

src/Microsoft.OpenApi/Any/OpenApiAnyCloneHelper.cs
```diff
@@ -32,5 +32,27 @@ public static IOpenApiAny CloneFromCopyConstructor(IOpenApiAny obj)
 
             return obj;
         }
+        
+        /// <summary>
+        /// Clones an instance of <see cref="IOpenApiAny"/> object from the copy constructor
+        /// </summary>
+        /// <param name="obj">The object instance.</param>
+        /// <returns>A clone copy or the object itself.</returns>
+        public static IOpenApiAny CloneFromCopyConstructor<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T>(T obj) where T : IOpenApiAny
```
baywet, line 41: can you tag the non generic one as obsolete so we have a better idea about which one is binding in which scenario please?

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
```diff
@@ -42,5 +43,88 @@ public static string GetDisplayName(this Enum enumValue)
             var attribute = enumValue.GetAttributeOfType<DisplayAttribute>();
             return attribute == null ? enumValue.ToString() : attribute.Name;
         }
+
+        /// <summary>
+        /// Gets the enum display for name <see cref="ParameterStyle" /> without the use of reflection.
+        /// </summary>
+        /// <param name="parameterStyle">The enum value.</param>
+        /// <returns>The display string to use.</returns>
+        internal static string GetDisplayName(this ParameterStyle parameterStyle) => parameterStyle switch
```
baywet, line 52: should we mark the non type specific methods as obsolete?

src/Microsoft.OpenApi/Models/OpenApiReference.cs
baywet: could we have the extension methods bind to the nullable type instead?

src/Microsoft.OpenApi/Models/OpenApiReference.cs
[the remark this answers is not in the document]
baywet: If we could bind to nullable type instead it would probably solve this error

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
captainsafia: I can replace these default cases with `IOE` exceptions instead.

src/Microsoft.OpenApi/Models/OpenApiReference.cs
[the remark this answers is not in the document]
captainsafia: We could, but this feels like a logical gap to me. Given the lack of null-checking in the current codepath, it strikes me that there is a break in the code if ExternalResource = true and Type = null?

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
captainsafia: I took a look at the parsing logic and it looks like the only place we use `GetDisplayName` in the reader implementation is in `OpenApiParameterDeserializer`? With that in mind, I've make the `GetDisplayName` overload for `ParameterLocation` public to support that particular scenario.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
captainsafia: Done! This ended up catching some invocation sites that I missed.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
baywet: That'd be trading a compile time failure for a runtime one. I'm not sure it'd be better?

src/Microsoft.OpenApi/Models/OpenApiReference.cs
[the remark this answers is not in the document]
baywet: External references resolution has been a major issue with this version. We've revamped it entirely for the upcoming v2.
I'd be ok with leaving it as is and consider it legacy code as long as we're not breaking things further. 
I'd also like @maggiekimani1 input on this one

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
captainsafia: Hmmm....I'm not sure I understand what you are proposing? We need a default case here in order to comply with the pattern exhaustiveness checks on the switch expression. What kind of delta were you hoping to see here?

captainsafia: No worries -- your comment did help me see that I was missing a handle for `ReferenceType.Path` here. I added it and also added a set of unit tests to verify our exhaustiveness [here](https://github.com/microsoft/OpenAPI.NET/pull/1717/commits/8ea0964d6f3e1ab6ea70cb1b7e925b227513288a).

baywet: yes! I really like this unit tests approach. At least we don't run the risk of adding a new enum value and forgetting to update the serialization code. Resolving.

src/Microsoft.OpenApi/Models/OpenApiReference.cs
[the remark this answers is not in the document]
captainsafia: @MaggieKimani1 Any thoughts on this? Would love to move this PR forward and I think this is the last outstanding item 😄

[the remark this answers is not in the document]
MaggieKimani1: I agree with @baywet on this one. We've optimized both local and external reference resolution in the next major version so we can leave this as is pending further investigation.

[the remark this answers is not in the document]
captainsafia: OK -- I wasn't sure if "leave as is" meant that we want to keep the CodeQL warning around. I pused https://github.com/microsoft/OpenAPI.NET/pull/1717/commits/212840898f980b7b4214fcd6a329012693fde1a7 with a guard for the value to remove the warning.

If someone was running into this before, they would've encountered a NRE in the codepath. Now, we'll produce just a reference to the external resource and avoid mapping to a particular reference ID.

Let me know if this is not the behavior we want to encode.

baywet (commented):
I think it'd be interesting to set-up a trimming test project to ensure no regressions happen in the future here.
https://github.com/microsoft/kiota-dotnet/pull/291
https://github.com/microsoft/kiota-dotnet/pull/295
https://github.com/microsoft/kiota-dotnet/pull/296

darrelmiller (dismissed):

MaggieKimani1 (approved):

captainsafia:
Thanks for the reviews @darrelmiller and @MaggieKimani1 -- I spotted a few more issues after adding a TrimmingTest project (mostly related to ValidationRuleset) so we'll need to hold off on merging this a bit longer.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
```diff
@@ -21,6 +23,7 @@ public static class EnumExtensions
         /// <returns>
         /// The attribute of the specified type or null.
         /// </returns>
+        [RequiresUnreferencedCode("GetAttributeOfType is not trim-compatible. Recommended to use native AoT-friendly type-specific overloads of GetDisplayName instead.")]
```
eerhardt, line 26: If instead of `type.GetMember(enumValue.ToString())` this used `type.GetField(enumValue.ToString())`, it would fall into the case covered by https://github.com/dotnet/runtime/issues/97737. In [.NET 9 the trim warnings](https://github.com/dotnet/runtime/pull/100814/) changed to no longer warn on GetField on a Type guaranteed to be an enum.

If we made that change, until .NET 9, we could suppress the warning saying that the trimmer never trims fields of an enum. For example: https://github.com/dotnet/runtime/blob/ac0eeb015a0266adb06696bb58c21e5bfe6d9b56/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/EnumConverter.cs#L249-L250

cc @MichalStrehovsky

eerhardt, line 26: > Recommended to use native AoT-friendly type-specific overloads of GetDisplayName instead.

It seems like this recommendation doesn't apply when the caller is trying to get an attribute from the enum.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
```diff
@@ -37,10 +40,96 @@ public static class EnumExtensions
...
+                ParameterStyle.DeepObject => "deepObject",
+                _ => throw new InvalidOperationException($"Unknown parameter style: {parameterStyle}")
+            };
+
+        /// <summary>
+        /// Gets the enum display for name <see cref="ParameterLocation" /> without the use of reflection. 
+        /// </summary>
+        /// <param name="parameterLocation">The enum value.</param>
+        /// <returns>The display string to use.</returns>
+        public static string GetDisplayName(this ParameterLocation parameterLocation) => parameterLocation switch
```
eerhardt, line 73: Why are some of these methods internal, and some public?

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -185,17 +185,15 @@ IEnumerator IEnumerable.GetEnumerator()
...
-                .Where(t => t.IsClass
-                            && t != typeof(object)
-                            && t.GetCustomAttributes(typeof(OpenApiRuleAttribute), false).Any())
-                .SelectMany(t2 => t2.GetProperties(BindingFlags.Static | BindingFlags.Public)
-                                .Where(p => validationRuleType.IsAssignableFrom(p.PropertyType)));
-
-            foreach (var property in rules)
+            foreach (var property in ruleTypeProperties)
             {
+                if (!typeof(ValidationRule).IsAssignableFrom(property.PropertyType))
```
eerhardt, line 193: I'm not sure how much perf is a concern here (probably isn't), but I recently learned that `typeof(X)` doesn't get cached locally, so each time through this loop the Type gets re-lookedup. If perf is a concern, keeping a local variable for the Type helps.

test/Microsoft.OpenApi.Trimming.Tests/Microsoft.OpenApi.Trimming.Tests.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <PublishTrimmed>true</PublishTrimmed>
```
eerhardt, line 7:
```suggestion
    <PublishAot>true</PublishAot>
```

? That way you can get AOT warnings as well. Or are you only concerned with trimming?

test/Microsoft.OpenApi.Trimming.Tests/Microsoft.OpenApi.Trimming.Tests.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
eerhardt, line 1: Is this hooked up to an action / CI step? That would be ideal so if new warnings were introduced, the PR validation fails.

test/Microsoft.OpenApi.Trimming.Tests/Microsoft.OpenApi.Trimming.Tests.csproj
```diff
@@ -0,0 +1,17 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <PublishTrimmed>true</PublishTrimmed>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
eerhardt, line 8:
```suggestion
    <TrimmerSingleWarn>false</TrimmerSingleWarn>
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```

Adding `TrimmerSingleWarn=false` is nice because it will show all the warnings from referenced assemblies - and won't collapse them into a single warning for the whole assembly.

src/Microsoft.OpenApi/Any/OpenApiAnyCloneHelper.cs
```diff
@@ -32,5 +35,27 @@ public static IOpenApiAny CloneFromCopyConstructor(IOpenApiAny obj)
...
```
eerhardt, line 44:
```suggestion
        public static T CloneFromCopyConstructor<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] T>(T obj) where T : IOpenApiAny
```

Since this is now generic, it might make the API more ergonomic if the result was the same type as what was passed in.

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
captainsafia: The `BuildDefaultRuleSet` is called once and cached so the overall cost of this is pretty small but it doesn't hurt to revert this change for a little perf. 😄

test/Microsoft.OpenApi.Trimming.Tests/Microsoft.OpenApi.Trimming.Tests.csproj
captainsafia: Good tip!

test/Microsoft.OpenApi.Trimming.Tests/Microsoft.OpenApi.Trimming.Tests.csproj
baywet: Just added it to the solution so the CI validates this project as well. 4cc030d1

eerhardt: Just adding it to the solution isn't enough. It needs to be `dotnet publish`'d in order for the warnings to show.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
baywet: We should have an "InternalsVisibleTo" attribute somewhere already making this method visible to the other assembly.

baywet: We should have an "InternalsVisibleTo" attribute somewhere already making this method visible to the other assembly.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
baywet: I don't expect anyone other than us to have taken a dependency on this. Regardless, removing it would be a source breaking change for the dependents. We should add an obsolete attribute as well so we remove this with v2

MichalStrehovsky: `type.GetField(enumValue.ToString())` - also pass the binding flags to look for public static members so that we don't accidentally get the "secret" field holding the enum value.

Note that this will probably still warn in 9.0 because I did the `T : enum` handling but not a variable of type `System.Enum` case - that would require extra work. We can try to squeeze it into 9.0 if you think it would be valuable and file a bug on it.

eerhardt: > Note that this will probably still warn in 9.0 because I did the T : enum handling but not a variable of type System.Enum case - that would require extra work. We can try to squeeze it into 9.0 if you think it would be valuable and file a bug on it.

I never caught this before. Did you test the [case in the original issue](https://github.com/dotnet/runtime/issues/97737#issue-2109092800)?

```csharp
        private static string? GetEnumName<T>(T value) where T : Enum
        {
            var type = value.GetType();
...
            if(type.GetField(name)?.GetCustomAttribute<EnumMemberAttribute>() is { } attribute)
                return attribute.Value;
```

Note that this isn't using `typeof(T)` but instead `value.GetType()`. See [why this is important to that scenario](https://github.com/dotnet/runtime/issues/97737#issuecomment-1918250306).

That scenario and this current one seem to be the same - variable `Enum foo` => `foo.GetType().GetField("")`

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
eerhardt: The typical stance I've used for using IVT between product code is if they "ship together" and will never have mismatched versions. If 2 libraries ship in separate NuGet packages, I don't use IVT between them. One could get updated without the other, and if the internal API changed the assemblies won't load correctly.

baywet: I don't have a strong opinion either way (besides consistency between the methods maybe). I think the other rule of thumb I'd use in this scenario is: do we expect people to need this at any point? I don't think anybody is going to re-implement serialization and deserialization of OAI specifications in dotnet when they already have a reference to OAI.net. (maybe a different format, if even?) so using IVT should be fine here IMHO.

captainsafia: I'd be curious to see what @MaggieKimani1 and @darrelmiller are thinking here.

I'm personally starting to lean towards public API over IVT for the reasons that @eerhardt mentioned and the fact that that seems to be the pre-existing convention for this library (as annoying it is to have to make a single API public).

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
captainsafia: It doesn't, but based on @baywet's feedback in https://github.com/microsoft/OpenAPI.NET/pull/1717#discussion_r1681331142 we can obsolete this for now.

IMO, this is something that should be handled by the OpenAPI-layer in ASP.NET Core or other implementors, especially because things can get interesting depending on if you have `JsonStringEnumConverter` enabled.

src/Microsoft.OpenApi/Extensions/EnumExtensions.cs
eerhardt: Note that if we do what's outlined in https://github.com/microsoft/OpenAPI.NET/pull/1717#discussion_r1681224383, we can get rid of these new methods.

captainsafia: > Note that if we do what's outlined in [#1717 (comment)](https://github.com/microsoft/OpenAPI.NET/pull/1717#discussion_r1681224383), we can get rid of these new methods.

Based on https://github.com/dotnet/runtime/issues/97737#issuecomment-2233902387, it looks like we'll have to wait for a follow-up change to the runtime to support this. It seems like this should happen before preview.7 so I am fine waiting.

captainsafia: OK -- took me a while to catch up with the fact that this code was always trim-safe despite the warnings 😅 

I think https://github.com/microsoft/OpenAPI.NET/pull/1717/commits/c117bc56f17317853763b065aac943a1eedee9f7 does the right thing here.

captainsafia:
@baywet @eerhardt I think this is good for one last review...

test/Microsoft.OpenApi.Trimming.Tests/Microsoft.OpenApi.Trimming.Tests.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <Nullable>enable</Nullable>
+    <PublishAot>true</PublishAot>
+    <TrimmerSingleWarn>false</TrimmerSingleWarn>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.OpenApi\Microsoft.OpenApi.csproj" />
+    <TrimmerRootAssembly Include="Microsoft.OpenApi" />
```
baywet, line 15:
```suggestion
    <TrimmerRootAssembly Include="Microsoft.OpenApi" />
    <!-- TODO: include readers dependency when we enable trimming support in v2 -->
```

baywet (dismissed):
Thank you for this great contribution! And for the ongoing discussion here.

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
```diff
@@ -16,22 +17,21 @@ public static class StringExtensions
         /// Gets the enum value based on the given enum type and display name.
         /// </summary>
         /// <param name="displayName">The display name.</param>
-        public static T GetEnumFromDisplayName<T>(this string displayName)
+        [UnconditionalSuppressMessage("Trimming", "IL2075", Justification = "Fields are never trimmed for enum types.")]
```
eerhardt, line 20:
```suggestion
```

This shouldn't be needed, right? Since you have a `DynamicallyAccessedMembers` attribute on the `T`.

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
```diff
@@ -16,22 +17,21 @@ public static class StringExtensions
...
+        public static T GetEnumFromDisplayName<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] T>(this string displayName)
         {
             var type = typeof(T);
             if (!type.IsEnum)
             {
                 return default;
             }
 
-            foreach (var value in Enum.GetValues(type))
+            foreach (var field in type.GetFields(BindingFlags.Public | BindingFlags.Static))
```
eerhardt, line 29: Does this have any functional changes from the existing code? Specifically with a `Flags` enum?

test/Microsoft.OpenApi.Trimming.Tests/Program.cs
```diff
@@ -0,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
```
eerhardt, line 1: (nit)
```suggestion
```

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
captainsafia: AFAICT, behavior seems to be the same between this implementation and the prior one for `[Flags]` enums. I'll add test coverage for this. FWIW, primary consumers of this API are non-Flags enums in Microsoft.OpenApi.Readers although since it is public API nothing prohibits end-users from using it. :/

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
captainsafia: Yep -- we can remove this.

baywet (dismissed):

baywet:
@maggieKimani1 for final review and merge

test/Microsoft.OpenApi.Tests/Attributes/DisplayAttributeTests.cs
```diff
@@ -14,6 +15,17 @@ public enum ApiLevel
...
+    [Flags]
+    public enum UserType
+    {
+        [DisplayAttribute("admin")]
+        Admin = 1,
+        [DisplayAttribute("editor")]
+        Editor = 2,
+        [DisplayAttribute("publisher")]
+        Publisher = 3
+    }
```
eerhardt, line 27: (nit) It might be nice to also test:

```suggestion
        [DisplayAttribute("all")]
        All = Admin | Editor | Publisher
    }
```

eerhardt (approved):
LGTM. Thanks!

MaggieKimani1 (approved):

MaggieKimani1:
LGTM!

MaggieKimani1 (approved):

PR #1736, opened 2024-07-19
Fix copy constructor for arrays and objects

martincostello:
- Fix `OpenApiArray` and `OpenApiObject` not copying their items when they are copied to a new instance.
- Add tests for all built-in `IOpenApiAny` types with schema examples.

Found as part of https://github.com/dotnet/aspnetcore/issues/56318#issuecomment-2238704199.

### DIFF
--- a/src/Microsoft.OpenApi/Any/OpenApiAnyCloneHelper.cs
+++ b/src/Microsoft.OpenApi/Any/OpenApiAnyCloneHelper.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-using System.Reflection;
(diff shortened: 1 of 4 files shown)


MaggieKimani1 (approved):

MaggieKimani1:
Thank you for the contribution!

martincostello:
@MaggieKimani1 Do you know when the next stable release to NuGet.org will be? I'd like to start using the changes from this PR as well as #1717. If it isn't going to be soon, is there a pre-release feed someone I could use instead?

MaggieKimani1:
Hey @martincostello 
[v1.6.16](https://www.nuget.org/packages/Microsoft.OpenApi) is already available on NuGet with the above fixes. 🙂

Issue #1764, opened 2024-07-29
Parsing handlers for unknown properties and extensions


MaggieKimani1:

baywet:
Closing since this doesn't have a lot of details and was most likely already completed with the release of v2

PR #1766, opened 2024-07-30
Add JSON schema support

### LINKED ISSUE #1733: Add a JSON schema model class with all known keywords
Support all keywords highlighted here:
- https://github.com/microsoft/OpenAPI.NET/issues/1274
- https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/meta/base.schema.yaml

MaggieKimani1:
This PR:
- Adds a JSON schema model class with all known keywords
- Implements deserialization of V2, V3 and V31 OpenApi files into the in-memory Object model
- Implements serialization for V2, V3 and V31 keywords
- Adds tests

Fixes #1733, #1763, #1764, #1765

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -767,17 +767,18 @@ namespace Microsoft.OpenApi.Models
                 return type; // Return the string as is
             }
 
-            else
+            if (type is Array array)
             {
-                var array = type as Array;
                 Type elementType = type.GetType().GetElementType();
                 Array copiedArray = Array.CreateInstance(elementType, array.Length);
-                for (int i = 0; i < array.Length; i++)
+                for (int i = 0; i < array?.Length; i++)
                 {
-                    copiedArray.SetValue(DeepCloneType(array.GetValue(i)), i);
+                    copiedArray.SetValue(DeepCloneType(array?.GetValue(i)), i);
                 }
                 return copiedArray;
             }
+
+            return null;
         }
     }
 }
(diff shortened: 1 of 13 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -0,0 +1,795 @@
...
+        /// </summary>
+        public decimal? MultipleOf { get; set; }
+
+        /// <summary>
+        /// Follow JSON Schema definition: https://tools.ietf.org/html/draft-fge-json-schema-validation-00
+        /// The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided.
+        /// Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level.
+        /// For example, if type is string, then default can be "foo" but cannot be 1.
+        /// </summary>
+        public OpenApiAny Default { get; set; }
```
darrelmiller, line 151: Suggest using JsonNode here unless we need to use OpenApiAny.  If we do, then let's also change enum.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -0,0 +1,795 @@
...
+
+        /// <summary>
+        /// Follow JSON Schema definition: https://tools.ietf.org/html/draft-fge-json-schema-validation-00
+        /// </summary>
+        public IList<JsonNode> Enum { get; set; } = new List<JsonNode>();
+
+        /// <summary>
+        /// Allows sending a null value for the defined schema. Default value is false.
+        /// </summary>
+        public bool Nullable { get; set; }
```
darrelmiller, line 282: Nullable should be removed and represented as a member of the type array.

darrelmiller (approved):

Issue #1719, opened 2024-07-08
Support in-memory only metadata dictionary in OpenAPI object model


captainsafia:
**Is your feature request related to a problem? Please describe.**
Currently, all properties within the OpenAPI object model are serialized to the output as part of the OpenAPI writer implementation. This makes it difficult to manage temporary state associated with the in-memory OpenAPI model. For example, ASP.NET Core needs to correlate a given OpenApiOperation with a unique identifier for each endpoint recognized by the framework. 

**Describe the solution you'd like**
The OpenAPI object model should support some sort of property bag that allows users to manage metadata associated with an object model in-memory without having to worry about information in this property bag being serialized.

```
public class OpenApiHasPropertyBag
{
	IDictionary<string, object> MetadataCollection { get; }
}
```

**Describe alternatives you've considered**
Currently, ASP.NET Core works around this limitation by storing the metadata in the `Extensions` property on `OpenApiSchema` and `OpenApiOperation` then scrubbing this metadata out from the in-memory model before it is serialized. This presents a couple of issues:

- The additional cost of having to scrub out metadata that doesn't need to be serialized
- The additional cost of mapping property bag values to `OpenApiAny` implementers

captainsafia:
@baywet @MaggieKimani1 @darrelmiller Any thoughts on the API shape proposed here? I know the library makes extensive use of interfaces for things like this but it strikes me that using a base class would reduce the implementation overhead.

With regard to what classes will extend `OpenApiHasPropertyBag`, I'm thinking that we should start with:

- OpenApiDocument
- OpenApIPath
- OpenApiOperation
- OpenApiSchema

baywet:
From what I remember from this code base, having a base class would go against the general design.
The extensive use of interfaces for every facet has been the design here.
I wonder if interface default implementations could help us get the best of both here?

From a scenario perspective I'm not sure I understand what you'd expect OAS.net to do with that property bag besides storing it in memory?

captainsafia:
> I wonder if interface default implementations could help us get the best of both here?

Good idea! An interface with a default property implementation can help fill the gap here. My other conondrum was more bike-sheddy. A lot of the interfaces already defined have end in some sort of `ble` suffix (Extensible, Serializable, Referenceable). I couldn't think of a good one for this but I'm open to ideas.

> From a scenario perspective I'm not sure I understand what you'd expect OAS.net to do with that property bag besides storing it in memory?

[...]

PR #1769, opened 2024-08-02
Add support for unserializable annotations on OpenAPI document

captainsafia:
Addresses https://github.com/microsoft/OpenAPI.NET/issues/1719.

This PR adds support for unserizable property bags to the `OpenApiDocument,`OpenApiOperation`, and `OpenApiSchema` to support managing in-memory metadata about a given OpenAPI instance without having these properties embedded into the serialized document.

- I landed on `Annotations` and `IOpenApiAnnotatable` as the names used in this API to align with the "ble" naming style used elsehwere. Open to other name ideas.
- `Annotations` are not eagerly initialized by the API to avoid allocating them when not necessary. It's up to the end-user to initialize the dictionary before appending elements to it.
- The interface is applied only to the `OpenApiDocument`, `OpenApiOperation`, and `OpenApiSchema` types at the moment to support ASP.NET Core's scenarios. We can evaluate what other types it makes sense to apply this interface to.
- The `Annotatations` property uses the same copy semantics used by other Dictionary properties in the copy constructor.

### DIFF
--- a/src/Microsoft.OpenApi/Interfaces/IOpenApiAnnotatable.cs
+++ b/src/Microsoft.OpenApi/Interfaces/IOpenApiAnnotatable.cs
@@ -1,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+
+namespace Microsoft.OpenApi.Interfaces
+{
+    /// <summary>
+    /// Represents an Open API element that can be annotated with
+    /// non-serializable properties in a property bag. 
+    /// </summary>
+    public interface IOpenApiAnnotatable
+    {
+        /// <summary>
+        /// A collection of properties associated with the current OpenAPI element.
+        /// </summary>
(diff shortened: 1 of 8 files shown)


MaggieKimani1 (approved):

captainsafia:
@MaggieKimani1 @darrelmiller Any change we can merge and publish this to v1.6.x soon? I'd like to update the `Microsoft.OpenApi` dependency in ASP.NET Core before we ship RC1.

MaggieKimani1:
> @MaggieKimani1 @darrelmiller Any change we can merge and publish this to v1.6.x soon? I'd like to update the `Microsoft.OpenApi` dependency in ASP.NET Core before we ship RC1.

Hey @captainsafia,
I think this is ready for merging. I can work on getting out a release for you tomorrow.

MaggieKimani1:
@captainsafia `v1.6.17` is now GA.🙂

captainsafia:
@MaggieKimani1 Thanks! I've updated ASP.NET Core to use the new version. We'll be going live with it for .NET 9 RC1.

PR #1783, opened 2024-08-19
Add server variable substitution logic.

### LINKED ISSUE #1781: The server URL builder (ParseServerUrl) should allow users to provide alternate values for variables
According to the OpenAPI spec (v3.0), the server variable object should allow alternate values and fall back to the default value if an alternate value isn't specified. Currently, the [ParseServerUrl](https://github.com/microsoft/OpenAPI.NET/blob/c345fc7496ee5d63fd754d4aa379cf094d14eee3/src/Microsoft.OpenApi/Models/OpenApiDocument.cs#L284-L294) function doesn't allow providing alternate values for the variables.

The function should allow providing alternate values, for example through a dictionary.

N/A

**Additional context**
image

calebkiage:
fixes #1781


src/Microsoft.OpenApi/Extensions/OpenApiServerExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+                throw new ArgumentException(
+                    string.Format(SRResource.ParseServerUrlDefaultValueNotAvailable, variable.Key), nameof(server));
+            }
+
+            if (variable.Value.Enum is { Count: > 0 } e && !e.Contains(value))
+            {
+                throw new ArgumentException(
+                    string.Format(SRResource.ParseServerUrlValueNotValid, value, variable.Key), nameof(values));
+            }
+
```
MaggieKimani1, line 49: Can we also add validation for Enums being empty which is disallowed by the spec?
Something like:
```suggestion
if (!variable.Value.Enum.Any())
{
     // throw exception
}
```

MaggieKimani1 (approved):

PR #1803, opened 2024-08-29
Copy over references for all IOpenApiReferenceable objects

### LINKED ISSUE #1799: The OpenApiFilter service does not copy all references from the original document.
Related to https://github.com/microsoft/kiota/issues/5263

Using the OpenApiFilter services does not copy all references in the original document resulting in some missing refs/invalid document.

Taking a look at the visitor at https://github.com/microsoft/OpenAPI.NET/blob/9a5a41521a5650cfecdee799f23266fc5da8cff3/src/Microsoft.OpenApi/Services/CopyReferences.cs#L26

The only references/components copied over are of type schemas,parameters,requestBodies and responses. This leads to leaving out other components that could be referenced like examples/headers/callbacks from the components. See properties at https://github.com/microsoft/OpenAPI.NET/blob/vnext/src/Microsoft.OpenApi/Models/OpenApiComponents.cs#L14

MaggieKimani1:
Fixes #1799

### DIFF
--- a/src/Microsoft.OpenApi/Services/CopyReferences.cs
+++ b/src/Microsoft.OpenApi/Services/CopyReferences.cs
@@ -174,12 +174,12 @@ namespace Microsoft.OpenApi.Services
 
         private void EnsureLinksExist()
         {
-            _target.Components.Callbacks ??= new Dictionary<string, OpenApiCallback>();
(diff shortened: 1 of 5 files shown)


darrelmiller (dismissed):

andrueastman (approved):
👍🏼

PR #1809, opened 2024-08-31
Use ConcurrentDictionary For Improving GetEnumFromDisplayName

Mahdigln:


darrelmiller:
Is this change for perf reasons?  Do you have any benchmarks for what the perf difference is between these solutions?

Mahdigln:
> Is this change for perf reasons? Do you have any benchmarks for what the perf difference is between these solutions?

I ran the benchmark with each method executed 1,000,000 times to simulate a high-usage scenario. This large number of iterations helps highlight the performance difference between the two approaches. Here's how the times break down:

    Without cache: 2230 ms for 1,000,000 executions.
    With cache: 63 ms for 1,000,000 executions.

This shows the significant advantage of using caching when dealing with repeated reflection operations

darrelmiller:
That seems like a pretty significant improvement.  Thank you for the work.

darrelmiller (approved):

Mahdigln:
> That seems like a pretty significant improvement. Thank you for the work.

Glad it was helpful! Hope to contribute more in the future.

PR #1815, opened 2024-09-04
Disable VisualStudio.Threading analyzers to eliminate build failures downstream

### LINKED ISSUE #1810: v1.6.19 adds dependency on Microsoft.VisualStudio.Threading analyzers
**Describe the bug**

Version 1.6.19 of the Microsoft.OpenApi packages introduce a dependency on Microsoft.VisualStudio.Threading, which in turn is adding analyzers from that package into dependent projects.

This then causes build failures when the consuming project does not conform to any of the rules the analyzer introduces.

For example: https://github.com/martincostello/api/pull/1967

```
Use "Async" suffix in names of methods that return an awaitable type (https://github.com/Microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD200.md)
```

**OpenApi File To Reproduce**

N/A

**Expected behavior**

Analyzers from this project do not affect the consumers of Microsoft.OpenApi packages.

**Screenshots/Code Snippets**

N/A

**Additional context**

None.

MaggieKimani1:
Fixes #1810

### DIFF
--- a/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
+++ b/src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
@@ -3,7 +3,7 @@
         <TargetFramework>netstandard2.0</TargetFramework>
         <LangVersion>latest</LangVersion>
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-        <Version>1.6.19</Version>
+        <Version>1.6.20</Version>
         <Description>OpenAPI.NET Readers for JSON and YAML documents</Description>
         <SignAssembly>true</SignAssembly>
         <!-- https://github.com/dotnet/sourcelink/blob/main/docs/README.md#embeduntrackedsources -->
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi.Readers/Microsoft.OpenApi.Readers.csproj
```diff
@@ -22,6 +22,13 @@
         <PackageReference Include="SharpYaml" Version="2.1.1" />
     </ItemGroup>
 
+    <!-- Workaround to disable Microsoft.VisualStudio.Threading.Analyzers -->
```
andrueastman, line 25: I don't think we need to this workaround, we just need to use the `privateAssets` tag with the dependency similar to https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/7312b38e76767f18c1e9ae69e3bc76285d98f1d0/src/Microsoft.Graph.Core/Microsoft.Graph.Core.csproj#L75

https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

andrueastman (approved):

PR #1820, opened 2024-09-05
Handle upcasting and downcasting of JSON schemas with type arrays

### LINKED ISSUE #1786: Handle upcast and downcast of JSON Schemas with multiple types
In OpenAPI 3.1, JSON Schema can have a type that is an array.

- If the array has only one value that is not null, treat as if the type was just a string
- If the array has only one value that is null, in v3/v2 emit as nullable/x-nullable and emit no type
- If the array has two values where one is null, then in v3/v2 emit as nullable/x-nullable and convert type to string of other value.
- If the array has two values where neither is null, or more than two values, then in v3/v2 do not emit the type field.  By omitting the type field we keep the schema from creating false negatives and it is more likely that downstream tooling will flag the issue.

MaggieKimani1:
Fixes #1786

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -553,10 +553,6 @@ namespace Microsoft.OpenApi.Models
             writer.WriteProperty(OpenApiConstants.Deprecated, Deprecated, false);
 
             // extensions
-            if (Extensions.ContainsKey(OpenApiConstants.NullableExtension))
-            {
-                Extensions.Remove(OpenApiConstants.NullableExtension);
-            }
             writer.WriteExtensions(Extensions, OpenApiSpecVersion.OpenApi3_0);
 
             writer.WriteEndObject();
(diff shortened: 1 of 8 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -557,6 +553,10 @@ public void SerializeInternal(IOpenApiWriter writer, OpenApiSpecVersion version,
             writer.WriteProperty(OpenApiConstants.Deprecated, Deprecated, false);
 
             // extensions
+            if (Extensions.ContainsKey(OpenApiConstants.NullableExtension))
```
darrelmiller, line 556: I do not believe we should making any changes to the document when writing.  If a developer chooses to add an x-Nullable extension in a 3.1 version document, then so be it. We should write it out as they requested.

darrelmiller: We should however handle removing the x-Nullable and Nullable during reading.

darrelmiller (changes requested):
I think if you remove the code that removes the x-nullable this PR looks good.

PR #1888, opened 2024-10-23
Remove DataTypeMismatch validation rule from the default ruleset

### LINKED ISSUE #1879: ValidateDataTypeMismatch is failing to parse examples
**Describe the bug**

dotnet run -- validate -d https://raw.githubusercontent.com/documentaciononyxsoft/documentacion/refs/heads/master/ApiIntervention.yaml
crit: Microsoft.OpenApi.Hidi.Handlers.ValidateCommandHandler[0]
      Command failed
Unhandled exception: System.InvalidOperationException: Could not validate the document, reason: Unable to cast object of type 'System.String[]' to type 'System.String'.
 ---> System.InvalidCastException: Unable to cast object of type 'System.String[]' to type 'System.String'.
   at Microsoft.OpenApi.Validations.Rules.RuleHelpers.ValidateDataTypeMismatch(IValidationContext context, String ruleName, JsonNode value, OpenApiSchema schema) in D:\github\microsoft\openapi.net\src\Microsoft.OpenApi\Validations\Rules\RuleHelpers.cs:line 58

**OpenApi File To Reproduce**
https://raw.githubusercontent.com/documentaciononyxsoft/documentacion/refs/heads/master/ApiIntervention.yaml

**Expected behavior**
I don't expect ValidateDataTypeMismatch. I was expecting that we had removed this rule from the default set of rules.

MaggieKimani1:
- Removes the `DataTypeMismatch` validation rule from the default ruleset by isolating it in a separate class to allow clients to opt in if need be.
Fixes #1879


andrueastman (commented):
Out of curiosity, do we want the rule completely removed from the code? 
Or do we want the rule to not be in the default set of rules (meaning a caller of the library could opt in to it if they so wished)?

MaggieKimani1:
> Out of curiosity, do we want the rule completely removed from the code? Or do we want the rule to not be in the default set of rules (meaning a caller of the library could opt in to it if they so wished)?

Oh actually yes, the intent is to remove it from the default set of rules and allow users to opt in. Let me correct that.

darrelmiller (approved):

PR #1897, opened 2024-10-28
Define JSON schema's Type property as a flaggable enum to allow storing multiple values

### LINKED ISSUE #1874: V2 - Why is OpenAPISchema.Type of type "object"?
This provides a really poor development experience, forcing the consumer to implement complex type assertions when it shouldn't be required.

And then we do things like those: https://github.com/microsoft/OpenAPI.NET/blob/3c644511afba381b0c580f31d7282ac1e98c8d41/src/Microsoft.OpenApi/Models/OpenApiSchema.cs#L839

Which is not only going to be slow, but also problematic for people using AOT/trimming.

We [already know](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-4.2.1) that OAS 3.1 and Json Schema only support the following types:
- null
- boolean
- object
- array
- number
- string

(maybe we could add compatibility with the integer mess).

Why not instead:
- Define an enum for those types
- Make the type property, an array/list/enumerable of that enum
- When serializing as v2/v3, ensuring that we emit the right compatible information.

MaggieKimani1:
Fixes #1874


src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -12,40 +12,81 @@ namespace Microsoft.OpenApi.Extensions
...
+            {
+                "null" => JsonSchemaType.Null,
+                "boolean" => JsonSchemaType.Boolean,
+                "integer" or "int" => JsonSchemaType.Integer,
+                "number" or "double" => JsonSchemaType.Number,
+                "string" => JsonSchemaType.String,
+                "array" => JsonSchemaType.Array,
+                "object" => JsonSchemaType.Object,
+                "file" => JsonSchemaType.String, // File is treated as string
+                _ => JsonSchemaType.Any,
```
darrelmiller, line 52: Remove JsonSchemaType.Any  and in this case throw an exception.  It should not stop parsing, but convert to a problem. I think this happens in ParseNode.

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
```diff
@@ -122,9 +122,21 @@ internal static partial class OpenApiV31Deserializer
...
-                        ? n.GetScalarValue()
-                        : n.CreateSimpleList((n2, p) => n2.GetScalarValue()).ToArray();
+                    if (n is ValueNode)
+                    {
+                        o.Type = OpenApiTypeMapper.IdentifierToEnumType(n.GetScalarValue());
+                    }
+                    else
+                    {
+                        var list = n.CreateSimpleList((n2, p) => n2.GetScalarValue());
+                        JsonSchemaType combinedType = JsonSchemaType.Any;
```
darrelmiller, line 132: Remove this use of Any.

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -12,40 +13,81 @@ namespace Microsoft.OpenApi.Extensions
     /// </summary>
     public static class OpenApiTypeMapper
     {
+        /// <summary>
+        /// Maps a JsonSchema data type to an identifier.
+        /// </summary>
+        /// <param name="schemaType"></param>
+        /// <returns></returns>
+        public static string ToIdentifier(JsonSchemaType? schemaType)
```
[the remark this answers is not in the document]
MaggieKimani1: Yeah this provides a cleaner experience. Let me implement it.

darrelmiller (approved):

andrueastman (approved):

PR #1915, opened 2024-11-12
fix: multiple performance fixes for type serialization feat: adds to identifier mapping to non nullable enum

baywet:


MaggieKimani1:
I understand that using a bitwise operation to determine the number of flags set on `JsonSchemaType` is more efficient (`O(1)`) than using the iterative count (`O(n)`) as I had implemented. However the performance gain seems minimal if JsonSchemaType contains only a few flags (e.g., up to 3).

baywet:
It's not just the loop, it's also:
- Enum.GetValues needs to go through the type system to get the information, which is much slower\
- Assigning count multiple times is slower than a single bitwise instruction

(besides the other improvements in terms of "caching" the enum values, avoiding auto-boxing, etc...

MaggieKimani1 (approved):

Issue #1918, opened 2024-11-12
v2 - loading the document requires knowing the format in advance


baywet:
```chsarp
new OpenApiStreamReader().Read(stream, out var diagnostic);
```

(or its async variant)

Did not require knowing the format of the document (JSON/YAML) in advance. This was handy as when building general purpose tools, it's not something that I know in advance. Parsing the file extension/url extension/response content type header _might_ help discovering it, but it's not reliable.

The equivalent method

```csharp
OpenApiDocument.LoadAsync(input, "unknown", settings)
```

Requires passing the format, which I believe is a regression in functionality. We should work to restore the same level of functionality.

Also, there are a number of places in the documentation and doc comments that still refer to the old initialization and require to be updated before GA.

https://github.com/search?q=repo%3Amicrosoft%2FOpenAPI.NET%20OpenApiStreamReader&type=code

darrelmiller:
We could create an overload that looks like this:

> OpenApiDocument.LoadAsync(input, settings);

We first try and parse as JSON.  If the JSON fails, we can say "either you have bad JSON, or you gave me YAML, and you have not registered the YAML reader, in which case you need to take a dependency on the OpenAPI.Readers package, and register the YAML reader."

1. Try and parse as JSON
2. If it fails, check to see if YAML parser is registered
3. If it is try parsing with the YAML parser

baywet:
Thank you for the additional information.

I don't understand why we to direct the parsing between two different parsing logics: 

> The YAML 1.2[3](https://yaml.org/spec/1.2.2/#fn:1-2-spec) specification was published in 2009. Its primary focus was making YAML a strict superset of JSON

[source](https://yaml.org/spec/1.2.2/#12-yaml-history)

We could simply feed everything to the YAML parser, couldn't we? It it happens to "only be JSON" there's extra parsing logic that would be used. It might have a performance impact since [yaml parsing ultimately converts back to JSON nodes](https://github.com/microsoft/OpenAPI.NET/blob/1338905859eaa083f89a8819227073feef8836a3/src/Microsoft.OpenApi.Readers/OpenApiYamlReader.cs#L81).

It's also worth noting [this is what the library used to do before the recent changes](https://github.com/microsoft/OpenAPI.NET/blob/9ff9789fff842e9e567dbe70a8a7966014ce8e3d/src/Microsoft.OpenApi.Readers/OpenApiTextReaderReader.cs#L39).

[...]

PR #1929, opened 2024-11-14
Fix: loading the document requires knowing the format in advance

### LINKED ISSUE #1917: v2 - We should NOT do sync over async
6 instances being suppressed 

https://github.com/search?q=repo%3Amicrosoft%2FOpenAPI.NET%20VSTHRD002%20&type=code

This will cause poor performance and potential dead locks on backend services.

One of those instances is marked as obsolete, we should simply clean it up (and make a pass to ensure we remove any other obsolete API)

The five others were introduced with this preview. We should remove those methods, and let the consumer do the blocking if they absolutely need to with the proper patterns in terms of thread and context management

MaggieKimani1:
This PR:
- Adds logic for inspecting the first character of the stream we receive during document loading to know whether its in JSON or YAML. This helps avoid a regression in functionality where users are required to pass the format of the document when loading an OpenAPI document
- Replaces sync methods with async ones to prevent deadlocks
- Passes a cancellation token to `OpenApiModelFactory.GetStreamAsync` to allow for cancellation of tasks/requests
- Updates tests with new loading pattern

Fixes #1917, #1918

### DIFF
--- a/src/Microsoft.OpenApi.Readers/OpenApiYamlReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiYamlReader.cs
@@ -46,10 +46,9 @@ namespace Microsoft.OpenApi.Readers
         }
 
         /// <inheritdoc/>
-        public T ReadFragment<T>(TextReader input,
-                                 OpenApiSpecVersion version,
-                                 out OpenApiDiagnostic diagnostic,
-                                 OpenApiReaderSettings settings = null) where T : IOpenApiElement
+        public async Task<ReadFragmentResult> ReadFragmentAsync<T>(TextReader input,
+                                                                   OpenApiSpecVersion version,
+                                                                   OpenApiReaderSettings settings = null) where T : IOpenApiElement
         {
             JsonNode jsonNode;
 
@@ -60,12 +59,12 @@ namespace Microsoft.OpenApi.Readers
             }
             catch (JsonException ex)
             {
-                diagnostic = new();
+                var diagnostic = new OpenApiDiagnostic();
                 diagnostic.Errors.Add(new($"#line={ex.LineNumber}", ex.Message));
                 return default;
             }
 
-            return ReadFragment<T>(jsonNode, version, out diagnostic);
+            return await ReadFragmentAsync<T>(jsonNode, version, settings);
         }
 
         /// <summary>
@@ -88,9 +87,9 @@ namespace Microsoft.OpenApi.Readers
         }
 
         /// <inheritdoc/>
-        public T ReadFragment<T>(JsonNode input, OpenApiSpecVersion version, out OpenApiDiagnostic diagnostic, OpenApiReaderSettings settings = null) where T : IOpenApiElement
+        public async Task<ReadFragmentResult> ReadFragmentAsync<T>(JsonNode input, OpenApiSpecVersion version, OpenApiReaderSettings settings = null) where T : IOpenApiElement
         {
-            return OpenApiReaderRegistry.DefaultReader.ReadFragment<T>(input, version, out diagnostic);
+            return await OpenApiReaderRegistry.DefaultReader.ReadFragmentAsync<T>(input, version, settings);
         }
     }
 }
--- a/src/Microsoft.OpenApi.Readers/OpenApiYamlReader.cs
+++ b/src/Microsoft.OpenApi.Readers/OpenApiYamlReader.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
 using System.IO;
@@ -47,7 +47,7 @@ namespace Microsoft.OpenApi.Readers
         /// <inheritdoc/>
         public async Task<ReadFragmentResult<T>> ReadFragmentAsync<T>(TextReader input,
                                                                       OpenApiSpecVersion version,
-                                                                      OpenApiReaderSettings settings = null) where T : IOpenApiElement
+                                                                      OpenApiReaderSettings settings = null,
                                                                       CancellationToken token = default) where T : IOpenApiElement
         {
             JsonNode jsonNode;
@@ -55,7 +55,7 @@ namespace Microsoft.OpenApi.Readers
             // Parse the YAML
             try
             {
-                jsonNode = LoadJsonNodesFromYamlDocument(input);
+                jsonNode = await Task.Run(() => LoadJsonNodesFromYamlDocument(input));
             }
             catch (JsonException ex)
             {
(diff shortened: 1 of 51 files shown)


test/Microsoft.OpenApi.Tests/Models/References/OpenApiHeaderReferenceTests.cs
```diff
@@ -82,8 +82,8 @@ public class OpenApiHeaderReferenceTests
         public OpenApiHeaderReferenceTests()
         {
             OpenApiReaderRegistry.RegisterReader(OpenApiConstants.Yaml, new OpenApiYamlReader());
-            _openApiDoc = OpenApiDocument.Parse(OpenApi, OpenApiConstants.Yaml).OpenApiDocument;
-            _openApiDoc_2 = OpenApiDocument.Parse(OpenApi_2, OpenApiConstants.Yaml).OpenApiDocument;
+            _openApiDoc = OpenApiDocument.ParseAsync(OpenApi).GetAwaiter().GetResult().OpenApiDocument;
```
baywet, line 85: this test and all the others below: switch to async lifetime, like the example I pushed for **OpenApiExampleReferenceTests**

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
```diff
@@ -162,8 +157,8 @@ public T ReadFragment<T>(JsonNode input,
             IOpenApiElement element = null;
             try
             {
-                // Parse the OpenAPI element
-                element = context.ParseFragment<T>(input, version);
+                // Parse the OpenAPI element asynchronously
+                element = await Task.Run(() => context.ParseFragment<T>(input, version));
```
baywet, line 161: why the extra Task.Run here?

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
```diff
@@ -180,16 +175,20 @@ public T ReadFragment<T>(JsonNode input,
...
+                OpenApiDiagnostic = diagnostic
+            };
         }
 
-        private JsonNode LoadJsonNodes(TextReader input)
+        private async Task<JsonNode> LoadJsonNodesAsync(TextReader input)
         {
-            var nodes = JsonNode.Parse(input.ReadToEnd());
-            return nodes;
+            var content = await input.ReadToEndAsync();
```
baywet, line 187: conditional compilation to pass the cancellation token

src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
```diff
@@ -183,90 +103,83 @@ public static async Task<ReadResult> ParseAsync(string input,
...
-        private static string GetContentType(string url)
+        private static async Task<string> GetContentTypeAsync(string url)
         {
             if (!string.IsNullOrEmpty(url))
             {
-#pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
-                var response = _httpClient.GetAsync(url).GetAwaiter().GetResult();
-#pragma warning restore VSTHRD002 // Avoid problematic synchronous waits
-
+                var response = await _httpClient.GetAsync(url);
```
baywet, line 179: pass a cancellation token

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
MaggieKimani1: The `context.ParseFragment` method is synchronous hence the Task.Run to run it async and avoid blocking the calling thread

baywet: right, but all of the work this method is doing is CPU bound? (no IOs)

src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
```diff
@@ -297,6 +219,100 @@ public static string GetFormat(string url)
...
+        private static string InspectTextReaderFormat(TextReader reader)
+        {
+            // Read the first line or a few characters from the input
+            var input = reader.ReadLine().Trim();
+            return InspectInputFormat(input);
+        }
+
+        private static string InspectInputFormat(string input)
+        {
+            return input.StartsWith("{") || input.StartsWith("[") ? OpenApiConstants.Json : OpenApiConstants.Yaml;
```
baywet, line 313:
```suggestion
            return input.StartsWith("{", StringComparison.OrdinalIgnoreCase) || input.StartsWith("[", StringComparison.OrdinalIgnoreCase) ? OpenApiConstants.Json : OpenApiConstants.Yaml;
```

(it'd be better to use the char overload, but since this is still targeting netstandard 2.0, it's not available)

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
MaggieKimani1: Yes you're right. Let me refactor that to remove the async operation.

src/Microsoft.OpenApi.Readers/OpenApiYamlReader.cs
```diff
@@ -47,14 +47,15 @@ public async Task<ReadResult> ReadAsync(TextReader input,
...
+                                                                      OpenApiReaderSettings settings = null,
+                                                                      CancellationToken token = default) where T : IOpenApiElement
         {
             JsonNode jsonNode;
 
             // Parse the YAML
             try
             {
-                jsonNode = LoadJsonNodesFromYamlDocument(input);
+                jsonNode = await Task.Run(() => LoadJsonNodesFromYamlDocument(input));
```
baywet, line 58: why are we adding the task run here?

MaggieKimani1: The caller is async here https://github.com/microsoft/OpenAPI.NET/blob/f07ff446a2833e09a6617ae80d534bab4b56fd68/src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs#L147

And the interface method's equivalent in the JsonReader class is also async https://github.com/microsoft/OpenAPI.NET/blob/f07ff446a2833e09a6617ae80d534bab4b56fd68/src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs#L124-L143

If I change this method signature to be synchronous, I'll also need to update this JSON content to be read synchronously:
https://github.com/microsoft/OpenAPI.NET/blob/f07ff446a2833e09a6617ae80d534bab4b56fd68/src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs#L189

MaggieKimani1:
@baywet just a head's up, the functionality on inspecting the input stream to determine the format might change based on my discussion with @darrelmiller here https://github.com/microsoft/OpenAPI.NET/issues/1918#issuecomment-2485894218

baywet:
@maggiekimani1 are we going to implement those changes in this PR or in a subsequent one?

MaggieKimani1:
> @MaggieKimani1 are we going to implement those changes in this PR or in a subsequent one?

I can implement it here so as not to lose the other changes on moving to async

PR #1982, opened 2024-12-10
Add support for handling unrecognized keywords

### LINKED ISSUE #1970: [API Review] : Implement support for handling unrecognized keywords in the OpenAPISchema
We will error out if we don't see any recognized keyword in `OpenAPISchema`, so we should have a way to support unrecognized keywords in json-schema.

MaggieKimani1:
Fixes #1970

### DIFF
--- a/src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
+++ b/src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
@@ -7,6 +7,8 @@ using Microsoft.OpenApi.Models.References;
 using Microsoft.OpenApi.Reader.ParseNodes;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
+using System.Text.Json.Nodes;
 
 namespace Microsoft.OpenApi.Reader.V31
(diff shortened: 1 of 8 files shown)


src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
```diff
@@ -254,7 +256,17 @@ public static OpenApiSchema LoadSchema(ParseNode node, OpenApiDocument hostDocum
 
             foreach (var propertyNode in mapNode)
             {
-                propertyNode.ParseField(schema, _openApiSchemaFixedFields, _openApiSchemaPatternFields);
+                bool isRecognized = _openApiSchemaFixedFields.ContainsKey(propertyNode.Name) ||
+                        _openApiSchemaPatternFields.Any(p => p.Key(propertyNode.Name));
```
baywet, line 260:
```suggestion
                        _openApiSchemaPatternFields.Any(static p => p.Key(propertyNode.Name));
```

baywet (changes requested):

baywet (approved):

PR #2000, opened 2024-12-19
fix: sets hidi version to a preview

baywet:
context hidi version 1.4.16 and 1.14.15 are actually based off v2 and anything odata related to odata won't work anymore until we have a preview of yoko, this aligns with the preview version to avoid confusion


baywet:
related https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/1295

MaggieKimani1 (approved):

PR #2024, opened 2024-12-23
feat: adds deconstructor to read result

baywet:


captainsafia (commented):
I like this proposal for making it a bit easier to unwrap the document and diagnostics out of the read result!

src/Microsoft.OpenApi/Reader/ReadResult.cs
```diff
@@ -18,5 +18,13 @@ public class ReadResult
         /// OpenApiDiagnostic contains the Errors reported while parsing
         /// </summary>
         public OpenApiDiagnostic Diagnostic { get; set; }
+        /// <summary>
+        /// Deconstructs the result for easier assignment on the client application.
+        /// </summary>
+        public void Deconstruct(out OpenApiDocument document, out OpenApiDiagnostic diagnostic)
```
darrelmiller, line 24: So much magic in the C# language these days!!!

darrelmiller (approved):

Issue #1998, opened 2024-12-19
v2 - security - correct the proxy design implementation


baywet:
The current proxy for OpenAPISchema was implemented using inheritance, probably to save time, which forces all properties to be virtual.

Not only this is not the [correct proxy design pattern implementation](https://refactoring.guru/design-patterns/proxy) which requires using an interface instead, but this poses a [potential security and reliability issue](https://github.com/microsoft/OpenAPI.NET/security/code-scanning/2211)

baywet:
This is also causing bugs as we copy the values from the target, but throw away the copy all the time here.

https://github.com/microsoft/OpenAPI.NET/blob/f5fb254d8edbde952344825fff6a90e2ec4425f1/src/Microsoft.OpenApi/Models/References/OpenApiSchemaReference.cs#L29

```csharp
schema = new OpenApiSchemaReference("foo", null);
// Assume foo is not nullable.
schema.Nullable = true;
// The reference to foo is still not nullable.
```

The implementation should leverage lazy loading, and KEEP the new OpenAPI schema once created.

baywet:
Another issue is that the compiler doesn't help people know which properties are writeable for a reference since it currently inherits from OpenAPISchema.

I believe that we should introduce a new interface IOpenAPISchema which ONLY has the getters for the properties.
Then all places that currently return an OpenAPISchema should be updated to return that interface. Forcing consumers to do a type assertion and handle the different cases (any property being or not writeable, or updating the source component).

The target should probably also be a public property, in case the consumers want to walk up the reference.

Lastly, the reference resolving mechanism is cumbersome for write scenarios: you need to manually register the components before they can be resolved.

baywet:
Additional details from additional internal discussions:
- we don't want to use an interface if we can avoid it, as adding properties would be a breaking change. Maybe introducing an abstract base class is a better middle ground.
- We'll need to work on additional registration methods to make creation of documents from other object models (yoko, asp.net, etc...) easier. I'll start addressing that in an upcoming PR.

darrelmiller:
The OpenApiSchema should not clone the object.

I'm not convinced about the derived base class thing.  Can you add just setters in a derived class? 
I'm fine with making target public.

We need to make all the references implement IOpenApiReferenceable because the walker relies on it.

PR #2031, opened 2024-12-27
fix/reference proxy single copy

baywet:
partial #1998

### DIFF
--- a/src/Microsoft.OpenApi/Models/References/OpenApiSchemaReference.cs
+++ b/src/Microsoft.OpenApi/Models/References/OpenApiSchemaReference.cs
@@ -14,28 +14,68 @@ namespace Microsoft.OpenApi.Models.References
     /// </summary>
     public class OpenApiSchemaReference : OpenApiSchema
     {
-        internal OpenApiSchema _target;
-        private readonly OpenApiReference _reference;
-        private string _description;
-        private JsonNode _default;
-        private JsonNode _example;
-        private IList<JsonNode> _examples;
-
         #nullable enable
-        private OpenApiSchema? _targetProxy;
-        #nullable restore
+        private OpenApiSchema? _target;
+        private readonly OpenApiReference _reference;
+        private string? _description;
+        private JsonNode? _default;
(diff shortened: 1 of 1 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -587,6 +587,21 @@ public static ReadResult Parse(string input,
         {
             return OpenApiModelFactory.Parse(input, format, settings);
         }
+        /// <summary>
+        /// Adds a schema to the components object of the current document.
+        /// </summary>
+        /// <param name="openApiSchema">The schema to add</param>
+        /// <param name="id">The id for the component</param>
+        /// <returns>Whether the schema was added to the components.</returns>
+        public bool AddComponentSchema(string id, OpenApiSchema openApiSchema)
```
darrelmiller, line 596: Is this not relevant for all types of components?  Should we just have a AddComponent method with overloads for the different types of components?

baywet: I'm iterating here, so far the only one that was problematic in the context of yoko was the schemas.
We also need to think about making collections read only maybe, and adding a method to remove entries, so everything is in sync.

MaggieKimani1: We already have an extension method for registering components to the workspace. Should we just reuse these? https://github.com/microsoft/OpenAPI.NET/blob/cce2495ee73bcab3acc3348c22aa400b08dc6830/src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs#L62-L139

baywet: @darrelmiller I've just updated the methods to take into account all the component types.
@MaggieKimani1 no, this method registers everything at once. This is problematic because ordering might matter in some contexts (typically OData), and it results in a perf hit when wanting to register only a few components. Also, it's only accessible on the workspace, my goal was to make it accessible on the document for better discovery.

PR #2032, opened 2024-12-31
fix: side effects in tag references

baywet:
This fixes multiple issues:
- tag references are not like other component references. The tag in the operation is just a string. We had a lot of special casing, etc...
- lot of side effects where we were impacting the source tag by assigning things to the reference. As discussed in #1998
- fixes references not being resolvable in a lot of contexts due to a missing document reference.

We should still break the inheritance, and use a base abstract class as well.

### DIFF
--- a/src/Microsoft.OpenApi/Models/References/OpenApiTagReference.cs
+++ b/src/Microsoft.OpenApi/Models/References/OpenApiTagReference.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. 
 
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Writers;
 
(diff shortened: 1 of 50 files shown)


darrelmiller:
I'm not sure I agree with this.  Tag references are very similar to using $id to reference a Schema, where the target object is implicitly created.  I agree that it was a bit of a stretch when $refs could only use locators, but now that we need to support $refs as identifiers, I think there is a strong parallel.

baywet:
The previous model results in tons of side effects. 
I get the fact that we want to facilitate the readers life, but anybody making changes and not paying attention to the underlying specification is in for troubles.

A simple example:
- parse a description with tags
- on an operation, grab a tag
- add an extension to it
- you've now updated that tag for the whole document. 

Same goes for the description, the tag name etc...

darrelmiller:
> A simple example:
> 
> parse a description with tags
> on an operation, grab a tag
> add an extension to it
> you've now updated that tag for the whole document.
> Same goes for the description, the tag name etc..

Isn't that the expected behaviour?

src/Microsoft.OpenApi/Models/References/OpenApiTagReference.cs
```diff
@@ -10,21 +12,24 @@ namespace Microsoft.OpenApi.Models.References
     /// <summary>
     /// Tag Object Reference
     /// </summary>
-    public class OpenApiTagReference : OpenApiTag
+    public class OpenApiTagReference : OpenApiTag, IOpenApiReferenceable
```
darrelmiller, line 15: Why does this need to implement IOpenApiReferenceable? None of the other proxy objects do.

darrelmiller: Apparently it is a bug in every other reference class because the walker expects all references to implement IOpenApiReferencable  /cc @MaggieKimani1

darrelmiller:
Sorry, I was being slow.  I agree that editing the reference should not edit the target.  I agree with your changes.

darrelmiller (approved):

baywet:
> Sorry, I was being slow.  I agree that editing the reference should not edit the target.  I agree with your changes.

That just means the vacations were good!

src/Microsoft.OpenApi/Models/References/OpenApiTagReference.cs
```diff
@@ -10,21 +12,24 @@ namespace Microsoft.OpenApi.Models.References
...
+        public OpenApiTag Target
         {
             get
             {
-                _target ??= Reference.HostDocument?.ResolveReferenceTo<OpenApiTag>(_reference);
-                _target ??= new OpenApiTag() { Name = _reference.Id };
-                OpenApiTag resolved = new OpenApiTag(_target);
-                if (!string.IsNullOrEmpty(_description)) resolved.Description = _description;
-                return resolved;
+                _target ??= Reference.HostDocument?.Tags.FirstOrDefault(t => StringComparer.Ordinal.Equals(t.Name, Reference.Id));
```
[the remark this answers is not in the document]
baywet: Thanks for the feedback!

References still need a LOT of love at this point #1998 

Relaxing the instantiation of references without NRT support #1202 makes it extremely difficult not to mess something up. Effectively I _want_ the compiler to yell at us (devs on OAI.NET) when some part of the code forgot to check for nullability etc...
It'll also provide a much better experience to consumers like you by providing feedback when something was forgotten, etc...
Eventually, when that gets implemented, and the reference proxy design pattern is corrected, I think it'll be much easier to relax the constraints I hardened with this (and other) PRs in preview5. And eventually provide a walker that sets the document all over the place when absent for references before serialization.

What do you think of this "plan"?

PR #2033, opened 2024-12-31
fix: passes missing host document references to all layers

baywet:
reviewers, please review #2032 first.


darrelmiller:
I'm not sure we have to pass the host document through every layer as I think we can store it in the ParsingContext object which is accessible via the node object.  I also think we may need to store the "Entry Document" also, when dealing with multi-document descriptions.
 
Security Schemes and Tags are only supposed to be resolved from the "Entry Document" and not the Host document.

darrelmiller (dismissed):

MaggieKimani1 (approved):

PR #2046, opened 2025-01-08
fix/reference copy

baywet:
partial #1998


baywet:
@MaggieKimani1 I just had a thought: we could define a generic IOpenApiReferenceable<T> interface, that derives from the existing interface, and defines the target property. Thoughts?

MaggieKimani1:
> @MaggieKimani1 I just had a thought: we could define a generic IOpenApiReferenceable interface, that derives from the existing interface, and defines the target property. Thoughts?

Yeah this seems like a cleaner approach for designing the proxy classes. Let me implement this.

baywet:
@maggiekimani1 since I created this PR, I can't approve it, but I'm happy with the current content.
I think you can approve and merge it.

Also, when that is done, can you put together a release PR please? (bump preview versions)

MaggieKimani1 (approved):

MaggieKimani1 (approved):

RachitMalik12:
@baywet what's the one liner here?

PR #1245, opened 2023-05-11
ISSUE-1131: Added diagnostic reporting recursive concatenation of dia…

elize-vdr:
Added diagnostic reporting recursive concatenation of diagnostic reports from referenced files to the main file's diagnostic report.
Also added a unit test in `C:\Users\Elize.van.der.Riet\Work\OpenAPI\OpenAPI.NET - Fork\OpenAPI.NET\test\Microsoft.OpenApi.Readers.Tests\OpenApiReaderTests\OpenApiDiagnosticTests.cs`, my apologies not sure if this is the right place for the test, please let me know if I should move it.

PR #2053, opened 2025-01-10
make IDiagnosticExtensions internal and fix namespace

mus65:
I assume this was neither meant to be public nor meant to be in the global namespace. I noticed this when trying the latest preview because it conflicted with our own AddRange extension method.

This was introduced in #1245 and made public in https://github.com/microsoft/OpenAPI.NET/commit/90dc7fc5587cee40f5964919387baa3bafd8b9d8#diff-65294c3e990da88445c0036dfe9f8e7bcfb00675881d47c22fb3c6000c3b9d76R56 .


MaggieKimani1:
@mus65 thanks for noticing this and for contributing the fix!

MaggieKimani1 (approved):

PR #2063, opened 2025-01-15
feat: configure AOT for trimming compatibility

### LINKED ISSUE #1875: V2 - should be compatible with AOT/trimming
While we don't need that on GA, we should at least ensure that there are no design decisions which are going to corner us.

We should at the mimum:
- add a test project
- add a target of net5/6/8

There are a number of places in this repo where reflection is present, and we should avoid it for performance/reliability reasons.

See kiota-dotnet for the test project setup.

MaggieKimani1:
Fixes #1875


test/Microsoft.OpenApi.Trimming.Tests/Microsoft.OpenApi.Trimming.Tests.csproj
```diff
@@ -4,17 +4,19 @@
     <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <PublishTrimmed>true</PublishTrimmed>
     <PublishAot>true</PublishAot>
     <TrimmerSingleWarn>false</TrimmerSingleWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <NoWarn>NU1903</NoWarn>
+    <NoWarn>NU1903; IL3000</NoWarn>
```
baywet, line 11: I'd support an in place suppression instead to avoid future changes introducing similar issues without us noticing.

src/Microsoft.OpenApi/Microsoft.OpenApi.csproj
```diff
@@ -6,6 +6,7 @@
         <Version>2.0.0-preview4</Version>
         <Description>.NET models with JSON and YAML writers for OpenAPI specification</Description>
         <SignAssembly>true</SignAssembly>
+        <IsTrimmable Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)','net5.0'))">true</IsTrimmable>
```
andrueastman, line 9: Sorry, I forgot to mention that you should probably add this to the csproj too since you have net8.0 targets here.

test/Microsoft.OpenApi.Trimming.Tests/Microsoft.OpenApi.Trimming.Tests.csproj
MaggieKimani1: There's no code line attached to the warning as it's caused by external code:
https://github.com/microsoft/OpenAPI.NET/actions/runs/12812328824/job/35723707783?pr=2063

We've just replicated the dotnet sdk's implementation here:
https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/c225b2e7d291de95ca28b9d5529b276d63019dd4/tests/Microsoft.Graph.DotnetCore.Core.Trimming/Microsoft.Graph.DotnetCore.Core.Trimming.csproj#L12

baywet (approved):

PR #2079, opened 2025-01-22
fix: a flaky behaviour for format property serialization

baywet:


MaggieKimani1:
Let's include this fix in v1.x as well.

Issue #2062, opened 2025-01-15
v2: Handle relative references in OpenApiSchemaReference type


captainsafia:
The OpenApiSchemaReference type included in this package provides support for resolving schema references to the top-level `components.schemas` object. The object doesn't play nice if the reference is a relative reference to a property in a recursive schema.

The serializer seems to detect these relative references correctly and serialize the correct references. It would be great if these resolutions also worked in the object model, particularly for users taking advantage of consuming APIs' in ASP.NET Core like OpenAPI schema transformers.

For example, given the following type hierarchy:

```csharp
public class A
{
	public B B { get; }
}

public class B
{
	public C C { get; }
}

public class C
{
	public B B { get; set; } 
}
```

And the following schema:

```
{
  "type": [
    "object",
    "null"
  ],
  "properties": {
    "b": {
      "type": [
        "object",
        "null"
      ],
      "properties": {
        "c": {
          "type": [
            "object",
            "null"
          ],
          "properties": {
            "b": {
              "$ref": "#/properties/b"
            }
          }
        }
      }
    }
  }
}
```

`#/properties/b` should map to the properties.b schema when interacting with the `OpenApiSchemaReference` object for the schema.

baywet:
While working on the proxy design pattern fix, I came to the same conclusion!
https://github.com/microsoft/OpenAPI.NET/pull/2111#discussion_r1935741530

The good news is that with the work I've done, it should now be a single place fix + providing a recursion method to resolve the actual schema when necessary.

RachitMalik12:
Spec is being worked on. Waiting on this before beginning implementation.

RachitMalik12:
@darrelmiller will follow up with @baywet to see if we can backport these fixes to 1.6.x

RachitMalik12:
https://github.com/MicrosoftDocs/openapi-docs/pull/145/files

baywet:
> [@darrelmiller](https://github.com/darrelmiller) will follow up with [@baywet](https://github.com/baywet) to see if we can backport these fixes to 1.6.x

To answer that part: not without major rework and breaking changes. Which makes the cost prohibitive. So customers who want a great support for all reference scenarios will have to upgrade to 2.X.

As an update, we have the spec mostly built (I need to review it to see if some gaps are still present), and we should be able to proceed with implementation soonish.

PR #2111, opened 2025-01-30
fix/references cleanup

### LINKED ISSUE #1998: v2 - security - correct the proxy design implementation

baywet:
fixes #1998
closes #2088
review #2109 first
This PR switches from copy constructors to dedicated methods to enable polymorphism and avoid accidental duplicatations of the object model

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
@@ -38,7 +38,7 @@ namespace Microsoft.OpenApi.Models
         /// <summary>
         /// Initializes a copy instance of an <see cref="IOpenApiRequestBody"/> object
         /// </summary>
-        public OpenApiRequestBody(IOpenApiRequestBody requestBody)
+        internal OpenApiRequestBody(IOpenApiRequestBody requestBody)
         {
             Utils.CheckArgumentNull(requestBody);
             Description = requestBody?.Description ?? Description;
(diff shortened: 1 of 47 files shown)


src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
```diff
@@ -123,23 +124,37 @@ public IEnumerable<IOpenApiParameter> ConvertToFormDataParameters(IOpenApiWriter
...
                 if ((paramSchema.Type & JsonSchemaType.String) == JsonSchemaType.String
                     && ("binary".Equals(paramSchema.Format, StringComparison.OrdinalIgnoreCase)
                     || "base64".Equals(paramSchema.Format, StringComparison.OrdinalIgnoreCase)))
                 {
-                    paramSchema.Type = "file".ToJsonSchemaType();
-                    paramSchema.Format = null;
+                    var updatedSchema = paramSchema switch {
+                        OpenApiSchema s => s, // we already have a copy
+                        // we have a copy of a reference but don't want to mutate the source schema
+                        // TODO might need recursive resolution of references here
```
baywet, line 135: on this one I have a question pending internally with @darrelmiller

baywet: confirmed with Darrel, references of references are possible.
I'll stand up another PR to update the return type of the target field, and enable some recursion work.

baywet: this is actually what's being discussed in #2062 !!

andrueastman (approved):

PR #2112, opened 2025-01-31
fix: a bug where 3.0 downcast of type null would not work

baywet:
null type in the object model now results in nullable true for 3.0 serialization

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -633,6 +633,7 @@ namespace Microsoft.OpenApi.Models
         {
             // check whether nullable is true for upcasting purposes
             var isNullable = Nullable || 
+                                Extensions is not null &&
                                 Extensions.TryGetValue(OpenApiConstants.NullableExtension, out var nullExtRawValue) && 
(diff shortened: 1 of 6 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -635,20 +631,39 @@ private void SerializeAsV2(
 
         private void SerializeTypeProperty(JsonSchemaType? type, IOpenApiWriter writer, OpenApiSpecVersion version)
         {
+            // check whether nullable is true for upcasting purposes
+            var isNullable = Nullable ||
+                            Type is JsonSchemaType.Null ||
```
andrueastman, line 636:
```suggestion
                            (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null))  ||
```
If I'm not wrong, the use of `is` won't return true if the enum has more than one flag set. e.g when Type is set as `JsonSchemaType.Object | JsonSchemaType.Null`

Is this the intention here?

baywet: thanks! great catch here!

andrueastman (approved):

PR #2117, opened 2025-02-03
fix/no more nullable property

baywet:
review #2112 first


andrueastman (approved):
Really appreciate the cleanup here 👍🏼

lyczkop:
why it was removed?
What I should use instead of this Nullable property?

lyczkop:
> why it was removed? What I should use instead of this Nullable property?

ok I think I found it... https://github.com/microsoft/OpenAPI.NET/blob/ed83b2815077717ecf214d1f757dff2460576b54/docs/upgrade-guide-2.md#openapischemas-type-property-is-now-a-flaggable-enum

PR #2118, opened 2025-02-04
Remove dependency on CurrentCulture in tests

dldl-cmd:
On a system configured for german language the tests weren't working because the . on the german system are written as , which results in invalid json, because the StreamWriter depends on the CurrentCulture by default.


baywet (approved):
Thanks for the contribution!

PR #2119, opened 2025-02-04
Fix reading relative path file reference

dldl-cmd:
A reference like` ./Directory/File.json#/components/schema/abc` was before read as: 
`refId` = `abc`
`externalResource` = `./Directory`

Which dropped the file name, which will be fixed by this commit


baywet (changes requested):
Thanks for the contribution!

Can you please add unit tests so we don't get any future regression?

baywet:
sorry for the closure, we changed the branching to align with other repositories. Can you also re-open the pull request to main please?

dldl-cmd:
> sorry for the closure, we changed the branching to align with other repositories. Can you also re-open the pull request to main please?

@baywet Of course no problem, here is the new PR: #2133

PR #2115, opened 2025-02-03
preview 6 release

baywet:
holding for #2112 #2113 #2114

PR #2121, opened 2025-02-04
fix/immutable reference

baywet:
makes document for references optional. Cleans up internal constructors for references, makes references immutable to avoid side effects

related #1998

### DIFF
--- a/src/Microsoft.OpenApi/Interfaces/IOpenApiReferenceHolder.cs
+++ b/src/Microsoft.OpenApi/Interfaces/IOpenApiReferenceHolder.cs
@@ -34,6 +34,6 @@ namespace Microsoft.OpenApi.Interfaces
         /// <summary>
         /// Reference object.
         /// </summary>
-        OpenApiReference Reference { get; set; }
+        OpenApiReference Reference { get; init; }
     }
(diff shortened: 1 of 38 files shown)


baywet:
@captainSafia this is where I wanted to get: 
- The proxy design pattern is properly implemented, one interface, two implementations. The consumers can use type assertions while reading a document to know what's mutable, what's not.
- All reference types now offer a single public constructor, which only requires the reference ID but also accepts document, external resource. This way creating a document from the DOM remains simple/doesn't require carrying document references all over the place.
- The target property is a pure function (no more backing field with potential side effects)
- The reference property is immutable (no more side effects)
- The properties of the reference type are also immutable (no more side effects)

The only drawback I can see is having to call RegisterComponents and/or SetReferenceHostDocument on the OpenAPIDocument based on what the consumer is doing. But we could "hide that" by calling those methods before serialization.

We'll try to release soon, in fact #2115 is already up so you can test that ASAP. I've noticed a much better developer experience and reliability on my end in kiota, openapi.net.odata, plugins, and API manifest libs. (linking locally)

andrueastman (approved):

PR #2124, opened 2025-02-05
fix: support non-standard MIME type in response header

### LINKED ISSUE #1989: Getting NullReferenceException when deserializing a yaml document
**Describe the bug**
I aim trying to deserialize a yaml open api file, but it's getting NullReferenceException

**OpenApi File To Reproduce**
https://api.codacy.com/api/api-docs/swagger.yaml

**Expected behavior**
The document shoud be serialized correctly

**Screenshots/Code Snippets**

Code to reproduce available here:
https://github.com/lucassklp/Rx.Http.CodeGen/tree/ce2a7b97ec91ba09e56f3c77d62e1730abe1463d

Methods i'm calling:
Image

Stacktrace:
Image

**Additional context**
Version: 2.0.0-preview2

MaggieKimani1:
Fixes #1989 where the document's format was not being correctly set for non-standard MIME types when inspecting the content type header in a HTTP response object.
E.g. the MIME type `text/x-yaml` which is still in use by older libraries or APIs

### DIFF
--- a/src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
+++ b/src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
@@ -279,10 +279,14 @@ namespace Microsoft.OpenApi.Reader
                     var mediaType = response.Content.Headers.ContentType.MediaType;
                     var contentType = mediaType.Split(";".ToCharArray(), StringSplitOptions.RemoveEmptyEntries)[0];
                     format = contentType.Split('/').LastOrDefault();
+                    if (!string.IsNullOrEmpty(format) && format.Contains('-'))
+                    {
+                        format = format.Split('-').LastOrDefault(); // for non-standard MIME types e.g. text/x-yaml used in older libs or apps
+                    }
 #if NETSTANDARD2_0
                     stream = await response.Content.ReadAsStreamAsync();
 #else
-                    stream = await response.Content.ReadAsStreamAsync(token).ConfigureAwait(false);;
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
```diff
@@ -279,10 +299,14 @@ private static ReadResult InternalLoad(MemoryStream input, string format, OpenAp
                     var mediaType = response.Content.Headers.ContentType.MediaType;
                     var contentType = mediaType.Split(";".ToCharArray(), StringSplitOptions.RemoveEmptyEntries)[0];
                     format = contentType.Split('/').LastOrDefault();
+                    if (!string.IsNullOrEmpty(format) && format.Contains('-'))
+                    {
+                        format = format.Split('-').LastOrDefault(); // for non-standard MIME types e.g. text/x-yaml used in older libs or apps
+                    }
```
baywet, line 305:
```suggestion
                    format = contentType.Split('/').Last().Split('-').Last().Split('+').Last();
                    // for non-standard MIME types e.g. text/x-yaml used in older libs or apps
```

Some context here:
- `string.Empty.Split('-').Last()` returns string empty
- using `-` [is not recommended by the RFC](https://www.rfc-editor.org/rfc/rfc6838#section-3.4)
- the `+` is to support vendor specific media types

baywet (changes requested):

baywet (approved):

PR #2146, opened 2025-02-13
feat: enable null reference type support

### LINKED ISSUE #1202: Enable null reference types on next major version

MaggieKimani1:
This PR:
- Adds null reference type support by enabling the `nullable` feature in the core and readers project
- Annotates all properties and method parameters/return types with ? for nullable types
- Fixes compiler warnings such as possible dereferences of null values.

Fixes #1202


src/Microsoft.OpenApi/Interfaces/IOpenApiVersionService.cs
```diff
@@ -19,7 +19,7 @@ internal interface IOpenApiVersionService
         /// <param name="summary">The summary of the reference.</param>
         /// <param name="description">A reference description</param>
         /// <returns>The <see cref="OpenApiReference"/> object or null.</returns>
-        OpenApiReference ConvertToOpenApiReference(string reference, ReferenceType? type, string summary = null, string description = null);
+        OpenApiReference? ConvertToOpenApiReference(string reference, ReferenceType? type, string? summary = null, string? description = null);
```
baywet, line 22: why do we accept a nullable reference type here? shouldn't it be required?

src/Microsoft.OpenApi/Interfaces/IOpenApiReferenceHolder.cs
```diff
@@ -15,11 +15,11 @@ public interface IOpenApiReferenceHolder<out T, V> : IOpenApiReferenceHolder whe
         /// <summary>
         /// Gets the resolved target object.
         /// </summary>
-        T Target { get; }
+        T? Target { get; }
         /// <summary>
         /// Copy the reference as a target element with overrides.
         /// </summary>
-        V CopyReferenceAsTargetElementWithOverrides(V source);
+        V? CopyReferenceAsTargetElementWithOverrides(V? source);
```
baywet, line 22: why do we accept a nullable source? if you're doing a copy, you should make sure the source object is not null.

src/Microsoft.OpenApi/Interfaces/IOpenApiReferenceHolder.cs
```diff
@@ -34,6 +34,6 @@ public interface IOpenApiReferenceHolder : IOpenApiSerializable
         /// <summary>
         /// Reference object.
         /// </summary>
-        OpenApiReference Reference { get; init; }
+        OpenApiReference? Reference { get; init; }
```
baywet, line 37: in what context would the reference object ever be null?

src/Microsoft.OpenApi/Models/OpenApiExtensibleDictionary.cs
```diff
@@ -27,16 +27,16 @@ protected OpenApiExtensibleDictionary():this(null) { }
         /// <param name="dictionary">The generic dictionary.</param>
         /// <param name="extensions">The dictionary of <see cref="IOpenApiExtension"/>.</param>
         protected OpenApiExtensibleDictionary(
```
baywet, line 29: wouldn't it be better to make the dictionary parameter non nullable and optional with a default value of `[]` (if that's possible) ?

src/Microsoft.OpenApi/Models/OpenApiOperation.cs
```diff
@@ -248,17 +248,25 @@ public void SerializeAsV2(IOpenApiWriter writer)
                     if (consumes.Contains("application/x-www-form-urlencoded") ||
                         consumes.Contains("multipart/form-data"))
                     {
-                        parameters.AddRange(RequestBody.ConvertToFormDataParameters(writer));
+                        var formDataParameters = RequestBody.ConvertToFormDataParameters(writer);
+                        if (formDataParameters != null)
```
baywet, line 252:
```suggestion
                        if (RequestBody.ConvertToFormDataParameters(writer) is {} formDataParameters)
```
nit: you can do something like that, which not only reduces allocations but also allows you to combine the condition with the parent one

src/Microsoft.OpenApi/Models/OpenApiReference.cs
```diff
@@ -18,23 +18,23 @@ public class OpenApiReference : IOpenApiSerializable, IOpenApiDescribedElement,
...
 
         /// <summary>
         /// External resource in the reference.
         /// It maybe:
         /// 1. a absolute/relative file path, for example:  ../commons/pet.json
         /// 2. a Url, for example: http://localhost/pet.json
         /// </summary>
-        public string ExternalResource { get; init; }
-
+        public string? ExternalResource { get; init; }
```
baywet, line 36: leaving the comment here because I can't below.
I'm not sure Type should be nullable

src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
```diff
@@ -102,15 +102,16 @@ public IOpenApiParameter ConvertToBodyParameter(IOpenApiWriter writer)
...
-                Examples = Content.Values.FirstOrDefault()?.Examples,
+                Schema = Content?.Values.FirstOrDefault()?.Schema ?? new OpenApiSchema(),
+                Examples = Content?.Values.FirstOrDefault()?.Examples,
                 Required = Required,
-                Extensions = Extensions.ToDictionary(static k => k.Key, static v => v.Value)  // Clone extensions so we can remove the x-bodyName extensions from the output V2 model.
+                Extensions = Extensions?.ToDictionary(static k => k.Key, static v => v.Value)
             };
-            if (bodyParameter.Extensions.ContainsKey(OpenApiConstants.BodyName))
+            // Clone extensions so we can remove the x-bodyName extensions from the output V2 model.
+            if (bodyParameter.Extensions != null && bodyParameter.Extensions.TryGetValue(OpenApiConstants.BodyName, out var bodyNameExtension))
```
baywet, line 111: test the type in the condition with is (type assertion) instead.
reduces allocations, and now you know bodyname can't be null this way

src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
```diff
@@ -121,34 +122,41 @@ public IEnumerable<IOpenApiParameter> ConvertToFormDataParameters(IOpenApiWriter
...
+                    if ((paramSchema.Type & JsonSchemaType.String) == JsonSchemaType.String
+                        && ("binary".Equals(paramSchema.Format, StringComparison.OrdinalIgnoreCase)
+                        || "base64".Equals(paramSchema.Format, StringComparison.OrdinalIgnoreCase)))
+                    {
+                        var updatedSchema = paramSchema switch
+                        {
+                            OpenApiSchema s => s, // we already have a copy
+                                                  // we have a copy of a reference but don't want to mutate the source schema
+                                                  // TODO might need recursive resolution of references here
+                            OpenApiSchemaReference r => (OpenApiSchema?)r.Target?.CreateShallowCopy(),
```
baywet, line 140:
```suggestion
                            OpenApiSchemaReference r when r.Target is not null => (OpenApiSchema)r.Target.CreateShallowCopy(),
                            OpenApiSchemaReference r2 when r.Target is null => throw new InvalidOperationException("Unresolved reference target"),
```

and drop the if you've added below

src/Microsoft.OpenApi/Models/OpenApiResponse.cs
```diff
@@ -101,7 +101,11 @@ public void SerializeAsV2(IOpenApiWriter writer)
             // description
             writer.WriteRequiredProperty(OpenApiConstants.Description, Description);
 
-            var extensionsClone = new Dictionary<string, IOpenApiExtension>(Extensions);
+            Dictionary<string, IOpenApiExtension>? extensionsClone = null;
```
baywet, line 104: nit: missed ternary operation opportunity

src/Microsoft.OpenApi/Models/OpenApiResponse.cs
```diff
@@ -136,7 +140,7 @@ public void SerializeAsV2(IOpenApiWriter writer)
 
                         foreach (var example in Content
                             .Where(mediaTypePair => mediaTypePair.Value.Examples != null && mediaTypePair.Value.Examples.Any())
-                            .SelectMany(mediaTypePair => mediaTypePair.Value.Examples))
+                            .SelectMany(mediaTypePair => mediaTypePair.Value.Examples!))
```
baywet, line 143: don't use bang operators , they usually lead to bugs.
Add an OfType<Foo>() after the select many instead, not only it'll filter the null values at runtime, it'll fix the compiler complaining about nullable values.

baywet (changes requested):
Great work here!
I've only done a partial review at this point starting with the interfaces, since changes there will require reviewing implementations again.

src/Microsoft.OpenApi/Interfaces/IOpenApiReferenceHolder.cs
MaggieKimani1: There's a possible null reference assignment here https://github.com/microsoft/OpenAPI.NET/blob/5db8757df642dbe651552ce4a7c740e94474eafc/src/Microsoft.OpenApi/Models/References/BaseOpenApiReferenceHolder.cs#L28

src/Microsoft.OpenApi/Models/OpenApiExtensibleDictionary.cs
MaggieKimani1: Yeah that's possible. However I didn't want to tamper with/change model data as it might have side effects during serialization

src/Microsoft.OpenApi/Interfaces/IOpenApiReferenceHolder.cs
baywet: I think it'd be an improvement since reference objects that do not have reference information are non-sensical to start with.

src/Microsoft.OpenApi/Models/OpenApiResponse.cs
MaggieKimani1: Using OfType() still doesn't work. There's still a possible null reference return at the SelectMany():
```c#
foreach (var example in Content
    .Where(mediaTypePair => mediaTypePair.Value.Examples is not null && mediaTypePair.Value.Examples.Any())
    .SelectMany(mediaTypePair => mediaTypePair.Value.Examples)
    .OfType<KeyValuePair<string, IOpenApiExample>>())
{
    writer.WritePropertyName(example.Key);
    example.Value.SerializeAsV2(writer);
}
```

src/Microsoft.OpenApi/Interfaces/IOpenApiVersionService.cs
MaggieKimani1: It can be null in the case of an externally referenced file:
https://github.com/microsoft/OpenAPI.NET/blob/85c3a4e694e3410dd21e3a7b75b21dd1f7d24ad8/src/Microsoft.OpenApi/Reader/V2/OpenApiV2VersionService.cs#L142-L151

src/Microsoft.OpenApi/Any/OpenApiAny.cs
```diff
@@ -12,21 +12,21 @@ namespace Microsoft.OpenApi.Any
...
     {
-        private readonly JsonNode jsonNode;
+        private readonly JsonNode? jsonNode;
 
         /// <summary>
         /// Initializes the <see cref="OpenApiAny"/> class.
         /// </summary>
         /// <param name="jsonNode"></param>
-        public OpenApiAny(JsonNode jsonNode)
+        public OpenApiAny(JsonNode? jsonNode)
```
baywet, line 21: why are we accepting a null node at all here? this is going to propagate all over the place.

src/Microsoft.OpenApi/Attributes/DisplayAttribute.cs
```diff
@@ -15,14 +15,14 @@ public class DisplayAttribute : Attribute
...
+        public DisplayAttribute(string? name)
         {
             Name = Utils.CheckArgumentNullOrEmpty(name);
         }
 
         /// <summary>
         /// The display Name.
         /// </summary>
-        public string Name { get; }
+        public string? Name { get; }
```
baywet, line 26: revert this, it's assigned in the only constructor and checked it's not null. I suspect the Utils method return type needs to be updated instead.

src/Microsoft.OpenApi/Exceptions/OpenApiReaderException.cs
```diff
@@ -39,7 +39,7 @@ public OpenApiReaderException(string message, ParsingContext context) : base(mes
         /// </summary>
         /// <param name="message">Plain text error message for this exception.</param>
         /// <param name="node">Parsing node where error occured</param>
-        public OpenApiReaderException(string message, JsonNode node) : base(message)
+        public OpenApiReaderException(string message, JsonNode? node) : base(message)
```
baywet, line 42: we're not doing anything with the node parameter, remove the constructor entirely

src/Microsoft.OpenApi/Exceptions/OpenApiUnsupportedSpecVersionException.cs
```diff
@@ -18,7 +18,7 @@ public class OpenApiUnsupportedSpecVersionException : Exception
         /// Initializes the <see cref="OpenApiUnsupportedSpecVersionException"/> class with a specification version.
         /// </summary>
         /// <param name="specificationVersion">Version that caused this exception to be thrown.</param>
-        public OpenApiUnsupportedSpecVersionException(string specificationVersion)
+        public OpenApiUnsupportedSpecVersionException(string? specificationVersion)
```
baywet, line 21: if we have an exception dedicated to telling users the specification version they are trying to read is not supported, having the version as the error message should be mandatory.

src/Microsoft.OpenApi/Models/OpenApiOperation.cs
```diff
@@ -248,17 +248,24 @@ public void SerializeAsV2(IOpenApiWriter writer)
                     if (consumes.Contains("application/x-www-form-urlencoded") ||
                         consumes.Contains("multipart/form-data"))
                     {
-                        parameters.AddRange(RequestBody.ConvertToFormDataParameters(writer));
+                        if (RequestBody.ConvertToFormDataParameters(writer) is {} formDataParameters)
```
baywet, line 251: collapse the if with the parent one

src/Microsoft.OpenApi/Models/OpenApiOperation.cs
```diff
@@ -248,17 +248,24 @@ public void SerializeAsV2(IOpenApiWriter writer)
...
+                        }
                     }
                     else
                     {
-                        parameters.Add(RequestBody.ConvertToBodyParameter(writer));
+                        var bodyParameter = RequestBody.ConvertToBodyParameter(writer);
+                        if (bodyParameter != null)
+                        {
+                            parameters.Add(bodyParameter);
+                        }
```
baywet, line 262: use type assertion and collapse with parent else

src/Microsoft.OpenApi/Models/References/BaseOpenApiReferenceHolder.cs
```diff
@@ -63,20 +63,20 @@ protected BaseOpenApiReferenceHolder(string referenceId, OpenApiDocument hostDoc
     /// <inheritdoc/>
     public virtual void SerializeAsV3(IOpenApiWriter writer)
     {
-        if (!writer.GetSettings().ShouldInlineReference(Reference))
+        if (Reference is not null && !writer.GetSettings().ShouldInlineReference(Reference))
```
baywet, line 66: I believe we should add the required keyword to the reference field, as it's already assigned in all constructors anyway and not mutable externally. So it can't be null.
This way we can clean up all those extra validations

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
```diff
@@ -61,7 +61,7 @@ public ReadResult Read(MemoryStream input,
         /// <param name="jsonNode">The JsonNode input.</param>
         /// <param name="settings">The Reader settings to be used during parsing.</param>
         /// <returns></returns>
-        public ReadResult Read(JsonNode jsonNode,
+        public ReadResult Read(JsonNode? jsonNode,
```
baywet, line 64: revert, since we're not allowing null values below

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
```diff
@@ -198,14 +201,17 @@ public T ReadFragment<T>(JsonNode input,
...
-                    diagnostic.Errors.Add(item);
+                    foreach (var item in errors)
+                    {
+                        diagnostic.Errors.Add(item);
+                    }
                 }
             }
 
-            return (T)element;
+            return (T?)element!;
```
baywet, line 214: avoid bang operators

baywet (commented):
We're making good progress here, here is another set of comments

src/Microsoft.OpenApi/Models/OpenApiResponse.cs
baywet: See b6d68bbb430b5e08d701f6a831fe07245dbecad8

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -107,7 +107,7 @@ public static Task SerializeAsync<T>(
         /// <param name="writer">The output writer.</param>
         /// <param name="specVersion">Version of the specification the output should conform to</param>
         /// <param name="cancellationToken">The cancellation token.</param>
-        public static Task SerializeAsync<T>(this T element, IOpenApiWriter writer, OpenApiSpecVersion specVersion, CancellationToken cancellationToken = default)
+        public static Task SerializeAsync<T>(this T? element, IOpenApiWriter writer, OpenApiSpecVersion specVersion, CancellationToken cancellationToken = default)
```
baywet, line 110: revert the change on element and the null propagation since we check that element is not null in the first line

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -82,7 +82,7 @@ public static Task SerializeAsync<T>(
             Stream stream,
             OpenApiSpecVersion specVersion,
             OpenApiFormat format,
-            OpenApiWriterSettings settings,
+            OpenApiWriterSettings? settings,
```
baywet, line 85: if settings is nullable, should we also make it optional?

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
```diff
@@ -30,19 +30,19 @@ internal static class StringExtensions
...
 
-            if (displayMap.TryGetValue(displayName, out var cachedValue))
+            if (displayName is not null && displayMap.TryGetValue(displayName, out var cachedValue))
             {
                 result = (T)cachedValue;
                 return true;
             }
 
-            result = default;
+            result = default!;
```
baywet, line 45: avoid using bang operators, make result nullable instead and add a "NotNullWhen(true)" attribute

src/Microsoft.OpenApi/Extensions/OpenApiReferencableExtensions.cs
```diff
@@ -54,12 +54,12 @@ public static IOpenApiReferenceable ResolveReference(this IOpenApiReferenceable
...
             OpenApiHeader headerElement,
-            string propertyName,
-            string mapKey,
+            string? propertyName,
+            string? mapKey,
             JsonPointer pointer)
         {
             if (OpenApiConstants.Examples.Equals(propertyName, StringComparison.Ordinal) &&
-                !string.IsNullOrEmpty(mapKey) &&
+                mapKey is not null &&
```
baywet, line 62: you are changing the behaviour but not checking for empty string anymore, this should be reverted (and other places in this document)

src/Microsoft.OpenApi/Models/OpenApiRequestBody.cs
```diff
@@ -121,34 +123,41 @@ public IEnumerable<IOpenApiParameter> ConvertToFormDataParameters(IOpenApiWriter
...
+                        && ("binary".Equals(paramSchema.Format, StringComparison.OrdinalIgnoreCase)
+                        || "base64".Equals(paramSchema.Format, StringComparison.OrdinalIgnoreCase)))
+                    {
+                        var updatedSchema = paramSchema switch
+                        {
+                            OpenApiSchema s => s, // we already have a copy
+                                                  // we have a copy of a reference but don't want to mutate the source schema
+                                                  // TODO might need recursive resolution of references here
+                            OpenApiSchemaReference r when r.Target is not null => (OpenApiSchema)r.Target.CreateShallowCopy(),
+                            OpenApiSchemaReference r2 when r2.Target is null => throw new InvalidOperationException("Unresolved reference target"),
```
baywet, line 142:
```suggestion
                            OpenApiSchemaReference => throw new InvalidOperationException("Unresolved reference target"),
```

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -715,7 +715,7 @@ private void SerializeAsV2(
             // readOnly
             // In V2 schema if a property is part of required properties of parent schema,
             // it cannot be marked as readonly.
-            if (!parentRequiredProperties.Contains(propertyName))
+            if (propertyName is not null && !parentRequiredProperties.Contains(propertyName))
```
baywet, line 718: string is null or empty instead?

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
```diff
@@ -90,15 +90,18 @@ public ReadResult Read(JsonNode jsonNode,
...
             {
-                var openApiErrors = document.Validate(settings.RuleSet);
-                foreach (var item in openApiErrors.OfType<OpenApiValidatorError>())
-                {
-                    diagnostic.Errors.Add(item);
-                }
-                foreach (var item in openApiErrors.OfType<OpenApiValidatorWarning>())
+                var openApiErrors = document?.Validate(settings.RuleSet);
+                if(openApiErrors is not null)
                 {
```
baywet, line 95: it'd probably be better to check whether the document is not null in the previous condition

src/Microsoft.OpenApi/Reader/OpenApiJsonReader.cs
```diff
@@ -198,14 +202,17 @@ public T ReadFragment<T>(JsonNode input,
             // Validate the element
             if (settings.RuleSet != null && settings.RuleSet.Rules.Any())
             {
-                var errors = element.Validate(settings.RuleSet);
-                foreach (var item in errors)
+                var errors = element?.Validate(settings.RuleSet);
```
baywet, line 205: validate element is not null in parent condition instead

test/Microsoft.OpenApi.Readers.Tests/ReferenceService/ConvertToOpenApiReferenceV3Tests.cs
```diff
@@ -1,145 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT license.
-
-using Microsoft.OpenApi.Models;
-using Microsoft.OpenApi.Reader;
-using Microsoft.OpenApi.Reader.V3;
-using Xunit;
-
-namespace Microsoft.OpenApi.Readers.Tests
```
baywet, line 9: why are we removing a bunch of unit tests?

baywet (changes requested):

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
MaggieKimani1: Unfortunately `string.IsNullOrEmpty` still treats a nullableString as string?, resulting in a nullability warning. The compiler still doesn't recognize that the nullable string must be non-null if the condition passes hence the explicit null check

src/Microsoft.OpenApi/Extensions/StringExtensions.cs
MaggieKimani1: The `NotNullWhen(true)` attribute is inaccessible since its an internal sealed class but I'll make the result nullable

src/Microsoft.OpenApi/Extensions/OpenApiReferencableExtensions.cs
MaggieKimani1: string.IsNullOrEmpty doesn't handle nullable strings properly, mapKey is still treated as string? once it exits the condition leading to compiler warnings in the subsequent code.
I can only think of using the null-forgiving operator in conjuction with `string.IsNullOrEmpty` otherwise we might have to explicitly check whether a string is null or empty as two separate checks.
Do you have a better approach?

baywet: > string.IsNullOrEmpty doesn't handle nullable strings properly, mapKey is still treated as string? once it exits the condition leading to compiler warnings in the subsequent code.

That's only because we're targeting netstandard2 and the stdlib does NOT have the nullability annotations. If we targeted net8 only (not suggesting we should change that), you wouldn't have compiler warnings anymore.

> Do you have a better approach?

not really, I think we could have both (is not null + string null or empty) to make the compiler happy and maintain the current behaviour.

What do you think?

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
baywet: same rationale as previous comment

test/Microsoft.OpenApi.Readers.Tests/ReferenceService/ConvertToOpenApiReferenceV3Tests.cs
baywet: thanks for the contextualization here. This PR is large. I simply wanted to ensure we were not removing unit tests for code that remains simply to make our lives easier: coverage should remain the same or improve :)

src/Microsoft.OpenApi/Extensions/OpenApiReferencableExtensions.cs
MaggieKimani1: Yeah I think this suffices for now.

src/Microsoft.OpenApi/Models/OpenApiResponse.cs
MaggieKimani1: Thank you!

src/Microsoft.OpenApi/Reader/V2/OpenApiContactDeserializer.cs
```diff
@@ -35,7 +42,7 @@ internal static partial class OpenApiV2Deserializer
             {s => s.StartsWith(OpenApiConstants.ExtensionFieldNamePrefix, StringComparison.OrdinalIgnoreCase), (o, p, n, _) => o.AddExtension(p, LoadExtension(p, n))}
         };
 
-        public static OpenApiContact LoadContact(ParseNode node, OpenApiDocument hostDocument)
+        public static OpenApiContact LoadContact(ParseNode node, OpenApiDocument? hostDocument)
```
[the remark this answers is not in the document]
MaggieKimani1: My bad. It should be passed as an argument in the ParseMap method

src/Microsoft.OpenApi/Validations/IValidationContext.cs
```diff
@@ -41,6 +41,6 @@ public interface IValidationContext
         /// <summary>
         /// 
         /// </summary>
-        OpenApiDocument HostDocument { get; }
+        OpenApiDocument? HostDocument { get; }
```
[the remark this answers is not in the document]
MaggieKimani1: Actually we don't need to pass a document instance to the validation context anymore. This was part of the initial JSON schema implementation where we were using a local JSON schema registry to perform lookups and resolve $refs but this became obsolete https://github.com/microsoft/OpenAPI.NET/commit/e76101a582b59d94a98f6103fe05c3d908ef79ef

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
```diff
@@ -176,7 +205,7 @@ public bool RegisterComponentForDocument<T>(OpenApiDocument openApiDocument, T c
                 IOpenApiExample => baseUri + ReferenceType.Example.GetDisplayName() + ComponentSegmentSeparator + id,
                 IOpenApiHeader => baseUri + ReferenceType.Header.GetDisplayName() + ComponentSegmentSeparator + id,
                 IOpenApiSecurityScheme => baseUri + ReferenceType.SecurityScheme.GetDisplayName() + ComponentSegmentSeparator + id,
-                _ => throw new ArgumentException($"Invalid component type {componentToRegister.GetType().Name}"),
+                _ => throw new ArgumentException($"Invalid component type {componentToRegister?.GetType().Name}"),
```
baywet, line 208: no need for null prop here, we're already checking against the component being null earlier

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
```diff
@@ -44,7 +42,7 @@ public OpenApiValidator(ValidationRuleSet ruleSet, OpenApiDocument hostDocument
         /// <summary>
         /// The host document used for validation.
         /// </summary>
-        public OpenApiDocument HostDocument { get; set; }
+        public OpenApiDocument? HostDocument { get; set; }
```
baywet, line 45: shouldn't we remove that property now that it's not affected anymore?

src/Microsoft.OpenApi/Validations/OpenApiValidator.cs
```diff
@@ -93,7 +91,7 @@ public void AddWarning(OpenApiValidatorWarning warning)
         public override void Visit(OpenApiResponses response) => Validate(response);
 
         /// <inheritdoc/>
-        public override void Visit(OpenApiExternalDocs externalDocs) => Validate(externalDocs);
+        public override void Visit(OpenApiExternalDocs? externalDocs) => Validate(externalDocs);
```
baywet, line 94: why only this method? align others or revert

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -18,9 +18,9 @@ public sealed class ValidationRuleSet
     {
         private Dictionary<Type, IList<ValidationRule>> _rulesDictionary = new();
 
-        private static ValidationRuleSet _defaultRuleSet;
+        private static ValidationRuleSet? _defaultRuleSet;
 
-        private List<ValidationRule> _emptyRules = new();
+        private readonly List<ValidationRule>? _emptyRules = new();
```
baywet, line 23: this is never null, remove the ?

src/Microsoft.OpenApi/Validations/ValidationRuleSet.cs
```diff
@@ -45,7 +45,7 @@ public ValidationRuleSet()
         /// </summary>
         /// <param name="type">The type that is to be validated</param>
         /// <returns>Either the rules related to the type, or an empty list.</returns>
-        public IList<ValidationRule> FindRules(Type type)
+        public IList<ValidationRule>? FindRules(Type type)
```
baywet, line 48: revert

src/Microsoft.OpenApi/Writers/IOpenApiWriter.cs
```diff
@@ -34,12 +34,12 @@ public interface IOpenApiWriter
         /// <summary>
         /// Write the property name.
         /// </summary>
-        void WritePropertyName(string name);
+        void WritePropertyName(string? name);
```
baywet, line 37: shouldn't the property name always be non null?

src/Microsoft.OpenApi/Services/OpenApiFilterService.cs
```diff
@@ -80,43 +80,46 @@ public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, Fun
                 },
 
                 Components = components,
-                Security = source.Security,
-                Servers = source.Servers
+                Security = source?.Security,
```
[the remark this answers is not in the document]
MaggieKimani1: Yes, source is no longer nullable. Missed this case while updating the rest. Fixing

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
MaggieKimani1: The compiler still raises a warning on dereference of a possible null reference even with the null check in place

src/Microsoft.OpenApi/Writers/IOpenApiWriter.cs
```diff
@@ -39,7 +39,7 @@ public interface IOpenApiWriter
         /// <summary>
         /// Write the string value.
         /// </summary>
-        void WriteValue(string value);
+        void WriteValue(string? value);
```
baywet, line 42: here I still don't understand why only this one would be nullable and not others? Should we revert that and have the caller type check?

src/Microsoft.OpenApi/Writers/OpenApiJsonWriter.cs
```diff
@@ -153,23 +153,23 @@ public override void WriteEndArray()
         /// </summary>
         /// <param name="name">The property name.</param>
         /// public override void WritePropertyName(string name)
-        public override void WritePropertyName(string name)
+        public override void WritePropertyName(string? name)
```
baywet, line 156: the name is non nullable in the parent interface, align and add defensive statements

src/Microsoft.OpenApi/Writers/OpenApiWriterBase.cs
```diff
@@ -99,7 +100,7 @@ protected OpenApiWriterBase(TextWriter textWriter, OpenApiWriterSettings setting
         /// <summary>
         /// Write the start property.
         /// </summary>
-        public abstract void WritePropertyName(string name);
+        public abstract void WritePropertyName(string? name);
```
baywet, line 103: name is not nullable in interface

src/Microsoft.OpenApi/Services/OpenApiWalker.cs
```diff
@@ -473,22 +476,22 @@ internal void Walk(OpenApiTagReference tag)
...
+        internal void Walk(OpenApiServer? server)
         {
             if (server == null)
             {
                 return;
             }
 
             _visitor.Visit(server);
-            Walk(OpenApiConstants.Variables, () => Walk(server.Variables));
+            Walk(OpenApiConstants.Variables, () => Walk(server?.Variables));
```
baywet, line 487:
```suggestion
            Walk(OpenApiConstants.Variables, () => Walk(server.Variables));
```

baywet (approved):
Thank you for the great work and the long haul on this one!!!

PR #2156, opened 2025-02-18
fix: refactor ToIdentifier() to normalize flaggable enums

### LINKED ISSUE #1924: v2 - type upgrade from string to flag enum - easy catch for consumers
While I support the use of a flaggable enum for performance reasons, I believe this is going to lead to a regression that's going to be hard to catch for most people. And which I think our own code is a victim of.
 
v1 Type is a string
v2 Type is a flaggable enum (currently)
 
It's super easy to not notice the change in some scenarios, especially the ones involving string interpolation, implicit operators. In that case, overlooking the fact that multiple values are present will lead to cases with multiple entries falling off a cliff.
 
I think to start with, our to identifier method should throw if multiple values are present, instead of returning null to signal something was wrong in the logic of the application https://github.com/microsoft/OpenAPI.NET/blob/cb5bb2797a4a4aa61c96d4fd453ac5257474848d/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs#L32

### LINKED ISSUE #2078: The JsonSchemaType should offer a method to produce the JSON Schema type value

The `JsonSchemaType` class is intended to represent a JSON Schema type, but its `ToString()` method currently produces a Pascal-cased string, e.g.:

```sharp
JsonSchemaType.Object.ToString() == "Object"
```
but the actual JSON schema type is lowercase "object". So if a user wants to obtain the _real_ JSON Schema type from a `JsonSchemaType` object, they currently must lowercase it themselves, and must use case-insensitive compares for type values in actual JSON schemas.

And there is the reverse problem, which is that `EnumParse` of a _correctly-cased_ JSON Schema type currently throws an exception rather than returning the appropriate `JsonSchemaType` object.

```csharp
Enum.Parse<JsonSchemaType>("object")
```

```text
Unhandled exception. System.ArgumentException: Requested value 'object' was not found.
   at System.Enum.TryParseByName[TStorage](RuntimeType enumType, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, TStorage& result)
   at System.Enum.TryParseByValueOrName[TUnderlying,TStorage](RuntimeType enumType, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, TUnderlying& result)
   at System.Enum.Parse[TEnum](String value, Boolean ignoreCase)
   at System.Enum.Parse[TEnum](String value)
   at Program.<Main>$(String[] args)
Command terminated by signal 6
```

I think the JsonSchemaType class should provide a method that returns the correctly-cased JSON Schema type. 

In addition, `Enum.Parse` of a correctly-cased JSON schema type should return the appropriate `JsonSchemaType` object.

One alternative is to change the `ToString()` method to return a correctly cased JSON Schema type value.

Another alternative is to provide a new method, such as `ToIdentifier()`, that returns the correctly cased value.

For `EnumParse`, it should accept either the Pascal-cased or correctly-cased JSON Schema type and return the appropriate `JsonSchemaType` object.

**Additional context**

[Recreate on dotnetfiddle](https://dotnetfiddle.net/GvlPQS)

[Discussion on aspnetcore PR 59480](https://github.com/dotnet/aspnetcore/pull/59480/files#r1911775847) that motivated this issue.

MaggieKimani1:
- Adds support for normalizing flaggable enums by returning an array of strings
- Supports casting of type arrays to the flaggable enum type
- Adds tests to validate

Fixes #2078
Fixes #1924

### DIFF
--- a/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
+++ b/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
@@ -72,6 +72,21 @@ namespace Microsoft.OpenApi.Extensions
             };
         }
 
+        /// <summary>
+        /// Converts a schema type's identifier into the enum equivalent
+        /// </summary>
+        /// <param name="identifier"></param>
+        /// <returns></returns>
+        public static JsonSchemaType ToJsonSchemaType(this string[] identifier)
+        {
+            JsonSchemaType type = 0;
+            foreach (var id in identifier)
+            {
+                type |= id.ToJsonSchemaType();
+            }
+            return type;
+        }
+
         private static readonly Dictionary<Type, Func<OpenApiSchema>> _simpleTypeToOpenApiSchema = new()
         {
             [typeof(bool)] = () => new() { Type = JsonSchemaType.Boolean },
--- a/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
+++ b/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
@@ -59,8 +59,7 @@ namespace Microsoft.OpenApi.Extensions
         /// <returns></returns>
         internal static string ToFirstIdentifier(this JsonSchemaType schemaType)
         {
-            var identifier = schemaType.ToIdentifiers();
-            return identifier[0];
+            return schemaType.ToIdentifiersInternal().First();
         }
 
         /// <summary>
--- a/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
+++ b/src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
@@ -38,15 +38,18 @@ namespace Microsoft.OpenApi.Extensions
         {
             return schemaType.ToIdentifiersInternal().ToArray();
         }
-        private static readonly Dictionary<JsonSchemaType, string> allSchemaTypes = [
-            { JsonSchemaType.Boolean, "boolean"},
+
+        private static readonly Dictionary<JsonSchemaType, string> allSchemaTypes = new()
+        {
+            { JsonSchemaType.Boolean, "boolean" },
             { JsonSchemaType.Integer, "integer" },
             { JsonSchemaType.Number, "number" },
             { JsonSchemaType.String, "string" },
             { JsonSchemaType.Object, "object" },
             { JsonSchemaType.Array, "array" },
-            { JsonSchemaType.Null, "null" },
-        ];
+            { JsonSchemaType.Null, "null" }
+        };
+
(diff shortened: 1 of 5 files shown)


baywet (changes requested):

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -152,21 +185,23 @@ public static Type MapOpenApiPrimitiveTypeToSimpleType(this OpenApiSchema schema
             {
                 throw new ArgumentNullException(nameof(schema));
             }
+            var isNullable = (schema.Type & JsonSchemaType.Null) == JsonSchemaType.Null;
+            var nonNullable = (schema.Type & ~JsonSchemaType.Null)?.FirstIdentifier();
```
baywet, line 189: can we remove the toIdentifier part here and use enum and not string comparisons in the switch?

baywet (changes requested):

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -33,20 +34,32 @@ public static class OpenApiTypeMapper
         /// </summary>
         /// <param name="schemaType"></param>
         /// <returns></returns>
-        public static string? ToIdentifier(this JsonSchemaType schemaType)
+        public static string[] ToIdentifier(this JsonSchemaType schemaType)
```
darrelmiller, line 37:
```suggestion
        public static string[] ToIdentifiers(this JsonSchemaType schemaType)
```

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
```diff
@@ -33,20 +34,32 @@ public static class OpenApiTypeMapper
...
+
+            return types.ToArray();
         }
+
+        /// <summary>
+        /// Returns the first identifier from a string array.
+        /// </summary>
+        /// <param name="schemaType"></param>
+        /// <returns></returns>
+        internal static string FirstIdentifier(this JsonSchemaType schemaType)
```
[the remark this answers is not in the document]
darrelmiller: Also add ToSingleIdentifier that will throw if there are more than one value.

src/Microsoft.OpenApi/Extensions/OpenApiTypeMapper.cs
MaggieKimani1: Yeah good call

baywet (changes requested):
A couple of additional remarks for performance.

Note: we could also use frozen dictionary for the static reference but I'm not sure the extra dependency is worth it.

PR #2160, opened 2025-02-20
fix: primitive parsing for strings as DateTimes is too greedy

### LINKED ISSUE #2137: v2 - Primitive parsing for strings as DateTimes is too greedy
Both @mikekistler and I have observed this bug while interacting with the new JsonNode-based example APIs in v2. The current logic tries to eagerly parse string JsonNodes as date times and write them as such when emitting the final document:

https://github.com/microsoft/OpenAPI.NET/blob/e917e4cb89243e9345899a1118f42dd5b8845ee2/src/Microsoft.OpenApi/Writers/OpenApiWriterAnyExtensions.cs#L118-L144

This can cause some rather unintended behaviors. For example, the current implementation casts example strings that are designated as a `DateOnly` time to `DateTime` based types. The current implementation will also errenously capture certain string examples (like a float `3.14`) as dates.

The following code:

```csharp
using System.Text.Json.Nodes;
using System.IO;

var schema = new OpenApiSchema()
{
	Type = JsonSchemaType.String,
	Example = JsonValue.Create("2024-01-02")
};

var schema2 = new OpenApiSchema()
{
	Type = JsonSchemaType.String,
	Example = JsonValue.Create("3.14")
};

var textWriter = new StringWriter();
var writer = new OpenApiJsonWriter(textWriter);
schema.SerializeAsV31(writer);
System.Console.WriteLine(textWriter.ToString());

textWriter = new StringWriter();
writer = new OpenApiJsonWriter(textWriter);
schema2.SerializeAsV31(writer);
System.Console.WriteLine(textWriter.ToString());
```

Produces:

```
{
  "type": "string",
  "example": "2024-01-02T00:00:00.0000000+00:00"
}
{
  "type": "string",
  "example": "2025-03-14T00:00:00.0000000+00:00"
}
```

When I would expect it to produce:

```
{
  "type": "string",
  "example": "2024-01-02"
}
{
  "type": "string",
  "example": "3.14"
}
```

MaggieKimani1:
Fixes #2137


test/Microsoft.OpenApi.Readers.Tests/V31Tests/OpenApiSchemaTests.cs
```diff
@@ -542,6 +542,43 @@ public void ParseSchemaWithUnrecognizedKeywordsWorks()
...
+}";
+
+            var expected2 = @"{
+  ""type"": ""string"",
+  ""example"": ""3.14""
+}";
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String,
+                Example = JsonValue.Create("2024-01-02")
```
baywet, line 560: we should also have a unit test where the actual instance is a DateTime instance. As I'm not sure TryGetValue in that case will return a string. See https://github.com/microsoft/OpenAPI.NET/pull/2157/files#diff-0f3a7509d4954ac2bbc88c7eb76ef6dca6b526cef456fa0d7610013421ce7f96R77 for more context

src/Microsoft.OpenApi/Writers/OpenApiWriterAnyExtensions.cs
```diff
@@ -109,72 +109,27 @@ private static void WriteObject(this IOpenApiWriter writer, JsonObject entity)
...
-            }
-            else if (valueKind is JsonValueKind.True)
-            {
-                writer.WriteValue(true);
-            }
+            if (jsonValue.TryGetValue(out string stringValue))
+                writer.WriteValue(stringValue);
+            else if (jsonValue.TryGetValue(out DateTime dateTimeValue))
+                writer.WriteValue(dateTimeValue.ToString("o", CultureInfo.InvariantCulture)); // ISO 8601 format
+            else if (jsonValue.TryGetValue(out DateTimeOffset dateTimeOffsetValue))
```
baywet, line 118: you could also add support for date and time only with conditional compilation directives NET6_OR_GREATER

This way people using those types will be supported as well.

src/Microsoft.OpenApi/Writers/OpenApiWriterAnyExtensions.cs
```diff
@@ -109,72 +109,31 @@ private static void WriteObject(this IOpenApiWriter writer, JsonObject entity)
...
+                writer.WriteValue(dateTimeOffsetValue.ToString("o", CultureInfo.InvariantCulture));
+#if NET6_0_OR_GREATER
+            else if (jsonValue.TryGetValue(out DateOnly dateOnlyValue))
```
baywet, line 121: and time only please :)

MaggieKimani1: Oh I didn't think those were two separate types. I understood DateOnly 😄

baywet (approved):

PR #2165, opened 2025-02-25
fix/static factories

### LINKED ISSUE #2022: v2 - get rid of OpenApiReaderRegistry
Having a static instance with concurrent dictionaries is going to at best negatively impact performance, at worst introduce race conditions.

In addition to that, the API itself it not obvious.

All the places where we call "GetReader" from, are places where we have access to the settings.

I suggest we nuke that class and make the readers a property of the settings class instead, with defaults.

baywet:
fixes #2022


src/Microsoft.OpenApi.Readers/OpenApiReaderSettingsExtensions.cs
```diff
@@ -0,0 +1,33 @@
+using Microsoft.OpenApi.Interfaces;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers;
+
+namespace Microsoft.OpenApi.Reader;
```
MaggieKimani1, line 5: Out of curiosity, should this class exist in a different namespace as the YamlReader?

src/Microsoft.OpenApi.Readers/OpenApiReaderSettingsExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    public static void AddYamlReader(this OpenApiReaderSettings settings)
+    {
+        var yamlReader = new OpenApiYamlReader();
+        settings.AddReaderToSettings(OpenApiConstants.Yaml, yamlReader);
+        settings.AddReaderToSettings(OpenApiConstants.Yml, yamlReader);
+    }
+    private static void AddReaderToSettings(this OpenApiReaderSettings settings, string format, IOpenApiReader reader)
+    {
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP || NET5_0_OR_GREATER
+        settings.Readers.Add(format, reader);
```
MaggieKimani1, line 25: Should we use `TryAdd` instead or are we keen on an exception being thrown if the key already exists in the dictionary?

src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
```diff
@@ -239,14 +234,15 @@ public static T Parse<T>(string input,
...
         }
 
-        private static readonly OpenApiReaderSettings DefaultReaderSettings = new();
+        private static readonly Lazy<OpenApiReaderSettings> DefaultReaderSettings = new(() => new OpenApiReaderSettings());
 
         private static async Task<ReadResult> InternalLoadAsync(Stream input, string format, OpenApiReaderSettings settings, CancellationToken cancellationToken = default)
         {
-            var reader = OpenApiReaderRegistry.GetReader(format);
+            settings ??= DefaultReaderSettings.Value;
+            var reader = settings.Readers[format];
```
andrueastman, line 242: By removing the registry, one thing I believe we may miss out on is that if a reader(the key) is absent, the error that the library user will get may not be too useful/helpful(they will get a [KeyNotFoundException](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.keynotfoundexception?view=net-9.0)).
Would it make sense to add a getReader method (can be private/internal) so as to throw a more useful error for such a scenario where the settings isn't setup right?

src/Microsoft.OpenApi.Readers/OpenApiReaderSettingsExtensions.cs
andrueastman: Sorry. Ignore my previous comment. Looks like it will still throw in latest versions of NET according to the docs. I was under the impression the behaviour changed. 
https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.add?view=net-9.0

baywet: Thanks, that was an oversight, I meant to use try add here hence the conditional compilation

src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
baywet: Added a method to throw more explicit exceptions, which also lead me to add a lot of additional unit tests. Thanks for pointing this out.

andrueastman (approved):

PR #2166, opened 2025-02-25
BREAKING CHANGE: Rename Microsoft.OpenApi.Readers to Microsoft.OpenApi.YamlReader

### LINKED ISSUE #1957: [API review] Rename Microsoft.OpenAPI.readers to Microsoft.OpenAPI.YamlReader
With the move to reading via System.text.json, the OpenAPI.readers only supports reading YAML files. The name should suggest this.

MaggieKimani1:
This PR aims to:
- Rename the `Microsoft.OpenApi.Readers` to `Microsoft.OpenApi.YamlReader`  for it to be more intuitive and align with the package's role.
- This is in line with the breaking change involving migration of the loading code into the core lib to avoid clients having to take a dependency on an external YAML library.
- Those working with YAML files will need to take a dependency on this new lib as `Microsoft.OpenApi.Readers` will become deprecated in `v2.0`.

Fixes #1957


src/Microsoft.OpenApi.YamlReader/Microsoft.OpenApi.YamlReader.csproj
andrueastman, line 1: Lets update the description field to say that this package only reads YAML

andrueastman (approved):

PR #2172, opened 2025-02-26
fix: deduplicates exclusive min/max properties in the object model

### LINKED ISSUE #1962: Upcast/downcast old ExclusiveMin/Max
## Upcast from 2/3 to 3.1

If ExclusiveMin/Max is true
   Copy Min/Max value to ExclusiveMin/Max value and remove Min/Max keyword

if ExclusiveMin/Max is false
  do nothing

## Downcast from 3.1 to earlier

if ExclusiveMin/Max has non-null value
  Write ExclusiveMin/Max as true and write min/max value

baywet:
fixes #1962


MaggieKimani1:
Thanks for taking this on!

MaggieKimani1 (approved):

PR #2174, opened 2025-02-27
Rename `OpenApiDocument.SecurityRequirements` as `Security`. Fixes #2155

### LINKED ISSUE #2155: `OpenApiDocument.SecurityRequirements` inconsistently named `SecurityRequirements`, instead of `Security` per the OpenAPI spec
**Describe the bug**
_Extremely nitpicky_ 😅, but I noticed that the `OpenApiDocument` property for the [`security` field of the OpenAPI object](https://spec.openapis.org/oas/v3.1.0.html#oasSecurity) is named `SecurityRequirements`, instead of `Security`.

https://github.com/microsoft/OpenAPI.NET/blob/e7c74b8b66ce6a43935fc5c6aca8a88ba6c75da2/src/Microsoft.OpenApi/Models/OpenApiDocument.cs#L73-L77

In general, the models & their properties seem to be named after their spec counterparts (but I haven't exhaustively searched for other inconsistencies).

While I wouldn't normally raise issues for issues as minor as this, I thought I'd make this one in this instance because renaming this property would be a breaking change, and v2 is an upcoming _major_ release; so if this is deemed worth fixing, I'd imagine now would be the time for it? (or in v3 hehe). I guess it could also be fixed in a backward-compatible way by renaming the property to `Security` and adding an `[Obsolete] SecurityRequirements => Security`?

**Expected behavior**
Presence of a `Security` property in `OpenApiDocument`, but it is named `SecurityRequirements` instead.

**Additional context**
The `OpenApiOperation` property for the [`security` field of the Operation object](https://spec.openapis.org/oas/v3.1.0.html#operationSecurity) is named `Security`, as expected.

https://github.com/microsoft/OpenAPI.NET/blob/e7c74b8b66ce6a43935fc5c6aca8a88ba6c75da2/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L91-L98

ravindUwU:
fixes #2155

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
@@ -73,7 +73,7 @@ namespace Microsoft.OpenApi.Models
         /// <summary>
         /// A declaration of which security mechanisms can be used across the API.
         /// </summary>
-        public IList<OpenApiSecurityRequirement>? SecurityRequirements { get; set; } =
+        public IList<OpenApiSecurityRequirement>? Security { get; set; } =
             new List<OpenApiSecurityRequirement>();
 
         private HashSet<OpenApiTag>? _tags;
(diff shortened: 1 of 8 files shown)


ravindUwU:

baywet (approved):
Thanks for the contribution!

Assuming it was left in draft because of the question in the issue, which I have replied to and confirmed no additional changes are required at this point. I'll switch it to ready.

PR #2213, opened 2025-02-28
Task: Migrate ACR Push pipeline to Azure Pipelines

Onokaev:
https://mcr.microsoft.com/en-us/artifact/mar/openapi/hidi/tags
Deploys as before


baywet (approved):
Thanks for the contribution!

Issue #2233, opened 2025-03-04
[V2] Handing of Tags on reading and trimming openApi documents


andrueastman:
There are a couple of issues with handling tags in V2 preview causing issues with plugin generation on Kiota when producing a trimmed openApi document.

#### 1. The `CopyReferences` walker does not copy over instances of `OpenApiTagReference`

Taking a look at referenceable items here
https://github.com/microsoft/OpenAPI.NET/blob/54b320231a2874febf09e15c14aeb370a2d71f19/src/Microsoft.OpenApi/Services/CopyReferences.cs#L19

Instances of `OpenApiTagReference` are not copied over to the new document.

Therefore, using this method with an openApiDocument leaves out the tags that may have been referenced in an openApi document. 

https://github.com/microsoft/OpenAPI.NET/blob/54b320231a2874febf09e15c14aeb370a2d71f19/src/Microsoft.OpenApi/Services/OpenApiFilterService.cs#L62

#### 2. `OpenApiTag`s in `OpenApiOperation`s are deserialized from the root document as `UnresolvedReference`s

Taking a look at the deserializer here
https://github.com/microsoft/OpenAPI.NET/blob/54b320231a2874febf09e15c14aeb370a2d71f19/src/Microsoft.OpenApi/Reader/V3/OpenApiOperationDeserializer.cs#L24

The openApi tag is loaded up as a reference by calling the constructor but not adding the tag to the root of the document. This therefore causes the Target of the reference to be null according to the definition below. 

https://github.com/microsoft/OpenAPI.NET/blob/a45793552aa47edfc41893690771766fca4a1f60/src/Microsoft.OpenApi/Models/References/OpenApiTagReference.cs#L20

Therefore, when we have a document trimmed with the `OpenApiFilterService`, any tags that existed in the document at the operation level will be references with invalid/unresolved targets. Trying to write the document will result to a null reference due to the null target as defined here. 

https://github.com/microsoft/OpenAPI.NET/blob/a45793552aa47edfc41893690771766fca4a1f60/src/Microsoft.OpenApi/Models/References/BaseOpenApiReferenceHolder.cs#L112

PR #2235, opened 2025-03-07
fix: fixes serialization of openApidocs with operation tags with settings to inline references

andrueastman:
fix: fixes serialization of openApidocs with operation tags with settings to inline references

Fixes #https://github.com/microsoft/OpenAPI.NET/issues/2233

### DIFF
--- a/test/Microsoft.OpenApi.Tests/Models/References/OpenApiTagReferenceTest.cs
+++ b/test/Microsoft.OpenApi.Tests/Models/References/OpenApiTagReferenceTest.cs
@@ -45,6 +45,7 @@ paths:
           description: The user was not found.
       tags:
         - $ref: '#/tags/user'
+        - users.user
 components:
   schemas:
     User:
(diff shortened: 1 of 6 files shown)


MaggieKimani1 (approved):

test/Microsoft.OpenApi.Tests/Models/References/OpenApiTagReferenceTest.cs
```diff
@@ -45,6 +45,7 @@ public class OpenApiTagReferenceTest
           description: The user was not found.
       tags:
         - $ref: '#/tags/user'
+        - users.user
```
baywet, line 48: I believe The case above with $ref is invalid? 
Unless that's like a negative test to make sure the ref syntax does not get caught by some logic when we don't want it to?

andrueastman: I think I understand what you mean here. The name of the `tag` should not be a ref to the tag at the root of the document, but should simply match the name as tags at the operation level are defined as a list of strings. 
Yes?

PR #2237, opened 2025-03-07
chore: add test for clarity of tags parsing

andrueastman:


baywet (approved):
thanks for making this clear in the test case!

Issue #1674, opened 2024-05-27
Relative reference in subdirectory OpenApi document fails to load


dldl-cmd:
# Relative reference in subdirectory OpenApi document fails to load

When loading a document which is splitted into multiple documents relative references are failing to load when a parent document is in a subdirectory.

What happens is, that a `System.IO.FileNotFoundException` is thrown:

```
Could not find file '<path to root of document structure>\Pets.json'.
StackTrace:
   at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.Strategies.FileStreamHelpers.ChooseStrategyCore(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.File.OpenRead(String path)
   at Microsoft.OpenApi.Readers.Services.DefaultStreamLoader.<LoadAsync>d__4.MoveNext() in C:\Users\user\source\repos\OpenAPI.NET\src\Microsoft.OpenApi.Readers\Services\DefaultStreamLoader.cs:line 47
   at Microsoft.OpenApi.Readers.Services.OpenApiWorkspaceLoader.<LoadAsync>d__4.MoveNext() in C:\Users\user\source\repos\OpenAPI.NET\src\Microsoft.OpenApi.Readers\Services\OpenApiWorkspaceLoader.cs:line 46
   at Microsoft.OpenApi.Readers.OpenApiYamlDocumentReader.<ReadAsync>d__3.MoveNext() in C:\Users\user\source\repos\OpenAPI.NET\src\Microsoft.OpenApi.Readers\OpenApiYamlDocumentReader.cs:line 105
[...]

PR #2243, opened 2025-03-10
Fix: relative references in subdirectory documents are not loading #1674

dldl-cmd:
Use OpenApiDocuments BaseUri as location of the document. This allows to have during loading further documents a base Url for retrieval, which can be combined with a relative Uri to get an absolute.

### DIFF
--- a/src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
+++ b/src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
@@ -18,7 +18,30 @@ namespace Microsoft.OpenApi.Services
     {
         private readonly Dictionary<string, Uri> _documentsIdRegistry = new();
         private readonly Dictionary<Uri, Stream> _artifactsRegistry = new();        
-        private readonly Dictionary<Uri, IOpenApiReferenceable> _IOpenApiReferenceableRegistry = new();
+        private readonly Dictionary<Uri, IOpenApiReferenceable> _IOpenApiReferenceableRegistry = new(new UriWithFragmentEquailityComparer());
+
+        private class UriWithFragmentEquailityComparer : IEqualityComparer<Uri>
+        {
+            public bool Equals(Uri x, Uri y)
+            {
+                if (ReferenceEquals(x, y))
+                {
+                    return true;
+                }
+
+                if (x is null || y is null)
+                {
+                    return false;
+                }
+
+                return x.AbsoluteUri == y.AbsoluteUri;
+            }
+
+            public int GetHashCode(Uri obj)
+            {
+                return obj.AbsoluteUri.GetHashCode();
+            }
+        }
 
         /// <summary>
         /// The base location from where all relative references are resolved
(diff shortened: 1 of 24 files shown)


baywet:
Thanks for the contribution!

@maggieKimani1 can you do an initial review here please?

MaggieKimani1:
@dldl-cmd sorry for the delay in response/review.
I have checked out the issue that this PR resolves and it seems that this is by design.
Have you considered creating a custom external refs loader, an implementation of our [IStreamLoader interface](https://github.com/microsoft/OpenAPI.NET/blob/main/src/Microsoft.OpenApi/Interfaces/IStreamLoader.cs) as per Darrel's [comment here](https://github.com/microsoft/OpenAPI.NET/issues/1674#issuecomment-2669030481)?
Our [default stream reader](https://github.com/microsoft/OpenAPI.NET/blob/main/src/Microsoft.OpenApi/Reader/Services/DefaultStreamLoader.cs) class is a bit limited as file/resource location might vary depending on use case, hence the exposed interface that you can use to resolve the relative references especially for documents contained in subdirectories, and pass the custom loader as part of the `OpenApiReaderSettings` as shown here https://github.com/microsoft/OpenAPI.NET/blob/ff287bce5844b0dbdb2a6ad5c062d747a8ab77e8/test/Microsoft.OpenApi.Readers.Tests/OpenApiWorkspaceTests/OpenApiWorkspaceStreamTests.cs#L53

**Reference**: 
https://github.com/microsoft/OpenAPI.NET/blob/ff287bce5844b0dbdb2a6ad5c062d747a8ab77e8/test/Microsoft.OpenApi.Readers.Tests/OpenApiWorkspaceTests/OpenApiWorkspaceStreamTests.cs#L85-L98

dldl-cmd:
@MaggieKimani1 implementing just an own stream loader cannot solve the problem. The reason is, that the stream loader has no idea, which document is the current parent and where this document is located. Therefore it cannot resolve the relative reference as it doesn't know to which location it is relative. An example:

- root.yaml
```yaml
....
$ref: ./DirA/DirB/DirC/second.yaml#/components/schema/Second
```
- DirA/DirB/DirC/second.yaml
```yaml
...
$ref: ../third.yaml#/components/schema/Third
```

- DirA/DirB/third.yaml
```yaml
...
$ref: ./DirC/Fourth.yaml#/components/schema/Fourth.yaml
```

When the reference `$ref: ./DirC/Fourth.yaml#/components/schema/Fourth.yaml` should be resolved. It is necessary to known that this reference is relative to the document at `DirA/DirB/third.yaml`.

MaggieKimani1:
Also do you mind resolving the conflicts?

MaggieKimani1 (changes requested):

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
```diff
@@ -224,12 +247,13 @@ public void AddDocumentId(string key, Uri value)
             }
         }
 
+#nullable enable
```
MaggieKimani1, line 250: Now that we have NRT configured, we can remove this directive

dldl-cmd:
> Also do you mind resolving the conflicts?

Sorry for my late reply, I was unavailable a few days. I didn't notice the merge conflict until now. I have started to resolve them, but I need to check at least one conflict more in detail. Hopefully I can resolve it in the next days.

MaggieKimani1 (approved):

PR #2251, opened 2025-03-13
feat: use http method object instead of enum

### LINKED ISSUE #2068: v2 consider making http method string constants instead of an enum
For v3.2 [the board is considering open ended list of methods to be supported](https://github.com/OAI/OpenAPI-Specification/issues/1747)

If we want to avoid shipping a breaking change then, we might as well go ahead and use string constants instead.

Of course, we might always decide that we don't want to delay v2 shipping further, and want to deal with that then.

@darrelmiller input would be interesting here.

Michael-Wamae:
feat: use http method object instead of enum

Fixes #2068


baywet (dismissed):
Great work here!

baywet (approved):

PR #2273, opened 2025-03-18
fix : remove duplicate unused property

Poltuu:


baywet (approved):
Thanks for the contribution!

PR #2275, opened 2025-03-18
fix: Include hidi in solution

MaggieKimani1:
Reverts the previous change to exclude hidi in the solution file and CI in order to fix ambiguity in package source mapping references.


MaggieKimani1:
Holding for the OData release https://github.com/microsoft/OpenAPI.NET.OData/pull/669

baywet (approved):

PR #2280, opened 2025-03-20
chore: update-discriminator-mapping

### LINKED ISSUE #2140: v2 - discriminator mappings should be dictionary<string, OpenApiSchemaReference>
Now that we've fixed the references behaviour, I believe that Discriminator mappings should be a `Dictionary<string, OpenApiSchemaReference>` instead of `<string, string>`

so currently

```csharp
Discriminator = new()
            {
                PropertyName = "@odata.type",
                Mapping = new Dictionary<string, string> {
                    {
                        "#microsoft.graph.directoryObject", "#/components/schemas/microsoft.graph.directoryObject"
                    },
                    {
                        "#microsoft.graph.file", "#/components/schemas/microsoft.graph.file"
                    }
                }
            },
```

tomorrow

```csharp
Discriminator = new()
            {
                PropertyName = "@odata.type",
                Mapping = new Dictionary<string, string> {
                    {
                        "#microsoft.graph.directoryObject", new OpenApiSchemaReference("microsoft.graph.directoryObject")
                    },
                    {
                        "#microsoft.graph.file", new OpenApiSchemaReference("microsoft.graph.file")
                    }
                }
            },

```

There are a number of reasons for doing that:
1. the syntax of the references (`#/components/schemas`) varies between v2 and v3, and might vary in v4. And we shouldn't expect the user to know about that/have to change it between different serialization versions.
2. using references would bring the resolution logic in place which would not only improve checking for document integrity but also let the user navigate to the target from the object model.

Before we do any work on that, I'd like @darrelmiller to chime in.

Michael-Wamae:
- chore: update-discriminator-mapping 

fixes #2140

### DIFF
--- a/test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiDiscriminatorTests.cs
+++ b/test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiDiscriminatorTests.cs
@@ -1,10 +1,15 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+using System;
+using System.Diagnostics;
 using System.IO;
 using System.Threading.Tasks;
 using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Models.References;
 using Microsoft.OpenApi.Reader;
+using Microsoft.VisualStudio.TestPlatform.Utilities;
+using Newtonsoft.Json;
 using Xunit;
 
 namespace Microsoft.OpenApi.Readers.Tests.V3Tests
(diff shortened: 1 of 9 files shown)


MaggieKimani1 (commented):
Can we please add a test case for when the value in the mapping dictionary is a URI reference pointing to a schema in an external file as in the spec:
image

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiDiscriminatorTests.cs
```diff
@@ -25,19 +27,20 @@ public async Task ParseBasicDiscriminatorShouldSucceed()
...
-                    Mapping =
+               new OpenApiDiscriminator
+               {
+                   PropertyName = "pet_type",
+                   Mapping =
                     {
-                            ["puppy"] = "#/components/schemas/Dog",
-                            ["kitten"] = "Cat"
+                            ["puppy"] = new OpenApiSchemaReference("Dog", openApiDocument),
+                            ["monster"] = new OpenApiSchemaReference("schema.json" , openApiDocument, "https://gigantic-server.com/schemas/Monster/schema.json")
```
baywet, line 41:
```suggestion
                            ["monster"] = new OpenApiSchemaReference("Sully" , openApiDocument, "https://gigantic-server.com/schemas/Monster/schema.json")
```

Just for sanity, so we differentiate the schema name from the file. I suspect this is going to impact your test file.

Michael-Wamae: In this case, wouldn't the test fall under [ExternalSchemaReference Tests](https://github.com/microsoft/OpenAPI.NET/blob/main/test/Microsoft.OpenApi.Readers.Tests/V31Tests/OpenApiComponentsTests.cs) where we add this scenario `[InlineData("https://gigantic-server.com/schemas/Monster/schema.json#/components/schemas/Sully", "Sully", "https://gigantic-server.com/schemas/Monster/schema.json")]`

baywet: let's keep the original then and ALSO add this test case, this way we're covered for:
- internal
- external single component files
- external multiple component files
what do you think?

test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiDiscriminatorTests.cs
```diff
@@ -38,6 +38,7 @@ public async Task ParseBasicDiscriminatorShouldSucceed()
                    Mapping =
                     {
                             ["puppy"] = new OpenApiSchemaReference("Dog", openApiDocument),
+                            ["kitten"] = new OpenApiSchemaReference("Cat" , openApiDocument, "Cat"),
```
baywet, line 41:
```suggestion
                            ["kitten"] = new OpenApiSchemaReference("Cat" , openApiDocument, "https://gigantic-server.com/schemas/animals.json"),
```

Sorry, this is what I meant.

baywet (approved):

MaggieKimani1 (approved):

PR #2284, opened 2025-03-24
feat: Remove default collection initialization for perf reasons

### LINKED ISSUE #1971: [API Review] Memory performance improvements
Currently we instantiate new objects every time for the following properties, this can have a large memory impact for large OpenAPI descriptions. For eg: For the Graph OpenAPI this will create 142k schemas, for a total of 1.1 million objects that might be unused. So we have an opportunity for some memory perf. improvements by lazily instantiating these. 
``` 
 public virtual IDictionary<string, OpenApiSchema> PatternProperties { get; set; } = new Dictionary<string, OpenApiSchema>();
        public virtual IList<JsonNode> Enum { get; set; } = new List<JsonNode>();
        public virtual IList<OpenApiSchema> AllOf { get; set; } = new List<OpenApiSchema>();
        public virtual IList<OpenApiSchema> OneOf { get; set; } = new List<OpenApiSchema>();
        public virtual IList<OpenApiSchema> AnyOf { get; set; } = new List<OpenApiSchema>();
        public virtual ISet<string> Required { get; set; } = new HashSet<string>();
        public virtual IDictionary<string, OpenApiSchema> Properties { get; set; } = new Dictionary<string, OpenApiSchema>();
        public virtual IDictionary<string, IOpenApiExtension> Extensions { get; set; } = new Dictionary<string, IOpenApiExtension>();
```

A couple of different ideas were identified: 
1. Lazy Instantiation: Explore using the init pattern and lazy get instantiation to reduce unnecessary object allocations in the `OpenAPISchema`
2. Structs vs. Classes: Evaluate the possibility of redefining `OpenAPISchema` and other types as structs instead of classes to improve performance. 
3. `IDictionary` properties should have a default value and no setter.

MaggieKimani1:
- Uses the Lazy<T> pattern to defer initialization of collections till the first time they are accessed.
- This helps reduce unnecessary resource/memory allocations

Fixes #1971


src/Microsoft.OpenApi/Models/OpenApiCallback.cs
```diff
@@ -13,17 +13,26 @@ namespace Microsoft.OpenApi.Models
...
 
+        private Lazy<IDictionary<string, IOpenApiExtension>>? _extensions = new(() => new Dictionary<string, IOpenApiExtension>(StringComparer.Ordinal));
 
         /// <summary>
         /// This object MAY be extended with Specification Extensions.
         /// </summary>
-        public IDictionary<string, IOpenApiExtension>? Extensions { get; set; } = new Dictionary<string, IOpenApiExtension>();
+        public IDictionary<string, IOpenApiExtension>? Extensions
+        {
+            get => _extensions?.Value;
```
baywet, line 33: by definition, _extensions is never null, so the null propagation operator is not necessary.

This is only going to defer memory usage, we'll always return the lazily loaded value upon serialization.
Now that we have NRT in place, it should be safer to... simply not initialize the values. (so remove the lazy all together)

src/Microsoft.OpenApi/Models/OpenApiCallback.cs
```diff
@@ -13,17 +13,26 @@ namespace Microsoft.OpenApi.Models
...
+            set => _extensions = value is null ? null : new(() => value);
```
baywet, line 34: This copy does not have the same comparer. See what I've done for OpenApiReaderSettings.Readers

baywet (changes requested):
For those sets of changes, it's always better to do it at one place with a draft, have a conversation, and then replicate all over.

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
```diff
@@ -382,8 +383,9 @@ public static IOpenApiSchema LoadSchema(ParseNode node, OpenApiDocument hostDocu
                 {
                     propertyNode.ParseField(schema, _openApiSchemaFixedFields, _openApiSchemaPatternFields, hostDocument);
                 }
-                else if (schema.UnrecognizedKeywords is not null && propertyNode.JsonNode is not null)
+                else if (propertyNode.JsonNode is not null)
                 {
+                    schema.UnrecognizedKeywords ??= new Dictionary<string, JsonNode>(StringComparer.Ordinal);
```
baywet, line 388: this is a problem I'm also facing in YOKO:
1. collection is null and needs to be initialized
2. which comparer to use is implementation details of the target class (here OpenApiSchema)
3. I have to go spelunking and do deep thinking about what to use

I don't have a cleaner solution to that than adding a getter (should it be static?) in OpenAPISchema like this:


```csharp
public static Dictionary<string, JsonNode> InitialUnrecognizedKeywordsValue => new Dictionary<string, JsonNode>(StringComparer.Ordinal);
```

This way the consumer can do that

```csharp
schema.UnrecognizedKeywords ??= OpenApiSchema.InitialUnrecognizedKeywordsValue;
```

The reason I'm suggesting that, and NOT a method like "void InitializeUnrecognizedKeywordsValue()" is because that method would not be recognized by NRT analysis.

Thoughts? CC @darrelmiller

darrelmiller: Do I have to set that schema UnrecognizedKeywords property for every schema object?  Is there no way to do this globally?

test/Microsoft.OpenApi.Tests/Validations/OpenApiSchemaValidationTests.cs
```diff
@@ -115,7 +115,7 @@ public void ValidateDefaultShouldNotHaveDataTypeMismatchForComplexSchema()
             var schema = new OpenApiSchema
             {
                 Type = JsonSchemaType.Object,
-                Properties =
+                Properties = new Dictionary<string, IOpenApiSchema>
```
darrelmiller, line 118: Does the shortcut syntax work where you only put "new" and the type is inferred?

darrelmiller:
>  remove default collection initialization altogether

Looks like I missed this conversation. Is there a summary of why we want to force users to initialize these collections manually?  Is ASP.NET team ok with this change?

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
baywet: Maybe via some kind of factories. What experience do you want to achieve?

baywet:
If we allocate by default with field initialization (current implementation), we allocate the memory when the object gets created. 

If we lazy initialize the collection: 
- we allocate memory we're going to discard right away on deserialization
- we have no way of telling the difference between empty and null

I believe the reason why we allocated on initialization historically was because: 
- nrt was not a thing, so there was no way for the consumers to get help from the compiler for those scenarios
- collection expression syntax didn't exist, making the syntax verbose. 

In addition to all that, in a lot of cases, the consumers end up doing object initialization syntax, which means they are initializing the collections anyway. 

For all those reasons I suggest we stick to the idiomatic expressions of the platform, and leave it null, letting the compiler help consumers while reducing memory usage.

MaggieKimani1:
> > remove default collection initialization altogether
> 
> Looks like I missed this conversation. Is there a summary of why we want to force users to initialize these collections manually? Is ASP.NET team ok with this change?

cc: @captainsafia

test/Microsoft.OpenApi.Tests/Validations/OpenApiSchemaValidationTests.cs
MaggieKimani1: No that doesn't work. 
Properties is an implementation of the `IDictionary` interface https://github.com/microsoft/OpenAPI.NET/blob/5d99cc422c7646ec74528287003c6f449e421eb9/src/Microsoft.OpenApi/Models/OpenApiSchema.cs#L203  so you have to explicitly define the concrete type here.

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
MaggieKimani1: @baywet would we have to implement the getters for all collections now that we're defaulting to null?

baywet: we the current implementation of this PR, no, null IS the default when nothing is assigned :)

darrelmiller:
> we allocate memory we're going to discard right away on deserialization
I had not considered this.  And I don't think we should change deserialization to leverage existing collections.

> we have no way of telling the difference between empty and null
I think most of the time that isn't an important distinction to make.  I think there are only a couple of collections in the model where that is significant. e.g. Paths and security requirements.

> the reason why we allocated on initialization historically was because:
The reason was so that a consumer editing/creating a document did not have instantiate a type that they may or may not know what that type is.  e.g. sometimes we use dictionary<string,foo> sometimes we have dedicated classes  OpenAPIResponses.

Eight years ago the object initialization syntax was not as developed and wasn't used nearly as much.

> stick to the idiomatic expressions of the platform

Is it?  I agree that C# has been introducing new ways to initialize objects/arrays/collections over the past few years, but is there some official design guidance that this is idiomatic?

Anyway, I think because the support for object initialization is better and because we want to optimize perf for the common case of deserialization, and because we now have NRT, I am more comfortable moving to a model where we don't lazy instantiate.

src/Microsoft.OpenApi/Services/CopyReferences.cs
```diff
@@ -99,7 +100,8 @@ private void AddParameterToComponents(IOpenApiParameter? parameter, string? refe
     {
         EnsureComponentsExist();
         EnsureParametersExist();
-        if (Components.Parameters is not null && parameter is not null && referenceId is not null && !Components.Parameters.ContainsKey(referenceId))
+        Components.Parameters ??= new Dictionary<string, IOpenApiParameter>();
```
baywet, line 103: I'd move the assignment within the condition, and use null prop operator in the condition, this way we don't assign if the other members of the condition are false.

baywet (commented):
This is great progress!

One thing I also wanted us to discuss but forgot to mention earlier: Why are collections returning their interface type? What benefit does it provide? the obvious drawback is that it prevents from using the collection initialization syntax `[]`, especially for dictionaries as we can see in the unit tests. CC @darrelmiller

test/Microsoft.OpenApi.Tests/Validations/OpenApiSchemaValidationTests.cs
baywet: but we could change the property type as well. (see my latest comment)

MaggieKimani1:
> One thing I also wanted us to discuss but forgot to mention earlier: Why are collections returning their interface type? What benefit does it provide? the obvious drawback is that it prevents from using the collection initialization syntax `[]`, especially for dictionaries as we can see in the unit tests. CC @darrelmiller

I'm not sure what the initial intention was for defining the dictionaries as interface types instead of concrete types.
Maybe @darrelmiller might have more historical context into it?
I'm open to refactoring it to simplify collection initialization syntax if we're all onboard with this change.

darrelmiller:
I suspect the reason we implemented these properties as interfaces was to leave the possibility of changing the concrete implementation if necessary.  I don't have any objection to changing to the concrete classes now there is a clear benefit.

src/Microsoft.OpenApi/Models/OpenApiOperation.cs
```diff
@@ -73,7 +73,7 @@ public ISet<OpenApiTagReference>? Tags
         /// The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a name and location.
         /// The list can use the Reference Object to link to parameters that are defined at the OpenAPI Object's components/parameters.
         /// </summary>
-        public IList<IOpenApiParameter>? Parameters { get; set; } = [];
+        public IList<IOpenApiParameter>? Parameters { get; set; }
```
baywet, line 76: this and security was probably missed. Any property declared as IList really

src/Microsoft.OpenApi/Models/OpenApiPathItem.cs
```diff
@@ -42,10 +41,8 @@ public class OpenApiPathItem : IOpenApiExtensible, IOpenApiReferenceable, IOpenA
         /// <param name="operation">The operation item.</param>
         public void AddOperation(HttpMethod operationType, OpenApiOperation operation)
         {
-            if (Operations is not null)
-            {
-                Operations[operationType] = operation;
-            }
+            Operations ??= new Dictionary<HttpMethod, OpenApiOperation>();
```
baywet, line 44:
```suggestion
            Operations ??= [];
```

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -173,19 +173,19 @@ public decimal? Minimum
...
         /// <inheritdoc />
-        public IList<IOpenApiSchema>? AnyOf { get; set; } = [];
+        public IList<IOpenApiSchema>? AnyOf { get; set; }
 
         /// <inheritdoc />
         public IOpenApiSchema? Not { get; set; }
 
         /// <inheritdoc />
-        public ISet<string>? Required { get; set; } = new HashSet<string>();
+        public ISet<string>? Required { get; set; }
```
baywet, line 188: ISet should also be declared as Hashsets directly

src/Microsoft.OpenApi/Validations/Rules/OpenApiPathsRules.cs
```diff
@@ -42,15 +42,15 @@ public static class OpenApiPathsRules
             new ValidationRule<OpenApiPaths>(nameof(PathMustBeUnique),
                 (context, item) =>
                 {
-                    var hashSet = new HashSet<string>();
+                    var HashSet = new HashSet<string>();
```
baywet, line 45: I'm not sure the casing change brings any value here and it collides with the call name, revert to avoid confusion?

MaggieKimani1: Yeah my bad this was an issue with using replace all

baywet (dismissed):
Thanks for the great work on this one! Please make sure you squash merge with a conventional commit :)

baywet (approved):

Michael-Wamae (approved):

PR #2285, opened 2025-03-26
Task: Move ACR deployment to deploy stage

Onokaev:
~~Completed fix, but holding until tomorrow to test end to end.
Deployment is blocked because of docker pull rate limits~~

Successful run: https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=184590&view=logs&j=94deceea-7f99-5fa2-dcf4-7ec60dc7cc5a&t=94deceea-7f99-5fa2-dcf4-7ec60dc7cc5a


.azure-pipelines/ci-build.yml
```diff
@@ -318,119 +325,127 @@ extends:
...
-                $version = $null
-                if ($content.Project.PropertyGroup -is [array]) {
-                  Write-Host "PropertyGroup is an array, checking each entry..."
-                  foreach ($pg in $content.Project.PropertyGroup) {
-                    if ($pg.Version) {
-                      $version = $pg.Version.ToString().Trim()
-                      Write-Host "Found version in PropertyGroup array: $version"
-                      break
+      - deployment: deploy_docker_image
+        environment: kiota-github-releases
```
baywet, line 329:
```suggestion
```
We don't need this environment since we're not writing to github releases here

.azure-pipelines/ci-build.yml
```diff
@@ -318,119 +325,127 @@ extends:
...
+        strategy:
+          runOnce:
+            deploy:
+              pool:
+                vmImage: 'ubuntu-latest'
+              steps:
+                - checkout: self
```
baywet, line 336: we might not be allowed to checkout the repository in a deployment stage, have you tested this?

Onokaev: It actually works. Let me remove the checkout step

Onokaev: Also, why is this not allowed?

baywet: deployment phases are not allowed to use checkout steps per security policy. Effectively they don't want scripts that could lead to some kind of injection I suppose.
If we revert to a regular job (not a deployment one) it'd be allowed though.
I have no doubt that it can work, but I'm fairly confident it'll trigger security items down the road.

.azure-pipelines/ci-build.yml
Onokaev: The job fails without an environment variable. I'll create a relevant one on ADO. 
I put it back to test it

.azure-pipelines/ci-build.yml
Onokaev: Got it! Thanks

.azure-pipelines/ci-build.yml
```diff
@@ -198,12 +202,29 @@ extends:
...
             sourceFolder: $(Build.ArtifactStagingDirectory)
             content: '*.nupkg'
-            
+
+        # Copy repository files to be used in the deploy stage
+        - task: CopyFiles@2
+          displayName: 'Copy repository files for deploy stage'
+          inputs:
+            SourceFolder: '$(Build.SourcesDirectory)'
+            Contents: |
```
baywet, line 211: have you considered this option instead of copying the whole repo which is effectively working around the security constraints?
https://stackoverflow.com/questions/66468815/how-to-publish-docker-image-as-an-artifact-in-azure-devops

Issue #2283, opened 2025-03-23
Empty tag causes error generating client


StefanCuypers:
### What are you generating using Kiota, clients or plugins?

API Client/SDK

### In what context or format are you using Kiota?

Nuget tool

### Client library/SDK language

Csharp

### Describe the bug

I'm trying to generate a client for the TrackUnit group API, but it throws an error: "error generating the client: Value cannot be null or empty: referenceId (Parameter 'referenceId')"
After some trial/error I found the cause to be the presence of "tags": [ "" ]
If I remove those it seems to work.

### Expected behavior

Ignore empty tags instead of generating an error.

### How to reproduce

You can run the following:
kiota generate -l CSharp -c BaseTrackUnitGroupsClient -n CCTrackUnitClient -d https://developers.trackunit.com/openapi/62bc9436ca283d0083119e2b -o . --ebc

### Open API description file

https://developers.trackunit.com/openapi/62bc9436ca283d0083119e2b

### Kiota Version

1.24.1+68eefbe046b9a9e4ee8b2029070a1cf097e5a984

### Known Workarounds

Download the openapi and remove the "tags": [ "" ] lines.

### Configuration

- OS: Windows 11, using Visual Studio 2022.
- I do not think it is relevant to this configuration

### Debug output

<details><summary>Click to expand log</summary>
Debug: KiotaBuilder kiota version 1.24.1
Debug: KiotaBuilder cache file C:\Users\stefan\AppData\Local\Temp\kiota\cache\generation\D5041F0FF32584C3284CEC6532AD86B93F3A8B85095C3677915F36655E2FB22C\62bc9436ca283d0083119e2b is up to date and clearCache is False, using it
Information: KiotaBuilder loaded description from remote source
Debug: KiotaBuilder step 1 - reading the stream - took 00:00:00.0113209
Critical: KiotaBuilder error generating the client: Value cannot be null or empty: referenceId (Parameter 'referenceId')
</details>

baywet:
Hi @StefanCuypers 
Thank you for using kiota and for reaching out.

I have transferred your issue to OpenAPI.net, the library kiota uses to parse OpenAPI documents. We're currently incubating the preview of v2 in kiota, and feedback like this is extremely helpful in stabilizing the new version of the library.

For maintainers/contributors, we should probably filter out empty values as we parse the document.

Is this something you'd like to submit a pull request for provided some guidance?

StefanCuypers:
Hi @baywet 
[...]

PR #2286, opened 2025-03-26
Fix issue Empty tag causes error generating Kiota client #2283

StefanCuypers:
When parsing the list of tags, removed tags that are empty so we don't fail on them.

### DIFF
--- a/src/Microsoft.OpenApi/Reader/V2/OpenApiOperationDeserializer.cs
+++ b/src/Microsoft.OpenApi/Reader/V2/OpenApiOperationDeserializer.cs
@@ -19,12 +19,26 @@ namespace Microsoft.OpenApi.Reader.V2
     /// </summary>
     internal static partial class OpenApiV2Deserializer
     {
+        /// <summary>
+        /// Have a default empty tag we can use to filter out empty tags.
+        /// </summary>
+        private static OpenApiTagReference emptyTagReference = new("empty");
         private static readonly FixedFieldMap<OpenApiOperation> _operationFixedFields =
             new()
             {
                 {
-                    "tags", (o, n, doc) => { 
-                        if (n.CreateSimpleList((valueNode, doc) => LoadTagByReference(valueNode.GetScalarValue(), doc), doc) is {Count: > 0} tags)
+                    "tags", (o, n, doc) => {
+                        if (n.CreateSimpleList(
+                            (valueNode, doc) =>
+                            {
+                                var val = valueNode.GetScalarValue();
+                                if (string.IsNullOrEmpty(val))
+                                    return emptyTagReference;   // Avoid exception on empty tag, we'll remove these from the list further on
+                                return LoadTagByReference(val , doc);
+                                },
+                            doc)
+                        // Filter out empty tags instead of excepting on them
+                        .Where(n => !object.ReferenceEquals(emptyTagReference, n)).ToList() is {Count: > 0} tags)
                         {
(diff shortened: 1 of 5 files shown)


StefanCuypers:

MaggieKimani1:
Hi @StefanCuypers, do you mind adding another commit that conforms to the [Conventional commits standard](https://www.conventionalcommits.org/en/v1.0.0/#summary)?
We use the commit messages to trigger releases and update the changelog.

src/Microsoft.OpenApi/Reader/V2/OpenApiOperationDeserializer.cs
```diff
@@ -19,12 +19,26 @@ namespace Microsoft.OpenApi.Reader.V2
     /// </summary>
     internal static partial class OpenApiV2Deserializer
     {
+        /// <summary>
+        /// Have a default empty tag we can use to filter out empty tags.
+        /// </summary>
+        private static OpenApiTagReference emptyTagReference = new("empty");
```
baywet, line 25:
```suggestion
```

src/Microsoft.OpenApi/Reader/V2/OpenApiOperationDeserializer.cs
```diff
@@ -19,12 +19,26 @@ namespace Microsoft.OpenApi.Reader.V2
...
                 {
-                    "tags", (o, n, doc) => { 
-                        if (n.CreateSimpleList((valueNode, doc) => LoadTagByReference(valueNode.GetScalarValue(), doc), doc) is {Count: > 0} tags)
+                    "tags", (o, n, doc) => {
+                        if (n.CreateSimpleList(
+                            (valueNode, doc) =>
+                            {
+                                var val = valueNode.GetScalarValue();
+                                if (string.IsNullOrEmpty(val))
+                                    return emptyTagReference;   // Avoid exception on empty tag, we'll remove these from the list further on
```
baywet, line 36:
```suggestion
                                    return null;   // Avoid exception on empty tag, we'll remove these from the list further on
```

src/Microsoft.OpenApi/Reader/V2/OpenApiOperationDeserializer.cs
```diff
@@ -19,12 +19,26 @@ namespace Microsoft.OpenApi.Reader.V2
...
+                                return LoadTagByReference(val , doc);
+                                },
+                            doc)
+                        // Filter out empty tags instead of excepting on them
+                        .Where(n => !object.ReferenceEquals(emptyTagReference, n)).ToList() is {Count: > 0} tags)
```
baywet, line 41:
```suggestion
                        .OfType<IOpenApiTag>().ToList() is {Count: > 0} tags)
```

baywet (changes requested):
Thanks for the contribution!

Here are a couple of recommendations.

@maggieKimani1 don't worry about the conventional commits just yet, we'll squash merge and amend the commit message as we do so

src/Microsoft.OpenApi/Reader/V2/OpenApiOperationDeserializer.cs
StefanCuypers: Thanks for the suggestions (I got a little confused with nullability and thought null was not possible, but this is better off course).
I' ve created a new commit with the modifications (changing the OfType to OpenApiTagReference since that is used on the HashSet further on, which complained on IOpenApiTag).

baywet (approved):

PR #2304, opened 2025-04-04
ci/performance testing

### LINKED ISSUE #1971: [API Review] Memory performance improvements

baywet:
fixes #1971


MaggieKimani1:
Is the run comparison tool for empty models CI step supposed to be failing?

baywet:
> Is the run comparison tool for empty models CI step supposed to be failing? I see the one for v1 is running successfully.

yes, this is strange, the memory footprint of an OpenAPISchema is not the same on my local machine and on the runners, and I don't know why.

MaggieKimani1 (approved):

performance/Program.cs
```diff
@@ -0,0 +1,17 @@
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Running;
+
+namespace performance
```
DevoUmair: ioj

PR #2305, opened 2025-04-06
fix: set format to binary for file uploads

### LINKED ISSUE #2303: v2 : OpenAPI 2.0 `file` Type Parsing Incorrectly Handled
**Describe the bug**
When parsing OpenAPI 2.0 specifications containing parameters with `type: file`, the library (`2.0.0-preview.16`) converts these to `string` type without setting `format: binary`. In my expectation, when viewing OpenAPI 3.x examples, the old `file` type should be parsed as `string` with `binary` format.

We can take a look at `/uploadImage` service from `petStore` example
there is a file parameter of type `file`
```json
    "/pet/{petId}/uploadImage": {
      "post": {
        "tags": [ "pet" ],
        "summary": "uploads an image",
        "description": "",
        "operationId": "uploadFile",
        "consumes": [ "multipart/form-data" ],
        "produces": [ "application/json" ],
        "parameters": [
          {
            "name": "petId",
            "in": "path",
            "description": "ID of pet to update",
            "required": true,
            "type": "integer",
            "format": "int64"
          },
          {
            "name": "additionalMetadata",
            "in": "formData",
            "description": "Additional data to pass to server",
            "required": false,
            "type": "string"
          },
          {
            "name": "file",
            "in": "formData",
            "description": "file to upload",
            "required": false,
            "type": "file"
          }
        ],
```

But after debugging, i found this parameter is parsed to `string` type with format = `null`.
I didn't find another usable property in schemas to *detect* it's binary file and not a string. 

**OpenApi File To Reproduce**
1. Load the PetStore OpenAPI 2.0 example: https://petstore.swagger.io/v2/swagger.json
2. Examine the `/pet/{petId}/uploadImage` operation
3. Observe the `file` parameter is parsed as: 
   type = "string",
   format = null // -> Expected Format = "binary"

**Expected behavior**
Parameters with `type: file` in OpenAPI 2.0 should be parsed as:
type = "string",
format = "binary"

**Screenshots/Code Snippets**

Image

Image

**Additional context**
* OpenAPI 3.x replaced file `type` with `string + binary` format (spec reference: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#considerations-for-file-uploads)
* Current behavior breaks compatibility for tools that rely on format: binary to detect file upload parameters
* This affects code generation where parameters should be treated as binary data (byte[]/Stream) rather than strings

MaggieKimani1:
Fixes #2303

### DIFF
--- a/src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
+++ b/src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
@@ -75,7 +75,7 @@ namespace Microsoft.OpenApi.Reader.V2
                         {                            
                             var schema = GetOrCreateSchema(o);
                             schema.Type = type.ToJsonSchemaType();
-                            if (type.Equals("file"))
+                            if ("file".Equals(type, StringComparison.OrdinalIgnoreCase))
                             {
                                 schema.Format = "binary";
                             }
--- a/src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
+++ b/src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
@@ -72,8 +72,13 @@ namespace Microsoft.OpenApi.Reader.V2
                     {
                         var type = n.GetScalarValue();
                         if (type != null)
-                        {
-                            GetOrCreateSchema(o).Type = type.ToJsonSchemaType();
+                        {                            
+                            var schema = GetOrCreateSchema(o);
+                            schema.Type = type.ToJsonSchemaType();
+                            if (type.Equals("file"))
+                            {
+                                schema.Format = "binary";
+                            }
                         }
                     }
                 },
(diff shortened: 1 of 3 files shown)


Michael-Wamae (dismissed):

src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
```diff
@@ -72,8 +72,13 @@ internal static partial class OpenApiV2Deserializer
                     {
                         var type = n.GetScalarValue();
                         if (type != null)
-                        {
-                            GetOrCreateSchema(o).Type = type.ToJsonSchemaType();
+                        {                            
+                            var schema = GetOrCreateSchema(o);
+                            schema.Type = type.ToJsonSchemaType();
+                            if (type.Equals("file"))
```
baywet, line 78:
```suggestion
                            if ("file".Equals(type, StringComparison.OrdinalIgnoreCase))
```

baywet (changes requested):

baywet (approved):

PR #2309, opened 2025-04-08
fix: read (Exclusive)Maximum and (Exclusive)Minimum values as strings and write their raw values during serialization

### LINKED ISSUE #1967: [API Review] Decimal vs Double for Max/Min in OpenAPISchema type
In the API review, the question of why the (Exclusive)Max/(Exclusive)Min properties was chosen to be of type Decimal vs Double. 
@darrelmiller has a contact to follow up to get an answer and this issue is used as a tracker for this.

MaggieKimani1:
Fixes #1967


src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
```diff
@@ -52,13 +52,13 @@ internal static partial class OpenApiV3Deserializer
...
             {
-                "minimum",
+               "minimum",
                 (o, n, _) =>
                 {
                     var min = n.GetScalarValue();
-                    if (min != null)
+                    if (!string.IsNullOrEmpty(min))
                     {
-                        o.Minimum = ParserHelper.ParseDecimalWithFallbackOnOverflow(min, decimal.MinValue);
```
baywet, line 61: out of curiosity, do we still need the helper method? can we remove it?

MaggieKimani1: No we don't need it anymore. Resolving...

baywet (approved):

baywet:
@maggieKimani1 I usually base that decision on: do I have conventional commits already? how many commits are in this PR? 5 or above... squash

PR #2322, opened 2025-04-12
Use Convert.ToHexString

martincostello:
Use `Convert.ToHexString()` when targeting .NET 8 instead of using a `StringBuilder`.

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
@@ -544,6 +544,9 @@ namespace Microsoft.OpenApi.Models
         {
             // Build the final string by converting each byte
             // into hex and appending it to a StringBuilder
+#if NET
+            return Convert.ToHexString(hash);
+#else
             var sb = new StringBuilder();
             for (var i = 0; i < hash.Length; i++)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -544,13 +544,17 @@ private static string ConvertByteArrayToString(byte[] hash)
         {
             // Build the final string by converting each byte
             // into hex and appending it to a StringBuilder
+#if NET
```
baywet, line 547:
```suggestion
#if NET5_0_OR_GREATER
```
Since this is only available starting in net5. https://learn.microsoft.com/en-us/dotnet/api/system.convert.tohexstring?view=net-9.0#system-convert-tohexstring(system-byte())
( in case we ever decide to change the targets for whatever reason)

baywet (changes requested):
Thanks for the contribution!

baywet (approved):

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
martincostello: My understanding is that `NET` means exactly the same thing - .NET Core 3.1 and lower are `NETCORE` (not `NET`) and there's no such thing as .NET (Core) 4, so the two directives are identical, except for their length in characters.

PR #2323, opened 2025-04-13
fix: handle deserializing and writing empty security requirements #1426

### LINKED ISSUE #1426: No difference between no security requirement and empty security requirement on Operations
**Describe the bug**
There is no way of telling 'empty' from 'not defined' security requirement objects, during deserialization.
I.e no way of distinguishing the 2 different security requirement objects
```yaml
paths:
  /pets:
    post:
      responses:
        '201':
          description: Created
      security: []
    get:
      responses:
        '200':
          description: A paged array of pets
```

Why is this important?

The specification states the following for Operation Security
> To remove a top-level security declaration, an empty array can be used.

i.e `security: []`

however it is expected that if security is not defined on an operation level
```yaml
get:
  responses:
    '200':
      description: A paged array of pets
```
That the global security is applied, defined on the top level document object.

in the following spec
``` yaml 
openapi: '3.0.0'
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
security:
  - Authorization:
    - readwrite
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    post:
      responses:
        '201':
          description: Created
      security: []
    get:
      responses:
        '200':
          description: A paged array of pets
components:
  securitySchemes:
    Authorization:
      type: http
      scheme: bearer
      bearerFormat: JWT
```
post deserialization, it is impossible to determine which security is actually applied between the 2 operations, as both security requirements are deserialized into the same value
`public IList<OpenApiSecurityRequirement> Security { get; set; } = new List<OpenApiSecurityRequirement>();`
which ends up being empty in both cases. 

The reason for this is most likely that during reading a new `OpenApiSecurityRequirement()` and as this is a dictionary, there is no way of specifying a null key, so to speak.

The problem being that the semantics between the 2 operations are different in the spec.

**To Reproduce**
Deserialize the above spec.

**Expected behavior**
To have a way of differentiating the 2 scenarios. (feasibly one would be empty, the other would be null.

**Screenshots/Code Snippets**
```csharp
var input = @"openapi: '3.0.0'
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
security:
  - Authorization:
    - readwrite
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    post:
      responses:
        '201':
          description: Created
      security: []
    get:
      responses:
        '200':
          description: A paged array of pets
components:
  securitySchemes:
    Authorization:
      type: http
      scheme: bearer
      bearerFormat: JWT";
var document = new OpenApiStringReader().Read(input, out var diagnostics);
```

paulmendix:
fix #1426.

make distinction between empty security requirements and no security requirements on an operation. empty security requirements are read as an empty list, no security requirements are read as null for OpenAPI v2/v3/v3.1. This is a breaking change, previously both cases were read as an empty list.

also includes a change to `OpenApiOperation.SerializeInternal` so it can serialize these two cases separately. this required a new method `OpenApiWriterExtensions.WriteOptionalOrEmptyCollection`.

includes unit tests, change to PublicApi.approved.txt to include the new method, and I removed a couple of unused usings and a typo in test name `SerializeDocWithSecuritySchemeWithInlineReferencesWorks`.


paulmendix:
I will contact the right people at my work so I can sign the CLA, this will take some time.

src/Microsoft.OpenApi/Reader/V31/OpenApiOperationDeserializer.cs
```diff
@@ -96,8 +97,11 @@ internal static partial class OpenApiV31Deserializer
                 },
                 {
                     "security", (o, n, t) =>
-                    {
-                        o.Security = n.CreateList(LoadSecurityRequirement, t);
+                    { 
+                        if (n.JsonNode is JsonArray)
+                        {
+                            o.Security = new List<OpenApiSecurityRequirement>(n.CreateList(LoadSecurityRequirement, t));
```
baywet, line 103:
```suggestion
                            o.Security = n.CreateList(LoadSecurityRequirement, t); 
```

test/Microsoft.OpenApi.Tests/Models/OpenApiDocumentTests.cs
```diff
@@ -2220,5 +2218,70 @@ public async Task SerializeDocWithSecuritySchemeWithInlineRefererencesWorks()
...
+                                   '200':
+                                     description: A list of repairs
+                                     content:
+                                       application/json:
+                                         schema:
+                                           type: object
+                           """;
+
+            var doc = (await OpenApiDocument.LoadAsync("Models/Samples/docWithoutOperationSecurity.yaml", SettingsFixture.ReaderSettings)).Document;
+            var stringWriter = new StringWriter();
```
baywet, line 2248: add an assert the security is null

test/Microsoft.OpenApi.Tests/Models/OpenApiDocumentTests.cs
```diff
@@ -2220,5 +2218,70 @@ public async Task SerializeDocWithSecuritySchemeWithInlineRefererencesWorks()
...
+                                     description: A list of repairs
+                                     content:
+                                       application/json:
+                                         schema:
+                                           type: object
+                                 security: [ ]
+                           """;
+
+            var doc = (await OpenApiDocument.LoadAsync("Models/Samples/docWithEmptyOperationSecurity.yaml", SettingsFixture.ReaderSettings)).Document;
+            var stringWriter = new StringWriter();
```
baywet, line 2281: add an assert the security is empty

baywet (changes requested):
Thanks for the contribution!
A couple of suggestions while we wait for the CLA.

test/Microsoft.OpenApi.Tests/Models/OpenApiDocumentTests.cs
paulmendix: Addressed

test/Microsoft.OpenApi.Tests/Models/OpenApiDocumentTests.cs
paulmendix: Addressed

src/Microsoft.OpenApi/Reader/V2/OpenApiOperationDeserializer.cs
```diff
@@ -96,7 +96,10 @@ internal static partial class OpenApiV2Deserializer
                 },
                 {
                     "security",
-                    (o, n, t) => o.Security = n.CreateList(LoadSecurityRequirement, t)
+                    (o, n, t) => { if (n.JsonNode is JsonArray)
+                    {
+                        o.Security = new List<OpenApiSecurityRequirement>(n.CreateList(LoadSecurityRequirement, t));
```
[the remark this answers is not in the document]
paulmendix: Thanks for your feedback @baywet !

I removed the unneeded extra List constructor calls. 

You also mention:

> Not sure why we had differences with v3. I my views:
> * if the property is null or absent from the description, should be null
> * if the property is present but empty array, should be empty
> * if the property is present with items, should be a list with items

as far as I can see that's implemented with this PR, but maybe you are referring to something different. Can you expand on this?

test/Microsoft.OpenApi.Tests/Models/Samples/docWithEmptyOperationSecurity.yaml
```diff
@@ -0,0 +1,20 @@
+openapi: 3.0.0
```
paulmendix, line 1: Another question for you @baywet 

I introduced two new contracts very similar to file [docWithSecurityScheme.yaml](https://github.com/microsoft/OpenAPI.NET/blob/main/test/Microsoft.OpenApi.Tests/Models/Samples/docWithSecurityScheme.yaml) to highlight the difference between security/no security/empty security. At the same time this introduces duplication into the contracts and the unit tests.

Are you okay with this duplication, or shall I remove that by taking out duplicate content from the contracts?

baywet (dismissed):
Thanks for the contribution!

@maggieKimani1 for final review and merge

src/Microsoft.OpenApi/Reader/V2/OpenApiOperationDeserializer.cs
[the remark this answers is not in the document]
baywet: this looks good to me now, thanks!

test/Microsoft.OpenApi.Tests/Models/Samples/docWithEmptyOperationSecurity.yaml
baywet: the duplication is fine, thanks for checking

paulmendix:
Thanks. I addressed the conflict. Ready for final review @MaggieKimani1

baywet (approved):

MaggieKimani1 (approved):

PR #2324, opened 2025-04-14
Use SHA512.HashDataAsync

martincostello:
Use `SHA512.HashDataAsync()` for .NET 8 to reduce allocations.

Complements the change in #2322.

I manually verified the hash before and after was the same by getting the hash from the `TestHashCodesForSimilarOpenApiDocuments` test using the debugger and then verifying it was the same after the change with:

```csharp
Assert.Equal("00BCB7FE78379A07956E06BF3180A8818F6E42E10BC38A7012A342CD6E1033BF03CB7E70E04B285C67BF4D78828D185C1FE438073C2474E40315F2ECE6310B78", doc1HashCode);
```

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
@@ -524,9 +524,7 @@ namespace Microsoft.OpenApi.Models
         /// <returns>The hash value.</returns>
         public async Task<string> GetHashCodeAsync(CancellationToken cancellationToken = default)
         {
-            byte[]? hash;
-
-#if NET
+#if NET7_OR_GREATER
             using var memoryStream = new MemoryStream();
             using var streamWriter = new StreamWriter(memoryStream);
 
@@ -534,7 +532,7 @@ namespace Microsoft.OpenApi.Models
 
             memoryStream.Seek(0, SeekOrigin.Begin);
 
-            hash = await SHA512.HashDataAsync(memoryStream, cancellationToken).ConfigureAwait(false);
+            var hash = await SHA512.HashDataAsync(memoryStream, cancellationToken).ConfigureAwait(false);
 #else
             using HashAlgorithm sha = SHA512.Create();
             using var cryptoStream = new CryptoStream(Stream.Null, sha, CryptoStreamMode.Write);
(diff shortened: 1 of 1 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -524,20 +524,37 @@ public void SetReferenceHostDocument()
         /// <returns>The hash value.</returns>
         public async Task<string> GetHashCodeAsync(CancellationToken cancellationToken = default)
         {
+            byte[]? hash;
+
+#if NET
```
baywet, line 529:
```suggestion
#if NET7_OR_GREATER
```
https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha512.hashdataasync?view=net-9.0

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -524,20 +524,37 @@ public void SetReferenceHostDocument()
...
+            using var memoryStream = new MemoryStream();
+            using var streamWriter = new StreamWriter(memoryStream);
+
+            await WriteDocumentAsync(streamWriter, cancellationToken).ConfigureAwait(false);
+
+            memoryStream.Seek(0, SeekOrigin.Begin);
+
+            hash = await SHA512.HashDataAsync(memoryStream, cancellationToken).ConfigureAwait(false);
```
baywet, line 537:
```suggestion
            var hash = await SHA512.HashDataAsync(memoryStream, cancellationToken).ConfigureAwait(false);
```

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -524,20 +524,37 @@ public void SetReferenceHostDocument()
...
-            SerializeAsV3(openApiJsonWriter);
-            await openApiJsonWriter.FlushAsync(cancellationToken).ConfigureAwait(false);
+            await WriteDocumentAsync(streamWriter, cancellationToken).ConfigureAwait(false);
 
-#if NET5_0_OR_GREATER
-            await cryptoStream.FlushFinalBlockAsync(cancellationToken).ConfigureAwait(false);
-#else
             cryptoStream.FlushFinalBlock();
+
+            hash = sha.Hash;
```
baywet, line 547:
```suggestion
            var hash = sha.Hash;
```

baywet (changes requested):
Thanks for the contribution!

baywet (approved):

MaggieKimani1 (approved):

PR #2325, opened 2025-04-15
fix: Improve handling of OpenAPI tag references

### LINKED ISSUE #2319: Only one tag is returned on operations when OpenApiTagReference is used
**Describe the bug**

Having updated Swashbuckle.AspNetCore to 2.0.0-preview.11 and ASP.NET Core 10 preview 3 in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3283, I've identified an issue where only one tag is returned when a larger set of tag referencess is assigned to an OpenAPI operation.

In creating a self-contained repro for the issue, I find that this is also an issue in 2.0.0-preview.16.

My guess is that there's an issue with how the custom `OpenApiTagComparer` class is implemented, which is used when assigning a value to the property here:

https://github.com/microsoft/OpenAPI.NET/blob/63a8a3480b20809367c2dc7616e23be87a51f8fe/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L29-L45

Specifically, it looks like it doesn't cater for an `OpenApiReference` where `Name` resolves to `null`:

https://github.com/microsoft/OpenAPI.NET/blob/63a8a3480b20809367c2dc7616e23be87a51f8fe/src/Microsoft.OpenApi/OpenApiTagComparer.cs#L20-L35

**Code To Reproduce**

```csharp
using Microsoft.OpenApi.Models.References;

var document = new OpenApiDocument();

var tags = new HashSet<OpenApiTagReference>();

tags.Add(new OpenApiTagReference("one", document));
tags.Add(new OpenApiTagReference("two", document));
tags.Add(new OpenApiTagReference("three", document));

Console.WriteLine("There are {0} original tags.", tags.Count);
Console.WriteLine("The original tags are: {0}", string.Join(", ", tags.Select(t => t.Reference.Id)));

var operation = new OpenApiOperation() { Tags = tags };

Console.WriteLine("There are {0} tags in the operation.", operation.Tags.Count);
Console.WriteLine("The operation tags are: {0}", string.Join(", ", operation.Tags.Select(t => t.Reference.Id)));
```

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net9.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.OpenApi" Version="2.0.0-preview.16" />
  </ItemGroup>
</Project>
```

**Expected behavior**

```text
There are 3 original tags.
The original tags are: one, two, three
There are 3 tags in the operation.
The operation tags are: one, two, three
```

**Actual behaviour**

```text
There are 3 original tags.
The original tags are: one, two, three
There are 1 tags in the operation.
The operation tags are: one
```

martincostello:
- Update `OpenApiTagComparer` to behave intuitively with `OpenApiTagReference` instances pointing to tags not defined in an `OpenApiDocument`.
- Verify OpenAPI tag references refer to a valid OpenAPI tag in the document on serialization.

Fixes #2319.

### DIFF
--- a/test/Microsoft.OpenApi.Tests/Models/OpenApiOperationTests.cs
+++ b/test/Microsoft.OpenApi.Tests/Models/OpenApiOperationTests.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Text.Json.Nodes;
 using System.Threading.Tasks;
+using Microsoft.OpenApi.Exceptions;
 using Microsoft.OpenApi.Extensions;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Models.Interfaces;
@@ -858,6 +859,35 @@ namespace Microsoft.OpenApi.Tests.Models
             baseOperation.Metadata["key1"] = "value2";
 
             Assert.NotEqual(baseOperation.Metadata["key1"], actualOperation.Metadata["key1"]);
+        }
+
+        [Theory]
+        [InlineData(OpenApiSpecVersion.OpenApi2_0)]
+        [InlineData(OpenApiSpecVersion.OpenApi3_0)]
+        [InlineData(OpenApiSpecVersion.OpenApi3_1)]
+        public async Task SerializeAsJsonAsyncThrowsIfTagReferenceIsUnresolved(OpenApiSpecVersion version)
+        {
+            var document = new OpenApiDocument()
+            {
+                Tags =
+                [
+                    new() { Name = "one" },
+                    new() { Name = "three" }
+                ]
+            };
+
+            var operation = new OpenApiOperation()
+            {
+                Tags =
+                [
+                    new OpenApiTagReference("one", document),
+                    new OpenApiTagReference("two", document),
+                    new OpenApiTagReference("three", document)
+                ]
+            };
+
+            var exception = await Assert.ThrowsAsync<OpenApiException>(() => operation.SerializeAsJsonAsync(version));
+            Assert.Equal("The OpenAPI tag reference 'two' does reference a valid tag.", exception.Message);
         }
     }
 }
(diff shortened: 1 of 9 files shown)


baywet (approved):
Thanks for the contribution!

test/Microsoft.OpenApi.Tests/Models/OpenApiOperationTests.cs
```diff
@@ -859,5 +860,34 @@ public void OpenApiOperationCopyConstructorWithAnnotationsSucceeds()
...
+                Tags =
+                [
+                    new OpenApiTagReference("one", document),
+                    new OpenApiTagReference("two", document),
+                    new OpenApiTagReference("three", document)
+                ]
+            };
+
+            var exception = await Assert.ThrowsAsync<OpenApiException>(() => operation.SerializeAsJsonAsync(version));
+            Assert.Equal("The OpenAPI tag reference 'two' does reference a valid tag.", exception.Message);
```
bkoelman, line 890: @martincostello Looks like this needs to be "does not".

martincostello: Oops, you're right. Do you want to do a PR to fix it as you found it?

bkoelman: Not really :)

martincostello: Ok I'll fix it shortly then. Should take less than 5 minutes.

martincostello: #2345

PR #2326, opened 2025-04-15
feat: openapiformat enum cleanup

### LINKED ISSUE #2306: v2 - OpenApiFormat enum could be cleaned up before release
Having an enum for serialization operations but string for deserialization is inconsistent.
I suggest we move the enum to hidi, or remove it entirely, and clean its usage from the serialization methods.

Michael-Wamae:
- openapiformat enum cleanup

fixes #2306


src/Microsoft.OpenApi.Hidi/Options/CommandOptions.cs
```diff
@@ -16,7 +16,7 @@ internal class CommandOptions
         public readonly Option<bool> CleanOutputOption = new("--clean-output", "Overwrite an existing file");
         public readonly Option<string> VersionOption = new("--version", "OpenAPI specification version");
         public readonly Option<string> MetadataVersionOption = new("--metadata-version", "Graph metadata version to use.");
-        public readonly Option<OpenApiFormat?> FormatOption = new("--format", "File format");
+        public readonly Option<string?> FormatOption = new("--format", "File format");
```
baywet, line 19:
```suggestion
        public readonly Option<string> FormatOption = new("--format", "File format");
```

(to align with the other string options)

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -92,8 +93,8 @@ public static Task SerializeAsync<T>(
 
             IOpenApiWriter writer = format switch
```
baywet, line 94: should be lowered here since it's the deeper most call.

src/Microsoft.OpenApi.Hidi/OpenApiService.cs
```diff
@@ -205,8 +205,8 @@ private static async Task WriteOpenApiAsync(HidiOptions options, OpenApiFormat o
 
                 IOpenApiWriter writer = openApiFormat switch
```
baywet, line 206: should be lowered here as well

baywet (changes requested):
Thanks for starting this.
In addition to the inline comments, can you also add an entry to the upgrade guide please?

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
Michael-Wamae: Not clear on this request

baywet: add to lower invariant to the format befor the switch

baywet (approved):

PR #2327, opened 2025-04-15
chore: inspect for mutations in serialization methods

### LINKED ISSUE #2080: v2 - inspect for mutations in serialization methods
in #2079 I've fixed a but where the serialization code was mutating the object. 
This is an issue since if you serialize the same object multiple times, you might get different results. (or depending on the ordering v2/v3)

We should inspect all serialization methods to ensure there are no more mutations of the properties, sub-objects in the serialization code.

MaggieKimani1:
Adds a unit test to inspect whether we mutate any public properties in the DOM by:
- Adding a complete OpenApi document - with 3.1 features and loading it into a DOM
- Serializing the DOM using a standard serializer - System.Text.JSON
- Serializing the same DOM using our current serialization infrastructure
- Serializing the DOM again using STJ
- Doing a deep comparison for the two STJ-serialized versions. If they're different, our custom serialization likely mutated the public DOM.

Fixes #2080


baywet (approved):
Great work!

PR #2352, opened 2025-05-14
chore: implement test cases for types of References in an OpenApi document

### LINKED ISSUE #2062: v2: Handle relative references in OpenApiSchemaReference type

MaggieKimani1:
- Adds test cases for types of references contained in an OpenApi document as covered in this [spec](https://github.com/MicrosoftDocs/openapi-docs/blob/openapi-references/OpenAPI/OpenAPI.NET/references-openapi.md)

Fixes #2062


test/Microsoft.OpenApi.Readers.Tests/V31Tests/ReferenceSamples/rootComponentSchemaReference.yaml
```diff
@@ -0,0 +1,24 @@
...
+              content:
+                application/json:
+                  schema:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/specialitem'
+components:
+    schemas:
+      specialitem:  # Use the item type but provide a different title for the type
+        title: Special Item
```
baywet, line 20: sanity check here, I thought it was only description and summary that could be overridden?

test/Microsoft.OpenApi.Tests/PublicApi/PublicApi.approved.txt
```diff
@@ -1650,6 +1650,7 @@ namespace Microsoft.OpenApi.Services
         public System.Uri? GetDocumentId(string? key) { }
         public bool RegisterComponentForDocument<T>(Microsoft.OpenApi.Models.OpenApiDocument openApiDocument, T componentToRegister, string id) { }
         public void RegisterComponents(Microsoft.OpenApi.Models.OpenApiDocument document) { }
+        public Microsoft.OpenApi.Models.Interfaces.IOpenApiSchema? ResolveJsonSchemaReference(string location) { }
```
baywet, line 1653: could this be internal instead?

test/Microsoft.OpenApi.Readers.Tests/V31Tests/ReferenceSamples/localReferenceToJsonSchemaResource.yaml
```diff
@@ -0,0 +1,23 @@
...
+            - object
+            - 'null'
+          properties:
+            c:
+              type:
+                - object
+                - 'null'
+              properties:
+                b:
+                 $ref: '#/properties/b'
```
baywet, line 23: This one looks invalid to me.
If I go back up one "schema level", I'm within the b schema definition, (L12), and there's no b property there. Am I reading it wrong?

src/Microsoft.OpenApi/Models/OpenApiReference.cs
```diff
@@ -100,7 +106,14 @@ public string? ReferenceV3
                     return Id;
                 }
 
-                return "#/components/" + Type.GetDisplayName() + "/" + Id;
+                return _referenceV3 = "#/components/" + Type.GetDisplayName() + "/" + Id;
```
baywet, line 109: nit: string interpolation

src/Microsoft.OpenApi/Models/OpenApiReference.cs
```diff
@@ -299,5 +312,15 @@ internal void SetSummaryAndDescriptionFromMapNode(MapNode mapNode)
                 Summary = summary;
             }
         }
+
+        internal void SetJsonPointerPath(string pointer)
+        {
+            // Eg of an internal subcomponent's JSONPath: #/components/schemas/person/properties/address
+            if ((pointer.Contains("#") || pointer.StartsWith("http", StringComparison.OrdinalIgnoreCase))
```
baywet, line 319: nit: use the char overload instead

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
```diff
@@ -330,6 +331,90 @@ public bool Contains(string location)
...
+            var currentSegment = pathSegments[0];
+            pathSegments = [.. pathSegments.Skip(1)]; // skip one segment for the next recursive call
+
+            switch (currentSegment)
+            {
+                case OpenApiConstants.Properties:
+                    var propName = pathSegments[0];
+                    if (schema.Properties != null && schema.Properties.TryGetValue(propName, out var propSchema))
+                        return ResolveSubSchema(propSchema, [.. pathSegments.Skip(1)]);
+                    break;
```
baywet, line 390: no need for a break after a return

baywet (commented):
Here are a couple of comments to help move things along.
What I've done:
- high level review of the code.
- review of the samples that are present.

(before reading the spec to come with the freshest eyes possible into that)

What I have not done:
- check against the spec for missing samples.
- deep review of the code.

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -587,17 +587,59 @@ private static string ConvertByteArrayToString(byte[] hash)
...
+                }
+
                 Uri? externalResourceUri = useExternal ? Workspace?.GetDocumentId(reference.ExternalResource) : null;
 
                 uriLocation = useExternal && externalResourceUri is not null
                     ? externalResourceUri.AbsoluteUri + relativePath
                     : BaseUri + relativePath;
             }
 
+            if (reference.Type is ReferenceType.Schema && !uriLocation.StartsWith("http", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
MaggieKimani1: No, actually I want this to work for both local and external references using locators, not identifiers($id).
So I'll update it to check for refs containing `#`

test/Microsoft.OpenApi.Readers.Tests/V31Tests/ReferenceSamples/rootComponentSchemaReference.yaml
MaggieKimani1: Ah yes, getting rid of the comment as the title is not meant to be overridden

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
MaggieKimani1: The break is necessary here to provide a clean exit during execution in case the condition in the if statement doesn't pass

baywet: you can treat that comment as nitpicking. when usually provides better readability IMHO

test/Microsoft.OpenApi.Readers.Tests/V31Tests/ReferenceSamples/localReferenceToJsonSchemaResource.yaml
MaggieKimani1: I'm getting it the same way, but this is the same example given in the [spec](https://github.com/MicrosoftDocs/openapi-docs/blob/c2b8bfcee7340f62efbe9bb87b9b68c3d0278afb/OpenAPI/OpenAPI.NET/references-openapi.md?plain=1#L270-L290) and [Safia's example](https://github.com/microsoft/OpenAPI.NET/issues/2062). 
So in your opinion the JSON pointer should be `#/a/properties/b` since a has a b property defined within it?

Maybe @darrelmiller can offer more insight?

darrelmiller: @MaggieKimani1 @baywet You don't "go back up one schema level".  You go back up until you reach a "JSON Schema Resource".  This is either a schema with a $Id, the root of the document, or in the case of OpenAPI to the component schema.  In this case "a" is the JSON Schema Resource which is effectively the root document as far as JSON Schema is concerned. The label "a" is outside of the JSON Schema Resource which is why it is not part of the $ref.

baywet: Thanks for clarifying. I made a similar comment on the spec side PR. I think that "routine" should be explicitly called out there as it's not "broadly known"

src/Microsoft.OpenApi/Models/OpenApiReference.cs
```diff
@@ -73,16 +74,22 @@ public class OpenApiReference : IOpenApiSerializable, IOpenApiDescribedElement,
...
             {
+                if (!string.IsNullOrEmpty(_referenceV3))
+                {
+                    return _referenceV3;
+                }
+
                 if (IsExternal)
                 {
-                    return GetExternalReferenceV3();
+                    return _referenceV3 = GetExternalReferenceV3();
```
baywet, line 92: why are we mutating the object in a getter? x2

src/Microsoft.OpenApi/Models/OpenApiReference.cs
```diff
@@ -299,5 +313,43 @@ internal void SetSummaryAndDescriptionFromMapNode(MapNode mapNode)
...
+        internal void SetJsonPointerPath(string pointer, string nodeLocation)
+        {
+            // Relative reference to internal JSON schema node/resource (e.g. "#/properties/b")
+            if (pointer.StartsWith("#/", StringComparison.OrdinalIgnoreCase) && !pointer.Contains("/components/schemas"))
+            {
+                ReferenceV3 = ResolveRelativePointer(nodeLocation, pointer);
+            }
+
+            // Absolute reference or anchor (e.g. "#/components/schemas/..." or full URL)
+            else if ((pointer.Contains('#') || pointer.StartsWith("http", StringComparison.OrdinalIgnoreCase))
```
baywet, line 326: should this also support local storage paths? (either relative or absolute)
what about relative http paths?

test/Microsoft.OpenApi.Readers.Tests/V31Tests/ReferenceSamples/rootComponentSchemaReference.yaml
baywet: should the title even be present then? what is it meant to validate?

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
```diff
@@ -330,6 +331,90 @@ public bool Contains(string location)
...
+            if (pathSegments.Length == 0)
+            {
+                return schema;
+            }
+            var currentSegment = pathSegments[0];
+            pathSegments = [.. pathSegments.Skip(1)]; // skip one segment for the next recursive call
+
+            switch (currentSegment)
+            {
+                case OpenApiConstants.Properties:
```
baywet, line 386: if we have a property name "properties" or "allOf" etc... isn't this going to throw all the logic off balance?
what do you think about having a third optional boolean parameter to the method "isKnownElement" or something like that that can be used to drive the lookup and address this issue?

src/Microsoft.OpenApi/Models/OpenApiReference.cs
MaggieKimani1: This hasn't been scoped out in the spec at the moment. I think once included, it can be implemented post GA

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
MaggieKimani1: Yes, I think that will help disambiguate between known keywords and user-defined properties

src/Microsoft.OpenApi/Models/OpenApiReference.cs
baywet: CC @darrelmiller in case this was an oversight and needs to be completed by GA

PR #2358, opened 2025-05-17
Nullable fixes

### LINKED ISSUE #2342: Redundant `type: object` emitted in 2.x preview
**Describe the bug**
Trying to use 2.0 preview 17, I see the following change in the output document, compared to 1.x:

Image

It seems that whenever `OpenApiSchema.Type` is set to `JsonSchemaType.Null` from code, this results in also emitting `"type": "object"`. This is redundant in many cases, for example when `allOf` is used.

While probably technically still correct, this difference makes it a pain to assess whether the new version produces a similar output document in our test suite. We have about 12,000 tests, many of which compare fragments of the OAS file, failing on changes like this. The fact that `"nullable": true` moved to the top doesn't help either, but I suspect that will be addressed in a future preview, tracked at https://github.com/microsoft/OpenAPI.NET/issues/1314.

So I wonder if this change was intentional. And if not, how hard it would be to not emit `"type": "object"` when the schema contains `allOf`, `oneOf`, etc.

### LINKED ISSUE #2344: Position of `nullable` property changed in output document
Please restore the position of the `nullable` property in the output document. Comparing 2.0 preview with 1.x stable, it has moved:

Image

While technically the position shouldn't matter, this difference makes it a pain to assess whether the preview version produces a similar output document in our test suite. We have about 12,000 tests, many of which compare fragments of the OAS file, failing on changes like this.

bkoelman:
This PR updates the deserializers to translate incoming `nullable` and `x-nullable` properties to the `OpenApiSchema.Type` property. And then translate back into separate properties in the V2/V3 serializers, preserving the same ordering as in the current 1.x versions.

I chose this approach to avoid having multiple versions of the truth internally. For example, a V2 document may be loaded that contains `x-nullable: true`, then the `OpenApiSchema.Type` is modified from user code, and the file is written back. It would be unexpected to have the presence of `x-nullable: true` in the original document override the `OpenApiSchema.Type` change from code. I'm assuming that consumers upgrading to v2.x are going to manipulate the `OpenApiSchema.Type` property to add/remove nullability, and then write out the version as configured (which is an implementation detail).

But for backwards compatibility, the presence of `x-nullable` in the `OpenApiSchema` is still taken into account, in case it gets explicitly added by user code.

The only downside is that we lose the relative position of `x-nullable: true` if multiple extensions are used. I believe this to be an acceptable tradeoff.

Furthermore, this PR simplifies and optimizes the logic to write the schema type. When writing V2/V3, multiple non-null types still result in not writing the type constraint. Writing nullability in V2/V3 is handled separately from writing the schema type.

Tests are added to verify `nullable`/`x-nullable` with/without a schema type round-trip in V2/V3.

Fixes #2342, fixes #2344.


bkoelman:

bkoelman:
Tagging @MaggieKimani1 and @baywet for review, based on commit history.

baywet (approved):
This is great work and that cleans up the code quite a bit! Thanks for the contribution!

PR #2359, opened 2025-05-19
fix: refactor to avoid adding duplicate entries

### LINKED ISSUE #2350: AddComponent method throws Argument exception for duplicate key
The `AddComponent` method throws an exception if a component with the specified name already exists.
Is this the desired behavior? If so, should it be documented as one of the exceptions the method can throw?
```
    /home/vscode/.nuget/packages/microsoft.extensions.apidescription.server/10.0.0-preview.4.25229.110/build/Microsoft.Extensions.ApiDescription.Server.targets(67,5): error :  ---> System.ArgumentException: An item with the same key has already been added. Key: ProblemDetails
```

MaggieKimani1:
Fixes #2350

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiDocument.cs
@@ -665,47 +665,52 @@ namespace Microsoft.OpenApi.Models
             Utils.CheckArgumentNull(componentToRegister);
             Utils.CheckArgumentNullOrEmpty(id);
             Components ??= new();
+
+            static Dictionary<string, TValue> AddToDictionary<TValue>(Dictionary<string, TValue>? dict, string key, TValue value)
+            {
+                dict ??= new Dictionary<string, TValue>();
+#if NET5_0_OR_GREATER
+                dict.TryAdd(key, value);
+#else
+                if (!dict.ContainsKey(key))
+                {
+                    dict.Add(key, value);
+                }
+#endif
+                return dict;
+            }
+
             switch (componentToRegister)
             {
                 case IOpenApiSchema openApiSchema:
-                    Components.Schemas ??= [];
-                    Components.Schemas.Add(id, openApiSchema);
+                    Components.Schemas = AddToDictionary(Components.Schemas, id, openApiSchema);
                     break;
                 case IOpenApiParameter openApiParameter:
-                    Components.Parameters ??= [];
-                    Components.Parameters.Add(id, openApiParameter);
+                    Components.Parameters = AddToDictionary(Components.Parameters, id, openApiParameter);
                     break;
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -665,47 +665,52 @@ public bool AddComponent<T>(string id, T componentToRegister)
             Utils.CheckArgumentNull(componentToRegister);
             Utils.CheckArgumentNullOrEmpty(id);
             Components ??= new();
+
+            static Dictionary<string, TValue> AddToDictionary<TValue>(Dictionary<string, TValue>? dict, string key, TValue value)
+            {
+                dict ??= new Dictionary<string, TValue>();
```
baywet, line 671: I'd argue that we should NOT initialize the dictionaries here as it:
- mixes concerns (add vs initialize) (readability)
- makes the method prototype more complex (readability)
- forces us to assign the dictionary value (on the component object) at every call (perf)

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -665,47 +665,52 @@ public bool AddComponent<T>(string id, T componentToRegister)
             Utils.CheckArgumentNull(componentToRegister);
             Utils.CheckArgumentNullOrEmpty(id);
             Components ??= new();
+
+            static Dictionary<string, TValue> AddToDictionary<TValue>(Dictionary<string, TValue>? dict, string key, TValue value)
```
baywet, line 669: I think this method should return a boolean value to indicate whether the component was added or not.

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -665,47 +665,52 @@ public bool AddComponent<T>(string id, T componentToRegister)
...
                     break;
                 case IOpenApiHeader openApiHeader:
-                    Components.Headers ??= [];
-                    Components.Headers.Add(id, openApiHeader);
+                    Components.Headers = AddToDictionary(Components.Headers, id, openApiHeader);
                     break;
                 case IOpenApiSecurityScheme openApiSecurityScheme:
-                    Components.SecuritySchemes ??= [];
-                    Components.SecuritySchemes.Add(id, openApiSecurityScheme);
+                    Components.SecuritySchemes = AddToDictionary(Components.SecuritySchemes, id, openApiSecurityScheme);
```
baywet, line 713: leaving the comment here as the lines below are out of scope and I can't leave a comment there.

Should we attempt to register the component if we have not added it to the target collection? probably not to avoid confusion.

baywet (changes requested):

Issue #1314, opened 2023-08-17
Make it possible to have a deterministic order in serialized OpenApi files


gturri:
**Is your feature request related to a problem? Please describe.**
I use OpenAPI.NET to serialize instances of `OpenApiDocument` and store them in a git repo.
So, ideally I would like that when I make a simple change to my API, I get a simple change in the serialized document, in order to have an easy to review git diff.

It is not the case currently, because serialization iterates on some `Dictionary` and that the order of the iteration is hence not deterministic.
For instance, if I add one endpoint to my API, I would like that my serialized OAS file looks like what it was at my previous generation, with just a few added lines. But in practice I may have more diffs, because my paths and my schemas may end up serialized in a different order.

**Describe the solution you'd like**
I can already achieve part of what I would like from my client code in the sense that I can get a deterministic order for the Schemas section, if I do

```
OpenApiDocument doc = BuildMyDoc();
doc.Components.Schemas = new SortedList<string, OpenApiSchema>(doc.Component.Schemas) // <-- that's the important line
doc.SerializeAsV3(new OpenApiYamlWriter(....));
```

It works rather fine, in the sense that it leads to a serialized `schemas` section where all schemas are order alphabetically (making it nice for my git diffs, and making the file easier to browse overall). However, it does not feel really "clean" in itself, and this approach does not work for other parts of the document. In particular for the paths; because they are represented by an `OpenApiPaths`  which ends up extending a `Dictionary` and I could not find a virtual method or an attribute I could override to achieve a similar deterministic order.

Given this context, a solution that would be nice would be to edit `IOpenApiSerializable` in order to turn `SerializeAsV3(IOpenApiWriter writer)` into `SerializeAsV3(IOpenApiWriter writer, bool enforceAlphabeticalOrder = false)`.
So for instance `OpenApiExtensibleDictionary.SerializeAsV3` could be implemented like

```
var items = enforceAlphabeticalOrder ? new SortedList<string, T>(this) : this;
foreach(var item in items) {
  ...
}
```

Having this feature behind a boolean `enforceAlphabeticalOrder` could make it possible for users interested in this feature to use it, and users not interested in it would not suffer any performance penalty by leaving it to false.

**Describe alternatives you've considered**
As explained above, I tried to achieve this order from my client code, but faced some hard limitations

**Additional context**
If the proposed approach looks good to you, and if it could help, I guess I could try to propose a pull request to implement this.

gturri:
[...]

PR #2363, opened 2025-05-21
feat: Add writer settings to enable sorting using a comparer

### LINKED ISSUE #2329: Change all collections to ordered collections

MaggieKimani1:
- Changes the API surface area back to IDictionary to allow future implementations to use OrderedDictionary without it being a breaking change and continue allowing Swashbuckle customers to provide custom ordering.
- IDictionary will require customers to explicitly provide the type until .NET adds implicit support
- Adds `OpenApiWriterSettings` options to meet the need for deterministic ordering as per [this comment](https://github.com/microsoft/OpenAPI.NET/pull/2347#issuecomment-2876898641):
		1.  `EnableSorting` to enable sorting dictionary keys using a default comparer
		2. `KeyComparer` to allow consumers to plugin a custom comparer to sort dictionary entries
		

Fixes #2329

### DIFF
--- a/src/Microsoft.OpenApi/Extensions/DictionaryExtensions.cs
+++ b/src/Microsoft.OpenApi/Extensions/DictionaryExtensions.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.OpenApi.Extensions
+{
+    /// <summary>
+    /// Dictionary extension methods
+    /// </summary>
+    public static class DictionaryExtensions
+    {
+        /// <summary>
+        /// Returns a new dictionary with entries sorted by key using the default comparer.
+        /// </summary>
+        public static IDictionary<TKey, TValue> Sort<TKey, TValue>(
+            this IDictionary<TKey, TValue> source)
+            where TKey : notnull
+        {
+#if NET7_0_OR_GREATER
+            ArgumentNullException.ThrowIfNull(nameof(source));
+#else
+            if (source == null)
+                throw new ArgumentNullException(nameof(source));
+#endif
+
+            return source.OrderBy(kvp => kvp.Key)
+                               .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+        }
+
+        /// <summary>
+        /// Returns a new dictionary with entries sorted by key using a custom comparer.
+        /// </summary>
+        public static IDictionary<TKey, TValue> Sort<TKey, TValue>(
+            this IDictionary<TKey, TValue> source,
+            IComparer<TKey> comparer)
+            where TKey : notnull
+        {
+#if NET7_0_OR_GREATER
+            ArgumentNullException.ThrowIfNull(nameof(source));
+            ArgumentNullException.ThrowIfNull(nameof(comparer));
+#else
+            if (source == null)
+                throw new ArgumentNullException(nameof(source));
+            if (comparer == null)
+                throw new ArgumentNullException(nameof(comparer));
+#endif
+            return source.OrderBy(kvp => kvp.Key, comparer)
+                               .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+        }
+    }
+}
--- a/src/Microsoft.OpenApi/Writers/OpenApiWriterExtensions.cs
+++ b/src/Microsoft.OpenApi/Writers/OpenApiWriterExtensions.cs
@@ -266,7 +266,7 @@ namespace Microsoft.OpenApi.Writers
         public static void WriteRequiredMap(
             this IOpenApiWriter writer,
             string name,
-            Dictionary<string, string>? elements,
+            IDictionary<string, string>? elements,
             Action<IOpenApiWriter, string> action)
         {
             writer.WriteMapInternal(name, elements, action);
@@ -282,7 +282,7 @@ namespace Microsoft.OpenApi.Writers
         public static void WriteOptionalMap(
             this IOpenApiWriter writer,
             string name,
-            Dictionary<string, JsonNode>? elements,
+            IDictionary<string, JsonNode>? elements,
             Action<IOpenApiWriter, JsonNode> action)
         {
             if (elements != null && elements.Any())
@@ -301,7 +301,7 @@ namespace Microsoft.OpenApi.Writers
         public static void WriteOptionalMap(
             this IOpenApiWriter writer,
             string name,
-            Dictionary<string, string>? elements,
+            IDictionary<string, string>? elements,
             Action<IOpenApiWriter, string> action)
         {
             if (elements != null && elements.Any())
@@ -320,7 +320,7 @@ namespace Microsoft.OpenApi.Writers
         public static void WriteOptionalMap(
             this IOpenApiWriter writer,
             string name,
-            Dictionary<string, bool>? elements,
+            IDictionary<string, bool>? elements,
             Action<IOpenApiWriter, bool> action)
         {
             if (elements != null && elements.Any())
@@ -339,7 +339,7 @@ namespace Microsoft.OpenApi.Writers
         public static void WriteOptionalMap(
             this IOpenApiWriter writer,
             string name,
-            Dictionary<string, HashSet<string>>? elements,
+            IDictionary<string, HashSet<string>>? elements,
             Action<IOpenApiWriter, HashSet<string>> action)
         {
             if (elements != null && elements.Any())
@@ -359,7 +359,7 @@ namespace Microsoft.OpenApi.Writers
         public static void WriteOptionalMap<T>(
             this IOpenApiWriter writer,
             string name,
-            Dictionary<string, T>? elements,
+            IDictionary<string, T>? elements,
             Action<IOpenApiWriter, T> action)
             where T : IOpenApiElement
         {
@@ -380,7 +380,7 @@ namespace Microsoft.OpenApi.Writers
         public static void WriteOptionalMap<T>(
             this IOpenApiWriter writer,
             string name,
-            Dictionary<string, T>? elements,
+            IDictionary<string, T>? elements,
             Action<IOpenApiWriter, string, T> action)
             where T : IOpenApiElement
         {
@@ -401,7 +401,7 @@ namespace Microsoft.OpenApi.Writers
         public static void WriteRequiredMap<T>(
             this IOpenApiWriter writer,
             string name,
-            Dictionary<string, T>? elements,
+            IDictionary<string, T>? elements,
             Action<IOpenApiWriter, T> action)
             where T : IOpenApiElement
         {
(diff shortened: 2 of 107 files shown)


src/Microsoft.OpenApi/Writers/OpenApiWriterSettings.cs
```diff
@@ -24,5 +25,15 @@ internal bool ShouldInlineReference(OpenApiReference reference)
...
+
+        /// <summary>  
+        /// Enables sorting of collections using the default comparer  
+        /// </summary>  
+        public bool EnableSorting { get; set; }
+
+        /// <summary>  
+        /// Custom comparer for sorting collections.  
+        /// </summary>  
+        public IComparer<string>? KeyComparer { get; set; }
```
baywet, line 37: we probably don't need the enable sorting flag, just the presence of a non-null value for the comparer should be enough

src/Microsoft.OpenApi/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+            return source.OrderBy(kvp => kvp.Key)
+                               .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+        }
+
+        /// <summary>
+        /// Returns a new dictionary with entries sorted by key using a custom comparer.
+        /// </summary>
+        public static IDictionary<TKey, TValue> Sort<TKey, TValue>(
+            this IDictionary<TKey, TValue> source,
+            IComparer<TKey> comparer)
```
baywet, line 35: nit: you could make the parameter optional and vary on a condition below to avoid duplication

src/Microsoft.OpenApi/Writers/OpenApiWriterExtensions.cs
```diff
@@ -458,6 +459,17 @@ private static void WriteMapInternal<T>(
 
             if (elements != null)
             {
+                var settings = writer.GetSettings();
+
+                if (settings?.EnableSorting == true)
```
baywet, line 464: probably can be cleaned up if we remove the setting.
I think people should tell us what sorting they want, if not, we simply rely on the collections' ordering (no additional sorting).

src/Microsoft.OpenApi/Writers/OpenApiWriterExtensions.cs
```diff
@@ -458,6 +459,17 @@ private static void WriteMapInternal<T>(
...
+                {
+                    elements = elements.Sort(); // sort with default comparer
+                }
+                else if (settings?.KeyComparer != null)
+                {
+                    elements = elements.Sort(settings.KeyComparer); // sort using custom comparer
```
baywet, line 470: probably needs to be duplicated for lists/arrays as well

MaggieKimani1: This includes HashSets yes?

baywet: I'd say yes according to Mike's comment in the original issue.

bkoelman:
First of all, thanks for reverting back to collection interfaces. This unblocks mine and other projects.

> The primary goal of sorting the collections is to provide a deterministic order.

But the order _is_ already deterministic, right? To the best of my knowledge, it has always been in this library, without the need for sorting support. This is because `List<>`, `HashSet<>` and `Dictionary<,>` preserve insertion order. So reading or writing the same DOM twice results in the exact same ordering. Can you provide a use case where ordering is _not_ deterministic already? The only thing I can think of is parallel processing, but that requires collections to be thread-safe, so I don't think that happens in this library.

> Ordering using the writer settings is opt-in and now that everything is back to interfaces, based on your feedback, developers are free to provide implementations with specific ordering characteristics.

I understand that using the writer settings is opt-in, and I'm happy with that. It's just that I can't think of a scenario where _anyone_ would _ever_ want to use it, given the way it is implemented in this PR: it's an all-or-nothing switch. Once a comparer is provided, then _all_ dictionaries _everywhere_ get sorted, which is what truly _nobody_ wants. I'm just alerting that likely a useless feature is being added, possibly due to miscommunication with others. This won't help ASP.NET or Swashbuckle in any way, because they want to preserve declaration order for C# properties. They _do_ want to sort the list of component schemas, but with the all-or-nothing comparer support provided here, they can't use that.

This is because the methods `WriteCollectionInternal` and `WriteMapInternal` (which do the sorting) are called from _everywhere_, not just for the list of component schemas.

mikekistler:
@bkoelman Here is a concrete example where we need to be able to sort path objects by path:

https://github.com/dotnet/eShop/pull/680

Specifically:

https://github.com/dotnet/eShop/pull/680/files#diff-37fbe077a16c7f3aa77bef886e92d1cab152057067f1e21bf54442489892cd27

In this PR, the Catalog service adds a new API version while the previous version _should_ be unchanged. But the diff shows 602 lines changed.

image

This is almost entirely because of reorderings, e.g. the "/api/catalog/items/by" path was originally ordered _after_ "/api/catalog/items", but the changes cause it to now be ordered _before_ "/api/catalog/items".

ASP.NET does not document the order in which endpoints are added to the document and developers should not depend on this ordering. So to achieve a deterministic ordering we want to emit the paths objects ordered by path.

bkoelman:
@mikekistler Thanks for providing a use case. Again, this is about sorting _paths_ only. Not about sorting _everything_ in the document. For example, the `CatalogBrand` schema has properties `id` and `brand`. It wouldn't be correct if their ordering changed too.

This example essentially reinforces my point that providing a comparer _won't_ solve the problem.

bkoelman:
Oh yeah, I'm well aware of that. In my early days I implemented my own collection types with stronger guarantees. Until after a few years I realized almost every .NET app indirectly depends on this implementation detail. `Dictionary<,>` internally maintains an array exactly for this, because it's so extremely useful, at the cost of requiring extra memory. The whole .NET ecosystem would break if this ever changes, so it's a theoretical concern.

The same applies to the CLR memory model, whose implementation has way stronger guarantees than documented, and systems have taken a dependency on it. A few years back, those docs were updated to match the implementation, after ARM support was added that initially didn't provide the same guarantees, and everything broke.

And let's not forget that various improvements in .NET Core were _reverted_ after a few releases when customers started reporting their .NET Framework code didn't work anymore. For example,  that `List<>` implements `IReadOnlyCollection<>`, which is odd, but it was too late to change.

So I wouldn't worry about this for a second. This library was been deterministic for many years. If that weren't the case, it would have surfaced, because OAS files are typically under source control in consuming systems.

bkoelman:
Assuming that Mike's code relies on aspnet doing the sorting, confirm with @captainsafia, who has already indicated that sorting will not be applied to properties,  but will only be applied to components [here](https://github.com/microsoft/OpenAPI.NET/issues/1314#issuecomment-2836887967):

> This isn't about reorder properties on schemas. It's about ordering in the for top-level items under the components property in the OpenAPI document.

mikekistler:
@DarrelMiller I'm afraid it will be a problem if the properties in schemas are sorted lexicographically.

There may just a few other cases where we don't want lexicographic sorting. Others I can think of:
- enum values
- the tag names in the tags array of an operation
- security schemes in the `security` of an operation -- maybe, unsure on this one.

In short, I think we would like a solution that lets us sort the things we want sorted, but leave alone things we'd prefer to leave in their "natural order".

For example, if we could provide a set of comparers on OpenAPI object types, e.g. `Comparer<PathObject>`, and the writer would sort those particular object types with that comparer if it was provided and otherwise leave the existing order. Could something like that work?

src/Microsoft.OpenApi/Writers/OpenApiWriterSettings.cs
```diff
@@ -21,5 +23,10 @@ internal bool ShouldInlineReference(OpenApiReference reference)
             return (reference.IsLocal && InlineLocalReferences)
                              || (reference.IsExternal && InlineExternalReferences);
         }
+
+        /// <summary>  
+        /// Custom comparer for sorting collections.  
+        /// </summary>  
+        public IComparer<string>? Comparer { get; set; }
```
baywet: I'd suggest to make both the name and the doc comment much more explicit than this as to what it does

src/Microsoft.OpenApi/Extensions/CollectionExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Dictionary extension methods
+    /// </summary>
+    public static class CollectionExtensions
+    {
+        /// <summary>
+        /// Returns a new dictionary with entries sorted by key using a custom comparer.
+        /// </summary>
+        public static IDictionary<TKey, TValue> Sort<TKey, TValue>(
```
baywet: can we make the class/methods internal so people don't take a dependency on it?

baywet:
@mikekistler @bkoelman would that work for you if instead of having a single comparer, we had a bunch of named comparers in the settings? (e.g. PathSortingComparer, ComponentIdsSortingComparer, SchemaPropertiesSortingComparer, etc...) this way everyone gets granular control over what they want to sort? And if new collection arise from the standard, adding new comparers is additive (no breaking changes)

baywet (approved):
LGTM besides the ongoing discussion about the desired behaviour

mikekistler:
> would that work for you if instead of having a single comparer, we had a bunch of named comparers in the settings? (e.g. PathSortingComparer, ComponentIdsSortingComparer, SchemaPropertiesSortingComparer, etc...) this way everyone gets granular control over what they want to sort? And if new collection arise from the standard, adding new comparers is additive (no breaking changes)

I think something like this will work for ASP.NET.  I'm not sure why these have to be named comparers, as I think it could be just a set of comparers that holds generic `Comparer<T>`, but named comparers are fine. If they are going to be named, I think a standard naming convention like "<Classname>Comparer" should be used. I think "SortingComparer" is a bit redundant and "Comparer" is preferable.

bkoelman:
> would that work for you if instead of having a single comparer, we had a bunch of named comparers in the settings?

Unfortunately, it's not that simple. Let's take an enum, which can be defined at various depths in the tree, ie, inside a `not`, `oneOf`, `allOf`, `if`/`then`/`else`, etc. If that enum originates from a user-defined model, it _should not_ be sorted. But other enums that exist only for technical reasons _should_ generally be sorted. For example, an enum for the discriminator in inheritance, because the ordering of CLR types in an assembly is not something OpenAPI consumers care about.

Besides that, only string comparers make sense (used for dictionary keys). A comparer on an `IList<IOpenApiSchema>` is impossible to implement in a generic way. How could someone ever sort a list that consists of different schema types? The only reasonable way to sort such a list is to implement a visitor/walker that recursively inspects subtrees.

I'm happy to see that this PR has been merged, because the interfaces unblock all use cases, as far as I'm aware. I recommend hiding the comparer support for the next .NET 10 preview release. As far as I understand it, ASP.NET won't benefit from their presence, but the good news is that they can now do whatever's needed themselves, since interfaces are back.

We're merely guessing at what the true concerns are for ASP.NET, so I think a conversation with the team should happen first, before releasing something that hasn't been fully thought through.

Issue #2367, opened 2025-05-21
Flatten namespace hierarcy


darrelmiller:
**Is your feature request related to a problem? Please describe.**
There are a set of namespaces that exist in the OpenAPI.Net library that have been created due to the organization of class files by folder.  However, there is no value added by having matching namespaces.  It requires the knowledge and use of "using" to bring in these namespaces.

**Describe the solution you'd like**
Align all most classes with a single top level namespace

Change the following namespaces to Microsoft.OpenApi:

Microsoft.OpenApi.Attributes
Microsoft.OpenApi.Exceptions
Microsoft.OpenApi.Helpers
Microsoft.OpenApi.Interfaces
Microsoft.OpenApi.Models
Microsoft.OpenApi.Models.Interfaces
Microsoft.OpenApi.Models.References
Microsoft.OpenApi.Properties

Change the following namespaces to Microsoft.OpenApi.Reader. It is useful to be in a different namespace because not everyone needs to read documents:

Microsoft.OpenApi.Reader.ParseNodes
Microsoft.OpenApi.Reader.Services

Leave these namespaces as they allow consistent class names across the versions

Microsoft.OpenApi.Reader.V2
Microsoft.OpenApi.Reader.V3
Microsoft.OpenApi.Reader.V31

Microsoft.OpenApi.MicrosoftExtensions

mconnew:
I know this issue is closed and finished with and there's no going back, but I wanted to provide some feedback about how this was a really bad idea and strongly discourage this sort of thing happening again in the future.  

[...]

PR #2368, opened 2025-05-26
chore: flatten namespace hierarchy 

MaggieKimani1:
Flattens namespaces to align with single top-level namespace
Fiixes #2367


darrelmiller:
Sorry, I missed a namespace in the list I created in the issue.  We should also remove Microsoft.OpenApi.Extensions.  This one is especially important because the discovery of extension methods is hard when they are in a different namespace.

MaggieKimani1:
> Sorry, I missed a few namespaces in the list I created in the issue. We should also remove
> 
> Microsoft.OpenApi.Extensions Microsoft.OpenApi.Services Microsoft.OpenApi.Validations Microsoft.OpenApi.Writers
> 
> The Extensions one is especially important because the discovery of extension methods is hard when they are in a different namespace.

What about `Microsoft.OpenApi.Expressions`?

darrelmiller (approved):

PR #2405, opened 2025-06-18
avoid stack overflow on cyclical references

baywet:
- **chore: adds additional multiple levels of references test cases**
- **chore: adds update to parent property**
- **fix: avoid stack overflow on cyclical references**


peombwa (approved):
LGTM

PR #2432, opened 2025-07-10
fix: removes loop methods from parsing context as its available in loop detector instead

baywet:


MaggieKimani1 (approved):

baywet:

PR #2435, opened 2025-07-10
fix: switches to a getter for API consistency

baywet:


baywet:

Issue #2442, opened 2025-07-11
Anti-Update practices


RaHorusFreak:
Hello team,

It's both frustrating and ironic to see how updating to Microsoft.OpenApi 2.0 ends up breaking an entire solution without any notice, warning, change log, or similar documentation. It's as if the update process has become: try it, if it breaks — revert; if it doesn't — you're lucky. These practices are steadily eroding the .NET ecosystem. Please, when similar changes are introduced, at the very least there should be explanations and actionable solutions offered to address such disruptive issues.

baywet:
Hi @RaHorusFreak 
Thank you for using the SDK and for reaching out.
Did you happen to find this page? 
https://github.com/microsoft/openapi.net/blob/main/docs%2Fupgrade-guide-2.md

madhon:
> Hi [@RaHorusFreak](https://github.com/RaHorusFreak) Thank you for using the SDK and for reaching out. Did you happen to find this page? https://github.com/microsoft/openapi.net/blob/main/docs%2Fupgrade-guide-2.md

That page doesn't address the breaking caused in the integration with Microsoft.AspNetCore.OpenApi, eg document transformers  due to the type changes/removals

Frulfump:
Maybe that will be produced as part of this issue #2426. Strange to not have that in place before releasing 2.0.0 GA.

madhon:
> Maybe that will be produced as part of this issue [#2426](https://github.com/microsoft/OpenAPI.NET/issues/2426). Strange to not have that in place before releasin 2.0.0 GA.

That issue only talks about documentation so cant see how as it requires code changes in the aspnetcore side to account for the breaking changes in this repo

Frulfump:
(Not stale). To be fair I expect breaking changes when it's a new major release even tough not everyone follows semver 2. But as pointed out there should be docs to explain how we should absorb them, especially since ASP.NET Core is a large user of this package.

baywet:
Hi everyone, 
Thank you for the additional information.

Any issue with with aspnet open API package should be filed in this repository instead https://github.com/dotnet/aspnetcore

The aspnet team is working hard to upgrade to microsoft.openapi 2 as part of dotnet 10.

If you are using net < 10, you should remain on version 1.X of this package.

I agree the upgrade guide should clarify that aspect as this is one of the main scenarios for this library. Would anyone volunteer a pull request?

Frulfump:
[...]

PR #2451, opened 2025-07-23
Add warning for ASP.NET Core version `< 10.0`

Frulfump:
Contributes to #2442

- Fixes title of document, was `Upgrade guide to OpenAPI.NET 2.1 ` before
- Fixes [MD059](https://github.com/DavidAnson/markdownlint/blob/v0.38.0/doc/md059.md) helpful link text
- Remove trailing and unnecessary white space
- Typo correction of British English `behaviour` to American English `behavior`
- Fix basic markdown lint issues

Ideally there should be some type of markdown lint that runs on CI and a typo checker (added benefit would avoid typo squatting issues as seen in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5278)

I get a markdownlint warning MD025/single-title/single-h1: Multiple top-level headings in the same document for the `# Introduction` header since front matter is being used but I left it as is.

### DIFF
--- a/docs/upgrade-guide-2.md
+++ b/docs/upgrade-guide-2.md
@@ -98,13 +98,13 @@ mySchema.AnyOf = new List<IOpenApiSchema> { otherSchema };
 
 ## Reduced Dependencies
 
-In OpenAPI v1, it was necessary to include the Microsoft.OpenApi.Readers library to be able to read OpenAPI descriptions in either YAML or JSON.  In OpenAPI.NET v2, the core Microsoft.OpenAPI library can both read and write JSON.  It is only necessary to use the newly renamed [Microsoft.OpenApi.YamlReader](https://www.nuget.org/packages/Microsoft.OpenApi.YamlReader/) library if you need YAML support. This allows teams who are only working in JSON to avoid the additional dependency and therefore eliminate all non-.NET library references.
+In OpenAPI v1, it was necessary to include the Microsoft.OpenApi.Readers library to be able to read OpenAPI descriptions in either YAML or JSON. In OpenAPI.NET v2, the core Microsoft.OpenAPI library can both read and write JSON. It is only necessary to use the newly renamed [Microsoft.OpenApi.YamlReader](https://www.nuget.org/packages/Microsoft.OpenApi.YamlReader/) library if you need YAML support. This allows teams who are only working in JSON to avoid the additional dependency and therefore eliminate all non-.NET library references.
 
 Once the dependency is added, the reader needs to be added to the reader settings as demonstrated below
 
 ```csharp
 var settings = new OpenApiReaderSettings();
-settings.AddYamlReader();  
+settings.AddYamlReader();
 
 var result = OpenApiDocument.LoadAsync(openApiString, settings: settings);
 ```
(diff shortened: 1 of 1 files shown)


docs/upgrade-guide-2.md
```diff
@@ -660,5 +668,5 @@ OpenApiSchemaReference schemaRef = new OpenApiSchemaReference("MySchema")
 
 ## Feedback
 
-If you have any feedback please file a GitHub issue [here](https://github.com/microsoft/OpenAPI.NET/issues)
+If you have any feedback please file [a new GitHub issue](https://github.com/microsoft/OpenAPI.NET/issues)
```
Frulfump, line 671: This otherwise would trigger markdown lint warning https://github.com/DavidAnson/markdownlint/blob/v0.38.0/doc/md059.md which says link text should be descriptive which especially helps some screen readers.

Frulfump:

Frulfump:
@baywet @adrian05-ms I think the license/cla check is stuck, the bot never swooped in either.
>When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repos using our CLA.

https://github.com/microsoft/OpenAPI.NET?tab=readme-ov-file#contributing

I just yoloed the agree to see if something would happen...
(saw that comment in another PR)

baywet:

docs/upgrade-guide-2.md
```diff
@@ -1,27 +1,30 @@
...
 ---
 
 # Introduction
 
-We are excited to announce the new version of the OpenAPI.NET library!  
+We are excited to announce the new version of the OpenAPI.NET library!
 OpenAPI.NET v2 is a major update to the OpenAPI.NET library. This release includes a number of performance improvements, API enhancements, and support for OpenAPI v3.1.
 
+> [!WARNING]
+> If you are using this library with ASP.NET Core version `< 10.0` then you must remain on version `1.x` as it's not compatible.
```
martincostello, line 15: This isn’t _strictly_ true. It’s not compatible with the Microsoft.AspNetCore.OpenApi library, which depends on Microsoft.OpenApi v1.


If your application doesn’t use that (either directly, or transitively via a dependency such as Swashbuckle.AspNetCore), then you can use v2 with ASP.NET Core 8 and 9.

Frulfump: @martincostello Oh I thought there was some (non-absorbable)breaking changes affecting < ASP.NET Core 10 based on the maintainers conversation in the linked issue. Thanks I'll prepare a new PR!

(I was actually thinking about pinging you in the issue but didn't want to disturb you 😅)

martincostello: It's useable, it's just potentially lot of work due to the breaking changes, but other than where you depend on other dependencies using v1, it can be used. For example for Swashbuckle users, it definitely is breaking and not possible to use until we ship our release that supports v2 in November.

I'd just make the warning not so absolute in its wording.

Frulfump: Ah that makes sense, I agree. Could you prepare the PR instead I'm on a cellular connection and it's not very stable.

martincostello: Let's see how the discussion in the issue you tagged me on pans out first.

PR #2466, opened 2025-08-16
ci: adds release please workflow to replace sunset app

kilifu:


kilifu:

Issue #1643, opened 2024-04-29
Example value of an empty OpenApiArray disappears after serialization


CarlosZaldivar2:
The `Value` property of `OpenApiExample` is treated as optional and an optional empty `IEnumerable` is ignored when writing both JSON and YAML. I don't think it's correct, since an empty array or an empty object could be a real example of a response body and we may want to show it to the API's user.

Minimal OpenAPI document with an empty array:
```json
{
    "openapi": "3.0.1",
    "info": {
        "title": "Example API",
        "version": "1.0"
    },
    "paths": {
        "/products": {
            "get": {
                "responses": {
                    "200": {
                        "description": "Success",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "array",
                                    "items": {
                                        "type": "object",
                                        "properties": {
                                            "id": {
                                                "type": "string"
                                            }
                                        }
                                    }
                                },
                                "examples": {
                                    "Success response - no results": {
                                        "value": []
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

The same document after it has been read and written back by OpenAPI.NET:
```json
{
    "openapi": "3.0.1",
    "info": {
        "title": "Example API",
        "version": "1.0"
    },
    "paths": {
        "/products": {
            "get": {
                "responses": {
                    "200": {
                        "description": "Success",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "array",
                                    "items": {
                                        "type": "object",
                                        "properties": {
                                            "id": {
                                                "type": "string"
                                            }
                                        }
                                    }
                                },
                                "examples": {
                                    "Success response - no results": {}
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

MaggieKimani1:
I'm still struggling to see the value of writing out empty collection values as this seems to be by design(subject to change) in the lib. https://github.com/microsoft/OpenAPI.NET/blob/b27416226b94ea3a44c589e4ec81234ae43cfa91/src/Microsoft.OpenApi/Writers/OpenApiWriterExtensions.cs#L292
Required objects are still written out even when empty.

@baywet what do you think?

CarlosZaldivar2:
@MaggieKimani1 How can I show an empty array as a valid response example, when the library removes the empty array?

baywet:
@maggieKimani1 I believe this is pretty much the only valid scenario here.
[...]

PR #2473, opened 2025-08-22
Fix missing examples when one example is with an empty array.

melnikvitaly:
Fix issue when response has several example and one is with empty array as `value`
**Expected:**  all examples are present
**Actual:** only example with empty array is present and without summary and description.

Fixes https://github.com/microsoft/OpenAPI.NET/issues/1643 in another way without such side effect.
Please release also to v1.


melnikvitaly:

baywet:
@melnikvitaly the auto-merge might have masked my last message. So just in case: can you please start a new PR from support/v1 to support/v1 for the same fix on the v1 branch.

baywet:
nevermind, I put together #2478

PR #2487, opened 2025-09-08
Fix URL processing in InternalLoad method 

jkporter:
fix: Allow file scheme URLs for Load

Strengthens checks of URI scheme and adjusts behavior to support file scheme. Falls back to pre-existing behavior otherwise


jkporter:

src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
```diff
@@ -320,11 +320,21 @@ private static ReadResult InternalLoad(MemoryStream input, string format, OpenAp
...
+                        if (url.StartsWith("file:", StringComparison.OrdinalIgnoreCase))
+                        {
+                            fileName = new Uri(url).LocalPath;
+                        }
+                        else
+                        {
+                            fileName = url;
+                        }
+
+                        format = Path.GetExtension(fileName).Split('.').LastOrDefault();
```
baywet, line 335: I'm not sure we want to detect the format from the file extension. In theory I could have a .json file full of yaml.

baywet (commented):
Thanks for the contribution!

src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
jkporter: You’re right that relying on file extensions is brittle — though to be fair, that’s not new in this code; it’s simply carried over from the prior implementation. Without actually inspecting file contents, extensions are about the only conventionally available signal from the file system.

That said, in practice, newer YAML specs are strict supersets of JSON, so any valid JSON file is also valid YAML. That means parsing everything as YAML generally works fine and would make the JSON/YAML distinction less critical in most cases.

baywet: Thank you for the additional information.

There's already a content inspection method available for re-use should you need it.
Also, using the YAML reader to parse JSON, even if it works, has serious performance implications, and it'd be difficult to troubleshoot.

Would you mind giving a try to the content inspection please?

jkporter: I wasn’t suggesting we always parse as YAML—just pointing it out as a possible fallback in ambiguous cases. The intent of this PR was really just to add support for file URIs loading alongside regular system paths, so I hadn’t considered broader format detection changes.

I looked for the existing content inspection method but wasn’t able to find it in the codebase. That said, I’m open to using it if you can point me to where it lives. My only hesitation is that introducing it here might expand the scope beyond just adding file URI handling in manner that aligns with the existing way system paths are handled.

As a related thought, if there is a move toward content inspection, then applying it to the HTTP path should also be considered. The Content-Type headers can be just as ambiguous or error-prone, so it would make sense in regards to consistency.

baywet: Let's do something then: what do you think about removing this line for the format so we can unblock this pull request and then follow up with another pull request/issue for the format aspect?

baywet: @jkporter gentle reminder on the above :)

jkporter: @baywet I've revised the code to set `format` to `null`.

src/Microsoft.OpenApi/Reader/OpenApiModelFactory.cs
```diff
@@ -320,11 +320,21 @@ private static ReadResult InternalLoad(MemoryStream input, string format, OpenAp
...
+                        if (url.StartsWith("file:", StringComparison.OrdinalIgnoreCase))
+                        {
+                            fileName = new Uri(url).LocalPath;
+                        }
+                        else
+                        {
+                            fileName = url;
+                        }
+
+                        format = null;
```
baywet:
```suggestion
                        format = Path.GetExtension(fileName).Split('.').LastOrDefault();;
```

I'm sorry, I think I go confused by the github diff view last time and missed the fact we already had similar code in place.

jkporter: Yeah, I thought it was odd you were advocated for removing the pre-existing behavior though setting the format to `null` interestingly works perfectly fine.

baywet: I think the fact that it was there in the first place is debatable, but it's orthogonal to your change, which is why removing it now does not make a lot of sense. And yet, that still works without it because of the format inspection (reading the stream). That inspection is "more accurate" but has performance impacts. Thanks for reverting the revert.

PR #2508, opened 2025-09-18
Fix OpenApiEncoding explode property serialization

kammerjaeger:
Fix: Explode serialization is wrong when style is "form"

This fixes the same problem as https://github.com/microsoft/OpenAPI.NET/pull/544 but for OpenApiEncoding.
The spec describes different behavior depending on if it is set or not (see: https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-12) and also different defaults depending on the style value.
This also needs to be cherry picked to main, I just need it for v1 but will create a pr to main with the change if this pr is fine.

Please let me know what you think.

PR #2512, opened 2025-09-19
Fix OpenApiEncoding explode property serialization (to main)

kammerjaeger:
Fix: Explode serialization is wrong when style is "form"

This fixes the same problem as https://github.com/microsoft/OpenAPI.NET/pull/544 but for OpenApiEncoding.
The spec describes different behavior depending on if it is set or not (see: https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-12) and also different defaults depending on the style value.

Cherry pick for https://github.com/microsoft/OpenAPI.NET/pull/2508 to the main branch

I don't have dotnet 10 installed yet so I hope this works without it.


baywet (approved):
Thanks for porting the fix!

PR #2524, opened 2025-10-03
Enhance server creation logic and add unit tests

### LINKED ISSUE #1294: When BaseUrl has a port, the default server url doesn't contain the port
**Describe the bug**
When setting a BaseUrl with a port, and the description file doesn't contain any servers, the port is omitted from the default server.

**To Reproduce**
Steps to reproduce the current behavior:
Create OpenApiReaderSettings with BaseUrl set to http://demo.testfire.net:8080
Create any OpenApiReader instance with these settings
Attempt to deserialize this properties file: http://demo.testfire.net:8080/swagger/properties.json

The resulting OpenApiDocument.Servers will contain the server "http://demo.testfire.net" (without the port number).

**Expected behavior**
The resulting OpenApiDocument.Servers should contain the server "http://demo.testfire.net:8080"

Mahdigln:
Fix port omission from BaseUrl in V2 server deserialization

 Problem
When setting a `BaseUrl` with a non-standard port and the OpenAPI V2 document doesn't contain any server definitions, the port number was being omitted from the default server URL.

Example:
- Input: `BaseUrl = "http://demo.testfire.net:8080"`
- Current behavior: Server URL = `"http://demo.testfire.net"` ❌
- Expected behavior: Server URL = `"http://demo.testfire.net:8080"` ✅

 Root Cause
The `MakeServers` method in `OpenApiDocumentDeserializer.cs` was using `UriComponents.NormalizedHost` which only extracts the hostname without the port number.

 Solution
Changed `UriComponents.NormalizedHost` to `UriComponents.Host | UriComponents.Port` to preserve port information when extracting host details from the BaseUrl. This approach is consistent with the existing pattern used in `OpenApiDocument.WriteHostInfoV2`.

Additionally, fixed the early return logic to allow server creation from BaseUrl even when no host, basePath, or schemes are defined in the document.

 Changes
1. **`src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs`**:
   - Line 154: Changed from `UriComponents.NormalizedHost` to `UriComponents.Host | UriComponents.Port`
   - Line 144: Modified early return condition to check for `defaultUrl != null`

2. **`test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiServerTests.cs`**:
   - Added 3 new test cases to verify port preservation in various scenarios

 Testing
- All existing tests pass ✅
- Added comprehensive test coverage for port preservation scenarios
- 16/16 OpenApiServerTests passing
- 78/78 V2Tests passing

Fixes #1294

### DIFF
--- a/src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
+++ b/src/Microsoft.OpenApi.Readers/V2/OpenApiDocumentDeserializer.cs
@@ -145,8 +145,8 @@ namespace Microsoft.OpenApi.Readers.V2
                 basePath = "/";
             }
 
-            // If nothing is provided, don't create a server
-            if (host == null && basePath == null && schemes == null)
+            // If nothing is provided and there's no defaultUrl, don't create a server
+            if (host == null && basePath == null && schemes == null && defaultUrl == null)
             {
                 return;
             }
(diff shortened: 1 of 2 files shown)


baywet:
@Mahdigln why did you close this one? While we're happy to still accept fixes for v1, the main active branch for v2 will need to have the fix as well.

For others reading #2525

PR #2526, opened 2025-10-03
fix: missing deserialization for header content property in 3.1 and 3.0

baywet:


baywet:
@copilot add deserializer unit tests to prevent regressions when parsing the content property for the OpenApiHeader content field, both for version 3 and 3.1

adrian05-ms (approved):

PR #2530, opened 2025-10-06
Improve server creation and URL handling logic-V2

### LINKED ISSUE #1294: When BaseUrl has a port, the default server url doesn't contain the port

Mahdigln:
Updated the `MakeServers` method to include checks for `defaultUrl` and refined host validation. Modified the `BuildUrl` method to correctly extract ports and remove default ports from URLs. Added unit tests in `OpenApiServerTests.cs` to ensure proper handling of base URLs and ports across various scenarios.

fixes #1294

### DIFF
--- a/src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
+++ b/src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
@@ -124,14 +124,14 @@ namespace Microsoft.OpenApi.Reader.V2
                 basePath = "/";
             }
 
-            // If nothing is provided, don't create a server
-            if (host == null && basePath == null && schemes == null)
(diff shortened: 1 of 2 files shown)


baywet (approved):
Thanks for the contribution!

PR #2684, opened 2026-01-17
Fix Typo in IOpenApiReadOnlyExtensible.cs

### LINKED ISSUE #123: OpenApiAny vs primitive type inconsistency
@darrelmiller @xuzhg 

We have been using an OpenApiAny type (e.g. OpenApiString, OpenApiDouble) to represent strings, numbers, arrays, etc. in the property that can be of type OpenApiAny.

However, in other properties of a normal object, we simply use the C# primitive types like string, int.

This is semantically inconsistent in my opinion. An "Any" type in the spec simply means it can be anything. If we use OpenApiString to mean a string inside that "Any" type, we should exclusively use OpenApiString everywhere. If we use the C# primitive, then we should use the C# primitive everywhere.

Using OpenApiAny allows us to control the type to match what JSON schema allows and add some types recognized by the spec like Password, but will be a pain for our customers if they need to assign values with OpenApiAny everywhere since it's less intuitive to do so compared to just C# primitives.

say25:
# Pull Request

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Changes Made

-
-
-

## Testing

- [ ] Unit tests added/updated
- [ ] All existing tests pass

## Checklist

- [ ] My code follows the code style of this project
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

## Versions applicability

- [ ] My change applies to the version 3.X of the library, if so PR link:
- [ ] I have evaluated the applicability of my change against the other versions above.

See [the contributing guidelines](https://github.com/microsoft/OpenAPI.NET/blob/main/CONTRIBUTING.md) for more information about how patches are applied across multiple versions.

### DIFF
--- a/src/Microsoft.OpenApi/Interfaces/IOpenApiReadOnlyExtensible.cs
+++ b/src/Microsoft.OpenApi/Interfaces/IOpenApiReadOnlyExtensible.cs
@@ -3,7 +3,7 @@
 namespace Microsoft.OpenApi;
 
 /// <summary>
-/// Represents an Extensible Open API element elements can be rad from.
+/// Represents an Extensible Open API element elements can be read from.
 /// </summary>
 public interface IOpenApiReadOnlyExtensible
 {


baywet:
> Thank you for the contribution! Would you mind please doing the same on the support/v2 branch please?

I put together #2685

PR #2706, opened 2026-01-22
feat(securityscheme): add oauth2MetadataUrl support (OpenAPI 3.2)

### LINKED ISSUE #2694: OpenAPI 3.2: `oauth2MetadataUrl` Missing from `Microsoft.OpenApi.OpenApiSecurityScheme`
## Summary

`Microsoft.OpenApi` can serialize OpenAPI **3.2** documents, but the **Security Scheme Object** is missing support for the OpenAPI 3.2 field:

- `oauth2MetadataUrl` (type: `string`, format: `uri`, applies to `type: oauth2`)

This prevents consumers from producing spec-complete OpenAPI 3.2 documents when describing OAuth2 servers using RFC 8414 metadata.

## Spec Reference

OpenAPI 3.2.0 adds `oauth2MetadataUrl` to the Security Scheme Object. citeturn0search0turn0search10
The field is defined as a URL to the OAuth2 authorization server metadata (RFC 8414). citeturn0search5turn0search2

## Current Behavior

`OpenApiSecurityScheme` does not expose a property for `oauth2MetadataUrl`, and the serializer does not emit it for OpenAPI 3.2 output.

Even when using `SerializeAsV32`, the switch over `SecuritySchemeType` only handles:

- `apiKey`: `name`, `in`
- `http`: `scheme`, `bearerFormat`
- `oauth2`: `flows`
- `openIdConnect`: `openIdConnectUrl`

There is no native way to model and serialize `oauth2MetadataUrl` other than vendor extensions.

## Expected Behavior

When `Type == SecuritySchemeType.OAuth2` and the target spec version is **OpenAPI 3.2+**, the library should allow setting and serializing:

```yaml
components:
  securitySchemes:
    oauth:
      type: oauth2
      oauth2MetadataUrl: https://idp.example.com/.well-known/oauth-authorization-server
      flows:
        clientCredentials:
          tokenUrl: https://idp.example.com/oauth/token
          scopes: {}
```

## Proposed API Change

Add a nullable `Uri` property to `OpenApiSecurityScheme`:

```csharp
public Uri? OAuth2MetadataUrl { get; set; }
```

### Serialization (OpenAPI 3.2+ only)

In `SerializeInternal(...)`, under `case SecuritySchemeType.OAuth2:` write the property **before/after** flows:

```csharp
if (version >= OpenApiSpecVersion.OpenApi3_2)
{
    writer.WriteProperty("oauth2MetadataUrl", OAuth2MetadataUrl?.ToString());
}
writer.WriteOptionalObject(OpenApiConstants.Flows, Flows, callback);
```

### Parsing / Reading

If the library includes readers/deserializers for security schemes, they should also recognize `oauth2MetadataUrl` when parsing OpenAPI 3.2 documents into `OpenApiSecurityScheme`.

## Why This Matters

OAuth2 Authorization Server Metadata (RFC 8414) is widely used to publish endpoints and capabilities. OpenAPI 3.2 explicitly supports linking to that metadata; without this field, OpenAPI 3.2 documents generated with `Microsoft.OpenApi` cannot fully represent the spec-defined OAuth2 security scheme information. citeturn0search5turn0search7

## Workarounds Today

- Use `Extensions["x-oauth2MetadataUrl"] = ...` (non-standard)
- Put the URL in `description` (lossy / not machine-readable)

## Request

Please add first-class support for `oauth2MetadataUrl` to `OpenApiSecurityScheme` and include it in OpenAPI 3.2 serialization/parsing for OAuth2 security schemes.

Thank you!

mdaneri:
# Pull Request

## Description
Adds first-class support for OpenAPI 3.2 `oauth2MetadataUrl` on OAuth2 security schemes, including serialization and OpenAPI 3.2 reader support.

## Type of Change
- [ ] Documentation update

## Related Issue(s)
Fixes #2694

## Changes Made
- Add `OAuth2MetadataUrl` to `OpenApiSecurityScheme`/`IOpenApiSecurityScheme` and reference wrapper.
- Serialize `oauth2MetadataUrl` only for OpenAPI 3.2+ when `type: oauth2`.
- Parse `oauth2MetadataUrl` in the OpenAPI 3.2 reader.
- Add unit tests for serialization and parsing.

## Testing
- [x] All existing tests pass

## Checklist

## Versions applicability

- [x] My change applies to the version 3.X of the library, if so PR link:

## Additional Notes
Unit tests executed locally:
- `test/Microsoft.OpenApi.Tests/Models/OpenApiSecuritySchemeTests.cs`
- `test/Microsoft.OpenApi.Readers.Tests/V32Tests/OpenApiSecuritySchemeTests.cs`


src/Microsoft.OpenApi/Models/OpenApiSecurityScheme.cs
```diff
@@ -118,7 +122,12 @@ private void SerializeInternal(IOpenApiWriter writer, OpenApiSpecVersion version
                     break;
                 case SecuritySchemeType.OAuth2:
                     // This property apply to oauth2 type only.
+                    // oauth2MetadataUrl
                     // flows
+                    if (version >= OpenApiSpecVersion.OpenApi3_2)
+                    {
+                        writer.WriteProperty(OpenApiConstants.OAuth2MetadataUrl, OAuth2MetadataUrl?.ToString());
```
baywet, line 131: this should probably also throw in case the property is present but the version is lower than expected to be consistent with other properties like this. And unit tests for that.

The alternative being to serialize it as an extension. But we shouldn't silently drop information.

baywet (changes requested):
Thanks for the contribution!

In addition to the requested change, you'll need to refresh the public API export.

mdaneri:
Always I forget about that 😓

mdaneri:
That test looks interesting! I might just borrow it and incorporate it into my project.

baywet:
> That test looks interesting! I might just borrow it and incorporate it into my project.

The dotnet team actually made it an analyzer, so you get feedback on build and not on tests anymore. We just haven't taken the time to update things here. See this as an example.

https://github.com/BinkyLabs/openapi-overlays-dotnet/blob/b9635b903f375498673a3ef801d7d862879e2e3b/src/lib/BinkyLabs.OpenApi.Overlays.csproj#L45

src/Microsoft.OpenApi/Models/OpenApiSecurityScheme.cs
mdaneri: I can add an extension for that

test/Microsoft.OpenApi.Tests/PublicApi/PublicApi.approved.txt
```diff
@@ -288,6 +288,7 @@ namespace Microsoft.OpenApi
         Microsoft.OpenApi.OpenApiOAuthFlows? Flows { get; }
         Microsoft.OpenApi.ParameterLocation? In { get; }
         string? Name { get; }
+        System.Uri? OAuth2MetadataUrl { get; }
```
baywet: This change is technically source breaking. Even though we recommend not to take a dependency on those interfaces. This still poses a problem with the transitive dependency aspect of things. I.E. if a dependency like aspnetcore.openapi or swashbuckle is accessing other properties in the DOM of type IOpenApiSecurityScheme and have not re-compiled after we ship that change, they'll get exceptions like MethodNotFound/ or that the interface is not implemented by the types that actually implement it.

baywet: One way we could try to mitigate this is through Default Interface Implementations, but they are only supported for net core 3.0 onwards. And we currently ship netstandard2.0 and net8.0. We'd need to figure out something for netstandard2.0

baywet: And even if we added the default implementation, it'd still remain binary breaking.

baywet: This effectively leaves us with the following choices:
- major rev, and the cost associated with supporting v3.
- major rev, and not support v3, only a valid approach as long as net11 does not have support for v3. And we should ensure we don't have any other breaking changes that need to be slated.
- decline this PR, and have an incomplete DOM.
- revert the interface change, but keep the rest. That degrades the experience quite a bit forcing type assertions/casts to access the property. And we'd clean this up in a future major version (OpenAPI 3.3?)
- introduce an additional interface to facilitate the type assertions. (variation of the above)

baywet: CC @martincostello in case you have a better suggestion here

baywet:
@mdaneri would you mind updating the benchmarks please?

```shell
cd performance/benchmark
dotnet run -c Release
```

(and commit the changes)

test/Microsoft.OpenApi.Tests/PublicApi/PublicApi.approved.txt
martincostello: Introduce this via a new interface instead?

Thinking out loud, you could add an `IOAuth2MetadataProvider` interface:

```csharp
public interface IOAuth2MetadataProvider
{
    Uri? OAuth2MetadataUrl { get; }
}
```

Then implement that interface on whatever concrete types need it, and require users to type cast to test for it for implementations that care about it:

```csharp
if (thing is IOAuth2MetadataProvider oauth2Metadata)
{
    // whatever
}
```

As already noted, adding it to other interfaces in their inheritance chain is source breaking and DIMs are problematic for TFMs that are too old to support it (i.e. for .NET Framework via `netstandard2.0`).

You could potentially avoid that by making the new interface conditionally compiled for `#if NET` and use DIMs, but that would make the API surface non-symmetrical for different TFMs.

baywet: Thank you for the additional information.

Yes that was the last option I was listing even though I didn't explicitly document it. Let's go with that then.

@mdaneri, would you mind introducing a new interface please? Additionally please:
- leave a super explicit doc comment this is a "temporary" interface.
- leave a todo comment to remove and collapse the interface with the existing one on the next major version.
- update the doc comment of the existing interface to point to this one. Example with type assertions would be nice as well. So people are not surprised by this pattern, know that's mostly because we (Microsoft) messed up, and know where to find the missing property.

baywet:
Hi @mdaneri
I have migrated the repository to use code analyzers in #2711, this should provide earlier feedback (on build instead of on tests) thanks to the warn as errors.
It also means there are conflicts with this PR now, apologies for the additional work.

To drive this to completion, whenever you have time, could you please:
1. address the comment about the interface split.
2. rebase (drop the old public APIs for the tests)
3. update the shipped apis with your changes.

baywet:
Gentle ping on this one @mdaneri

mdaneri:
I'll work on this tomorrow

mdaneri:
> @mdaneri would you mind updating the benchmarks please?
> 
> ```shell
> cd performance/benchmark
> dotnet run -c Release
> ```
> 
> (and commit the changes)

Done

test/Microsoft.OpenApi.Tests/PublicApi/PublicApi.approved.txt
```diff
@@ -0,0 +1,2164 @@
+[assembly: System.Reflection.AssemblyMetadata("IsTrimmable", "True")]
```
baywet: please remove this extraneous file.

baywet: @mdaneri sorry if I wasn't clear, I meant delete the file.

baywet:
@mdaneri can you please also run the following:

```shell
# add the missing public api entries
dotnet format --diagnostics RS0016
# discard changes to cs files to avoid creating conflicts
git checkout *.cs
```

and then commit the text file?

Sorry for not providing the context here: I moved this repos from using custom tests for the public API surface to using the "new" dotnet analyzers. And your PR was originally created before that move, hence the "migration" here.

mdaneri:
Could you please update the contributing guide? It seems like some important information is missing. I’m afraid I might forget all of it next time! :)

baywet:
> Could you please update the contributing guide? It seems like some important information is missing. I’m afraid I might forget all of it next time! :)

This is an excellent suggestion, working on it.

Also adding this section for the failing benchmarks, would you mind updating them please?

## Updating the benchmark information

To ensure performance of the library does not degrade over time, we have continuous benchmarks running. You might see the continuous integration failing if your pull request changed any model under __src/Microsoft.OpenApi/Models__.

```txt
Benchmark result for EmptyApiSchema does not match the existing benchmark result (original!=new). Allocated bytes differ: 408 != 416
```

To update the benchmarks, run the following script:

```shell
cd performance/benchmark
dotnet run -c Release
```

Then commit the report files using a "chore" commit.

baywet:
also put together #2734 to address the contributing documentation gaps.

PR #2730, opened 2026-02-18
fix/unevaluated to v2

baywet:
port of #2728 to v2


baywet:
@copilot Update the benchmark results

To do so, run the following script:

```shell
cd performance/benchmark
dotnet run -c Release
```

Then commit the report files using a "chore" commit.

adrian05-ms (approved):

Issue #414, opened 2019-07-26
OpenApiComparer is not implemented


adrientoub:
Hello,

My team is using this repository to generate our OpenAPI document, and we started using the [OpenApiComparer](https://github.com/microsoft/OpenAPI.NET/blob/master/src/Microsoft.OpenApi/Services/OpenApiComparer.cs) recently in our project. This method returns a value so we believed it was working until recently when after making a change this still returned an empty list.

After looking at the implementation it's quite clear that this method does nothing but it is documented as being a method that "Compares two OpenApiDocuments and returns a list of differences.".

I believe this method should be removed or should throw instead.

After looking a bit more, there seems to be a valid implementation of this in the vnext branch that was never merged into master.

darrelmiller:
Huh!  I had no idea.  I don't whether it is in finished enough to consider merging in.  If it is I can create a PR based on it.

Backs:
Any new information? Looks lite this code is productoin ready and was added 2 year ago, but still not in master

darrelmiller:
This code has been merged into the vnext branch.  I would like to include it in the v1.2 release but there are still some failing tests.  I have marked them as skip for the moment, but they need to be fixed before we can release.

FLAMESpl:
Hi, do you have ETA for complete feature or at least new update? Latest prerelease available on nuget throws StackOverflowException for my swagger doc.

darrelmiller:
My current thinking is to move comparer capability out to a separate package.  There are several contributed solutions and I don't feel like picking a winner.

FLAMESpl:
What is the status of this task? Is it shelved or is it in backlog for considerable future?

nikjw-lmk:
Did you end up moving the comparer capability into another repo? It doesn't appear to be here anymore so was wondering if it can be found somewhere else. Thanks.

darrelmiller:
Hey @nikjw-lmk . I moved the code into a feature branch https://github.com/microsoft/OpenAPI.NET/tree/feature/comparer/src/Microsoft.OpenApi/Services  I haven't moved it out into a separate module.  I have been focused on trying to get the OpenAPIWorkspace into a releasable state.

Chrille79:
Any progress?

PR #2746, opened 2026-02-20
fix(library): preserve patternProperties semantics in OpenAPI 2.0 and 3.0 serialization

### LINKED ISSUE #2716: OpenAPI 3.0: Preserve PatternProperties via Vendor Extension during Serialization
## Summary

When generating **OpenAPI 3.0** documents using OpenAPI.NET, schemas that define
`PatternProperties` lose their semantic meaning during serialization.

Currently, the serializer either:

- drops the information entirely, or
- emits `additionalProperties: false`, which is semantically incorrect for
  dictionary-like schemas with patterned keys.

This issue proposes preserving `PatternProperties` **in a spec-compliant way**
for OpenAPI 3.0 by emitting a vendor extension, while keeping a reasonable
fallback for standard tooling.

---

## Background

`patternProperties` is a core JSON Schema feature that has existed since early
drafts and is fully supported in:

- JSON Schema (all modern drafts)
- OpenAPI **3.1+** (which aligns with JSON Schema 2020-12)

However, **OpenAPI 3.0** uses a restricted schema dialect and does **not**
support `patternProperties`.

OpenAPI.NET already exposes:

```csharp
OpenApiSchema.PatternProperties
```

but this information is effectively lost when serializing to OpenAPI 3.0.

---

## Current Behavior

Given a schema intent like:

```json
{
  "type": "object",
  "patternProperties": {
    "^[a-z][a-z0-9_]*$": {
      "type": "integer",
      "format": "int32"
    }
  }
}
```

OpenAPI 3.0 serialization may produce:

```json
{
  "type": "object",
  "additionalProperties": false
}
```

This output is:

- syntactically valid OpenAPI 3.0
- **semantically incorrect**, because it forbids all dynamic keys

---

## Expected / Proposed Behavior

When serializing **OpenAPI 3.0** and `OpenApiSchema.PatternProperties` is present:

### 1. Preserve semantic intent using a vendor extension

Emit a vendor extension such as:

```json
"x-jsonSchema-patternProperties": {
  "^[a-z][a-z0-9_]*$": {
    "type": "integer",
    "format": "int32"
  }
}
```

Vendor extensions are explicitly allowed in OpenAPI 3.0 and are the only
spec-compliant mechanism to preserve this information.

### 2. Provide a best-effort OpenAPI 3.0 fallback

Additionally emit one of the following (depending on feasibility):

- If all pattern properties share the same schema:

  ```json
  "additionalProperties": {
    "type": "integer",
    "format": "int32"
  }
  ```

- Otherwise:

  ```json
  "additionalProperties": true
  ```

This ensures existing OpenAPI 3.0 tooling continues to function predictably,
while advanced consumers can still read the extension.

---

## Non-Goals

- This proposal does **not** attempt to make `patternProperties` a first-class
  OpenAPI 3.0 keyword (that would be non-spec).
- This proposal does **not** change OpenAPI 3.1+ behavior.

---

## Scope of Change

- Affects **serialization only**
- Only for `OpenApiSpecVersion.OpenApi3_0`
- No changes to the public object model
- No breaking changes

---

## Benefits

- Prevents silent loss of schema semantics
- Produces valid OpenAPI 3.0 output
- Allows downstream tools to recover JSON Schema intent
- Aligns with OpenAPI’s intended use of vendor extensions

---

## Suggested Extension Name

```text
x-jsonSchema-patternProperties
```

Rationale:

- Explicitly indicates JSON Schema keyword preservation
- Avoids ambiguity with standard OpenAPI fields
- Minimizes risk of collision with other extensions

---

## References

- OpenAPI 3.0 Vendor Extensions
  <https://swagger.io/docs/specification/v3_0/openapi-extensions/>

- OpenAPI 3.1 JSON Schema Alignment
  <https://spec.openapis.org/oas/v3.1.0.html>

- JSON Schema `patternProperties`
  <https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties>

---

## Closing Notes

This change would allow OpenAPI.NET to preserve important schema semantics
without violating the OpenAPI 3.0 specification, and without requiring users
to fork or post-process the library output.

Thank you for considering this enhancement.

mdaneri:
# Pull Request

## Description
Preserves `PatternProperties` semantic intent when serializing schemas to OpenAPI 3.0 and OpenAPI 2.0, where `patternProperties` is not a first-class keyword.

This change writes `x-jsonSchema-patternProperties` for downlevel specs and emits a safe `additionalProperties` fallback to avoid incorrect `additionalProperties: false` output for patterned dictionary-like schemas.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Related Issue(s)
- Closes the reported issue about loss of `PatternProperties` semantics during OpenAPI 3.0 serialization.

## Changes Made
- Added `OpenApiConstants.PatternPropertiesExtension` with value `x-jsonSchema-patternProperties`.
- Updated schema serialization for OpenAPI 3.0 and OpenAPI 2.0 to emit the extension when `PatternProperties` exists.
- Added best-effort fallback behavior:
  - emits `additionalProperties` schema when all pattern schemas are equivalent;
  - otherwise emits `additionalProperties: true`.
- Kept OpenAPI 3.1+ behavior unchanged (`patternProperties` remains a standard keyword).
- Added/updated unit tests for v2/v3/v31 serialization scenarios.

## Testing
- [x] All existing tests pass

Executed:
- Focused tests for patternProperties serialization (v2/v3/v31)
- Full solution tests via `dotnet test Microsoft.OpenApi.sln --collect:"XPlat Code Coverage"`

## Checklist

## Versions applicability

## Additional Notes
This is a serialization-only change; no public object model behavior was changed beyond adding one constants entry.


src/Microsoft.OpenApi/Models/OpenApiConstants.cs
```diff
@@ -135,6 +135,11 @@ public static class OpenApiConstants
         /// </summary>
         public const string UnevaluatedPropertiesExtension = "x-jsonschema-unevaluatedProperties";
 
+        /// <summary>
+        /// Extension: x-jsonSchema-patternProperties
+        /// </summary>
+        public const string PatternPropertiesExtension = "x-jsonSchema-patternProperties";
```
baywet:
```suggestion
        public const string PatternPropertiesExtension = "x-jsonschema-patternProperties";
```

(to align with the other property, plus mixing camel and kebab casing is a bit odd)

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -611,6 +634,54 @@ internal void WriteJsonSchemaKeywords(IOpenApiWriter writer)
...
+            {
+                return null;
+            }
+
+            using var stringWriter = new StringWriter(CultureInfo.InvariantCulture);
+            var jsonWriter = new OpenApiJsonWriter(stringWriter, new OpenApiJsonWriterSettings { Terse = true });
+            serializableSchema.SerializeAsV31(jsonWriter);
+
+            return JsonNode.Parse(stringWriter.ToString());
+        }
```
baywet, line 681: I think this approach is going to seriously degrade performance when we hit the revealing condition (which the performance tests are not). Plus it's not future proof in the sense that properties will evolve from 3.1 to 3.2, and eventually 3.3, etc...

I'd much rather see an IEqualityComparer be implemented, plus that's been a request for this lib for a while, so it'd be a move in the right direction, enable reusability, etc... #414

baywet (changes requested):
Thanks for the contribution! Here are some comments to help bring this one to completion. Let me know if you have any additional comments or questions.

baywet:
@mdaneri gentle ping on this one :)

baywet:
@mdaneri here is what I've done:
- I've put copilot to the task to at least emit the pattern properties as an extension in #2755 so we don't loose the information in earlier versions.
- I've handled the merge into this PR.
- I've handled automatic feedback on this PR.

All in the goal to unblock the releases which have been waiting for a couple of weeks now.

There's one comment to address from me about the roundtrip serialization before we can merge this one. I'll let you take care of this one whenever you have time.

baywet:
Closing due to inactivity. Don't hesitate to re-open if this is still relevant.

PR #2825, opened 2026-04-16
ci: upgrades repository to net10

baywet:

PR #2826, opened 2026-04-16
ci: upgrades repository to net10

baywet:
port of #2825


.github/workflows/ci-cd.yml
```diff
@@ -29,27 +34,32 @@ jobs:
...
+          dotnet test Microsoft.OpenApi.slnx -c Release -v n
 
   validate-trimming:
     name: Validate Project for Trimming
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v6
 
-      - name: Setup .NET
+      - name: Setup .NET 8
```
adrian05-ms, line 51: Shouldn't we remove this? won't the step above overwrite this?

baywet: No, that's because our library is that we're publishing our targeting this version, so we need both versions to be installed locally for the whole process to go through successfully. 
Sometimes some versions of the SDK are pre-installed on the machine but when those things gets changed, this is when your pipelines start breaking for no reason at all. 
You should always install all the versions you need and if a version is already present it will result in no operation. If it's absent it will be installed, but at least you insure this way that you have all the things you need to produce the artifact you want to produce.

baywet: Additionally, most sdks will stay from.net to node.js and etc. Are able to have multiple versions co-located in the same box without resulting into conflicts or suppressions

adrian05-ms (approved):

PR #2828, opened 2026-04-17
feat: add TOML v1.0 serialization support

### LINKED ISSUE #123: OpenApiAny vs primitive type inconsistency

Thyago-Oliveira-Perez:
Adds OpenApiTomlWriter (mirrors OpenApiJsonWriter / OpenApiYamlWriter),
       a new OpenApiConstants.Toml constant, and SerializeAsTomlAsync extension
       methods so callers can export an OpenAPI document as TOML.

       Public API declared in PublicAPI.Unshipped.txt; RS0026 suppressed for
       consistency with existing SerializeAsJson/YamlAsync overloads.

# Pull Request

## Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Changes Made

- Added OpenApiTomlWriter to support writing OpenAPI documents in TOML format
- Introduced OpenApiConstants.Toml constant
- Added SerializeAsTomlAsync extension methods (string and stream overloads)
- Integrated TOML into existing SerializeAsync format selection
- Declared new public APIs in PublicAPI.Unshipped.txt
- Suppressed RS0026 for consistency with existing JSON/YAML serialization methods
- Added unit tests covering TOML writer behavior and serialization

## Testing

- [X] Unit tests added/updated
- [X] Manual testing performed
- [X] All existing tests pass
- [X] Create a Demo project to test the changes and how it performs. Link: https://github.com/Thyago-Oliveira-Perez/OpenApiTomlDemo

## Checklist

- [X] My code follows the code style of this project
- [X] I have performed a self-review of my own code
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

## Versions applicability

- [X] My change applies to the version 3.X of the library, if so PR link:
- [ ] I have evaluated the applicability of my change against the other versions above.

See [the contributing guidelines](https://github.com/microsoft/OpenAPI.NET/blob/main/CONTRIBUTING.md) for more information about how patches are applied across multiple versions.

### DIFF
--- a/src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
+++ b/src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
@@ -42,6 +42,20 @@ namespace Microsoft.OpenApi
         }
 
         /// <summary>
+        /// Serializes the <see cref="IOpenApiSerializable"/> to the Open API document (TOML) using the given stream and specification version.
+        /// </summary>
+        /// <typeparam name="T">the <see cref="IOpenApiSerializable"/></typeparam>
+        /// <param name="element">The Open API element.</param>
+        /// <param name="stream">The output stream.</param>
+        /// <param name="specVersion">The Open API specification version.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        public static Task SerializeAsTomlAsync<T>(this T element, Stream stream, OpenApiSpecVersion specVersion, CancellationToken cancellationToken = default)
+            where T : IOpenApiSerializable
+        {
+            return element.SerializeAsync(stream, specVersion, OpenApiConstants.Toml, cancellationToken);
+        }
+
+        /// <summary>
         /// Serializes the <see cref="IOpenApiSerializable"/> to the Open API document using
         /// the given stream, specification version and the format.
         /// </summary>
@@ -91,6 +105,7 @@ namespace Microsoft.OpenApi
                 OpenApiConstants.Json when settings is OpenApiJsonWriterSettings jsonSettings => new OpenApiJsonWriter(streamWriter, jsonSettings),
                 OpenApiConstants.Json => new OpenApiJsonWriter(streamWriter, settings),
                 OpenApiConstants.Yaml => new OpenApiYamlWriter(streamWriter, settings),
+                OpenApiConstants.Toml => new OpenApiTomlWriter(streamWriter, settings),
                 _ => throw new OpenApiException(string.Format(SRResource.OpenApiFormatNotSupported, format)),
             };
             return element.SerializeAsync(writer, specVersion, cancellationToken);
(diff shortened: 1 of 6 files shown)


Thyago-Oliveira-Perez:

baywet (commented):
Hi @Thyago-Oliveira-Perez 
Thank you for using the SDK and for reaching out.
Thanks for the contribution!

I'm a bit mixed on this proposal and wished you had started by creating an issue first to have a discussion.

OpenAPI descriptions are by design deeply nested, [and TOML does not seem to be designed for that](https://jsontoyamlconverter.com/yaml-vs-json/toml/). Additionally, the ecosystem support for TOML, and demand for it is low to inexistant. (no prior issue on the matter)

As for this PR: I'm really questioning the usefulness of having the ability to serialize to, but not deserialize from TOML. I understand this might be useful to convert and push to other systems that only accept TOML, but I'd be surprised to see any system that only accepts TOML in 2026. Implementing the deserialization would need to be done in a companion library like yaml is done today. And would allow us to validate roundtrips actually work.

Before we make any further code changes, can you please expand on why you think TOML is necessary? Why it needs to be in the "official library" (and not in a community owned peer)? And what you're trying to achieve (beyond supporting TOML)?

Thanks!

Thyago-Oliveira-Perez:
Thanks for the feedback @baywet.

That’s fair, and I agree I should have opened an issue first.

The motivation here isn’t really adding TOML support for the sake of it, but solving a "problem" I'm having.

I’ve been using OpenAPI as input for AI-assisted workflows (mainly generating frontend code), and JSON/YAML aren’t great in that context. TOML ended up being better in that context as it is more predictable format and consume less tokens for prompts. So the idea is more about having a readable/export-friendly projection.

On serialization without deserialization, I agree it’s a limitation. My assumption was that this would behave more like an export format, but I see your point that proper support would likely require a companion package (like YAML).

Regarding being in the official library: my thinking was mostly around keeping it close to the existing serialization logic to avoid fragmentation, but I’m totally open to moving this to a separate/experimental package if that fits better.

Happy to open an issue and continue the discussion there if you prefer.

baywet:
Hi @Thyago-Oliveira-Perez 
Thank you for the additional information.

> TOML ended up being better in that context as it is more predictable format and consume less tokens for prompts.

I'm not sure how that claim stands? Depending on the scenarios, and from cursory search, TOML appears to use the same number of characters (blank spaces excluded) than YAML or JSON. Except for deeply nested structures where it's worse. Do you have examples that would illustrate this?

> On serialization without deserialization, I agree it’s a limitation. My assumption was that this would behave more like an export format, but I see your point that proper support would likely require a companion package (like YAML).

From an integration tests perspective, having the ability to roundtrip often leads to the discovery of defects, improving the quality.

> Regarding being in the official library: my thinking was mostly around keeping it close to the existing serialization logic to avoid fragmentation, but I’m totally open to moving this to a separate/experimental package if that fits better.

I think it'd make more sense to have this a its own library for now. And if we start seeing a lot of demand for it, we'll evaluate moving things into this repository, does that make sense?

baywet:
(closing as I believe we've reached a conclusion here)

PR #2852, opened 2026-05-25
feat(reader): remove ParseNode infrastructure

baywet:
## Summary
- remove the ParseNode wrapper infrastructure from reader parsing
- parse directly with System.Text.Json JsonNode/JsonObject APIs and explicit ParsingContext propagation
- prune unused JsonNode helper methods
- refresh benchmark report artifacts

## Benchmark Comparison vs Vincent Biret's Last Update (eb1258a6)

**Hardware difference:**
- **Vincent's run**: Intel i7-1185G7 3.00GHz (x64), .NET SDK 8.0.418, .NET 8.0.24 runtime
- **Current run**: Snapdragon X 12-core X1E80100 3.40 GHz (ARM64), .NET SDK 10.0.300, .NET 8.0.27 runtime

**Results comparison (Vincent → Current):**

| Benchmark | Time Change | Memory Change |
|-----------|-------------|---------------|
| **PetStoreYaml** | 405.2 μs → 276.3 μs **(32% faster)** | 361.39 KB → 305.91 KB **(15% less)** |
| **PetStoreJson** | 173.0 μs → 112.8 μs **(35% faster)** | 223.53 KB → 168.05 KB **(25% less)** |
| **GHESYaml** | 951,103.8 μs → 608,668.3 μs **(36% faster)** | 345343.16 KB → 250121.85 KB **(28% less)** |
| **GHESJson** | 435,425.1 μs → 244,147.6 μs **(44% faster)** | 206865.71 KB → 107293.42 KB **(48% less)** |

**Gen0/Gen1 GC collections:** Similar or slightly lower across all benchmarks.

**Key takeaway:** The refactored code shows **significant improvements** in both execution time (30-45% faster) and memory allocation (15-50% less). However, note this is **not an apples-to-apples comparison** due to different hardware (ARM vs x64) and runtime versions. For a true performance assessment, you'd need to rerun Vincent's baseline commit on the same Snapdragon X hardware.

## Validation
- dotnet build Microsoft.OpenApi.slnx --no-restore --verbosity minimal
- dotnet test Microsoft.OpenApi.slnx --no-build --verbosity minimal
- dotnet run -c Release (from performance\benchmark)


src/Microsoft.OpenApi/Models/BaseOpenApiReference.cs
gavinbarron, line 87:
```suggestion
                if (!string.IsNullOrEmpty(Id) &&
```

src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -17,106 +19,107 @@ internal static partial class OpenApiV2Deserializer
         private static readonly FixedFieldMap<OpenApiDocument> _openApiFixedFields = new()
         {
             {
-                "swagger", (_, _, _) => {}
+                "swagger", (_, _, _, c) => {}
```
gavinbarron: maybe this should be a discard as it's not used?

Same commend for the below functions where it's not being used

src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -17,106 +19,107 @@ internal static partial class OpenApiV2Deserializer
...
         };
 
-        private static void MakeServers(IList<OpenApiServer> servers, ParsingContext context, RootNode rootNode)
+        private static void MakeServers(IList<OpenApiServer> servers, ParsingContext context, JsonNode JsonNode)
         {
             var host = context.GetFromTempStorage<string>("host");
             var basePath = context.GetFromTempStorage<string>("basePath");
             var schemes = context.GetFromTempStorage<List<string>>("schemes");
-            var defaultUrl = rootNode.Context.BaseUrl;
+            var defaultUrl = context.BaseUrl;
```
gavinbarron, line 122: why we were going indirect to the context previously?

src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -133,7 +136,7 @@ private static void MakeServers(IList<OpenApiServer> servers, ParsingContext con
             //Validate host
             if (!string.IsNullOrEmpty(host) && !IsHostValid(host!))
             {
-                rootNode.Context.Diagnostic.Errors.Add(new(rootNode.Context.GetLocation(), "Invalid host"));
+                context.Diagnostic.Errors.Add(new(context.GetLocation(), "Invalid host"));
```
gavinbarron, line 139: Same question wrt direct vs indirect context access here.

src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
```diff
@@ -17,106 +19,107 @@ internal static partial class OpenApiV2Deserializer
...
 
         private static readonly PatternFieldMap<OpenApiDocument> _openApiPatternFields = new()
         {
             // We have no semantics to verify X- nodes, therefore treat them as just values.
-            {s => s.StartsWith(OpenApiConstants.ExtensionFieldNamePrefix, StringComparison.OrdinalIgnoreCase), (o, p, n, _) => o.AddExtension(p, LoadExtension(p, n))}
+            {s => s.StartsWith(OpenApiConstants.ExtensionFieldNamePrefix, StringComparison.OrdinalIgnoreCase), (o, p, n, _, c) => o.AddExtension(p, LoadExtension(p, n, c))}
         };
 
-        private static void MakeServers(IList<OpenApiServer> servers, ParsingContext context, RootNode rootNode)
+        private static void MakeServers(IList<OpenApiServer> servers, ParsingContext context, JsonNode JsonNode)
```
gavinbarron: JsonNode is unused in this method, does it need to be passed in?

If so then it should be camelCased

src/Microsoft.OpenApi/Reader/V2/OpenApiHeaderDeserializer.cs
```diff
@@ -16,11 +18,11 @@ internal static partial class OpenApiV2Deserializer
         {
             {
                 "description",
-                (o, n, _) => o.Description = n.GetScalarValue()
+                (o, n, _, c) => o.Description = n.GetScalarValue()
```
gavinbarron: should we use discard names in here too?

src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
```diff
@@ -209,10 +211,10 @@ private static void LoadStyle(OpenApiParameter p, string v)
             }
         }
 
-        private static void LoadParameterExamplesExtension(OpenApiParameter parameter, ParseNode node, OpenApiDocument? hostDocument)
+        private static void LoadParameterExamplesExtension(OpenApiParameter parameter, JsonNode node, OpenApiDocument? hostDocument, ParsingContext context)
```
gavinbarron, line 214: should we take the opportunity to clean up the unused hostDocument parameter here?

src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
```diff
@@ -223,22 +225,22 @@ private static OpenApiSchema GetOrCreateSchema(OpenApiParameter p)
             };
         }
 
-        private static void ProcessIn(OpenApiParameter o, ParseNode n, OpenApiDocument hostDocument)
+        private static void ProcessIn(OpenApiParameter o, JsonNode n, OpenApiDocument hostDocument, ParsingContext context)
```
gavinbarron, line 228: another unused hostDocument parameter

src/Microsoft.OpenApi/Reader/V2/OpenApiResponseDeserializer.cs
```diff
@@ -100,60 +102,60 @@ private static void ProcessProduces(MapNode mapNode, OpenApiResponse response, P
             context.SetTempStorage(TempStorageKeys.ResponseProducesSet, true, response);
         }
 
-        private static void LoadResponseExamplesExtension(OpenApiResponse response, ParseNode node, OpenApiDocument? hostDocument)
+        private static void LoadResponseExamplesExtension(OpenApiResponse response, JsonNode node, OpenApiDocument? hostDocument, ParsingContext context)
```
gavinbarron, line 105: usused hostDocument parameter

src/Microsoft.OpenApi/Reader/V2/OpenApiResponseDeserializer.cs
```diff
@@ -100,60 +102,60 @@ private static void ProcessProduces(MapNode mapNode, OpenApiResponse response, P
...
 
-            foreach (var examplesNode in mapNode)
+            foreach (var examplesNode in JsonObject)
             {
                 // Load the media type node as an OpenApiExample object
                 var example = new OpenApiExample();
-                var exampleNode = examplesNode.Value.CheckMapNode(examplesNode.Name);
-                foreach (var valueNode in exampleNode)
+                var exampleNode = examplesNode.Value.CheckMapNode(examplesNode.Key, context);
+                foreach (var JsonNode in exampleNode)
```
gavinbarron:
```suggestion
                foreach (var jsonNode in exampleNode)
```

src/Microsoft.OpenApi/Reader/V2/OpenApiSecuritySchemeDeserializer.cs
```diff
@@ -97,23 +99,20 @@ internal static partial class OpenApiV2Deserializer
...
-        public static IOpenApiSecurityScheme LoadSecurityScheme(ParseNode node, OpenApiDocument hostDocument)
+        public static IOpenApiSecurityScheme LoadSecurityScheme(JsonNode node, OpenApiDocument hostDocument, ParsingContext context)
         {
             // Reset the local variables every time this method is called.
             // TODO: Change _flow to a tempStorage variable to make the deserializer thread-safe.
             _flowValue = null;
             _flow = new();
 
-            var mapNode = node.CheckMapNode("securityScheme");
+            var JsonObject = node.CheckMapNode("securityScheme", context);
```
gavinbarron:
```suggestion
            var jsonObject = node.CheckMapNode("securityScheme", context);
```

src/Microsoft.OpenApi/Services/OpenApiFilterService.cs
```diff
@@ -142,26 +144,26 @@ public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, Fun
...
             {
-                rootNode.Attach(source.Value, source.Key);
+                JsonNode.Attach(source.Value, source.Key);
             }
-            return rootNode;
+            return JsonNode;
         }
 
-        private static Dictionary<HttpMethod, OpenApiOperation>? GetOpenApiOperations(OpenApiUrlTreeNode rootNode, string relativeUrl, string label)
+        private static Dictionary<HttpMethod, OpenApiOperation>? GetOpenApiOperations(OpenApiUrlTreeNode JsonNode, string relativeUrl, string label)
```
gavinbarron: All of the change in this method is GHCP needlessly renaming a parameter as best I can tell

src/Microsoft.OpenApi/Services/OpenApiFilterService.cs
```diff
@@ -142,26 +144,26 @@ public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, Fun
         /// <returns>The created <see cref="OpenApiUrlTreeNode"/>.</returns>
         public static OpenApiUrlTreeNode CreateOpenApiUrlTreeNode(Dictionary<string, OpenApiDocument> sources)
         {
-            var rootNode = OpenApiUrlTreeNode.Create();
+            var JsonNode = OpenApiUrlTreeNode.Create();
```
gavinbarron: this seems like a pointless change...

src/Microsoft.OpenApi/Services/OpenApiFilterService.cs
gavinbarron, line 1: Does this file even needs these changes?

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiHeaderTests.cs
```diff
@@ -1,4 +1,5 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+using System.Text.Json.Nodes;
```
gavinbarron: nit: shouldn't using statements follow after the copyright?

gavinbarron (commented):
most of the comments about dicards/casing and unused parameters apply broadly.

Although you have explained the unused parameters in a side chat, so disregard those comments

src/Microsoft.OpenApi/Models/BaseOpenApiReference.cs
baywet: I'm not sure I understand the suggestion here? is it a blank space suggestion? this line has not changed?

src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
baywet: the context was carried by the ParseNode (both custom types of this library). JsonNode (from system.text.json) does not carry the context, so it need to be a new parameter, etc... (We might be able to use extension properties from C# 14, but I don't know what the runtime and performance implications are, so I wanted to keep things simple)

src/Microsoft.OpenApi/Reader/V2/OpenApiSecuritySchemeDeserializer.cs
baywet: fixed the casing

src/Microsoft.OpenApi/Reader/V2/OpenApiHeaderDeserializer.cs
baywet: used the discard instead.

src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
baywet: fixed

src/Microsoft.OpenApi/Reader/V2/OpenApiResponseDeserializer.cs
baywet: fixed

src/Microsoft.OpenApi/Services/OpenApiFilterService.cs
baywet: fixed

src/Microsoft.OpenApi/Services/OpenApiFilterService.cs
baywet: reverted all changes in this file

test/Microsoft.OpenApi.Readers.Tests/V2Tests/OpenApiHeaderTests.cs
baywet: fixed

src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
baywet: we can't, it's to match the prototype. We'd need to introduce an intermediate lambda which has minimal binary size and memory negative impacts but also makes the code more convoluted.

src/Microsoft.OpenApi/Reader/V2/OpenApiParameterDeserializer.cs
baywet: we can't, it's to match the prototype. We'd need to introduce an intermediate lambda which has minimal binary size and memory negative impacts but also makes the code more convoluted.

src/Microsoft.OpenApi/Reader/V2/OpenApiResponseDeserializer.cs
baywet: we can't, it's to match the prototype. We'd need to introduce an intermediate lambda which has minimal binary size and memory negative impacts but also makes the code more convoluted.

src/Microsoft.OpenApi/Models/BaseOpenApiReference.cs
gavinbarron: It's an unchanged line so GitHub messes up the suggestion, the original line is
```csharp
                if (!string.IsNullOrEmpty(Id) && Id is not null &&
```

i.e. the null check already happened and the second part of the compound conditional is redundant

src/Microsoft.OpenApi/Reader/V2/OpenApiDocumentDeserializer.cs
gavinbarron: Fair enough, it just struck me as odd as the context was passed in the original form of the method and so I wanted to verify that we didn't need node relative context.

gavinbarron (approved):

src/Microsoft.OpenApi/Models/BaseOpenApiReference.cs
baywet: ahh, that's because we're also targeting netstandard2.0 and the standard library is not annotated with the nullability attributes in that version. This ends up confusing the compiler and leads to false warnings/errors at compile time.
The alternative being to use a null forgiving operator (!) but I refrain from using those as much as possible. This way they are easier to flag when they should NOT be used and might cause some issues when the code runs (because of missing checks/flawed logic).

PR #2854, opened 2026-05-26
fix(library): handle circular schema references

baywet:
## Summary
- Handle circular schema references without failing document loading.
- Add regression coverage for circular schema references.

## Testing
- dotnet test .\\test\\Microsoft.OpenApi.Readers.Tests\\Microsoft.OpenApi.Readers.Tests.csproj --no-restore --no-logo
- dotnet test .\\test\\Microsoft.OpenApi.Tests\\Microsoft.OpenApi.Tests.csproj --no-restore --no-logo

PR #2855, opened 2026-05-26
fix(library): handle circular schema references

baywet:
port of #2854 to v2


baywet:
@copilot refresh the benchmark reports, change nothing else

adrian05-ms (approved):

PR #2891, opened 2026-06-15
docs: fixes comment for Responses property

baywet:
context https://github.com/dotnet/aspnetcore/pull/67136#discussion_r3414791535

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiOperation.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiOperation.cs
@@ -75,7 +75,7 @@ namespace Microsoft.OpenApi
         public IOpenApiRequestBody? RequestBody { get; set; }
 
         /// <summary>
-        /// REQUIRED. The list of possible responses as they are returned from executing this operation.
+        /// The list of possible responses as they are returned from executing this operation.
         /// </summary>
         public OpenApiResponses? Responses { get; set; } = [];
 


src/Microsoft.OpenApi/Models/OpenApiOperation.cs
```diff
@@ -75,7 +75,7 @@ public ISet<OpenApiTagReference>? Tags
         public IOpenApiRequestBody? RequestBody { get; set; }
 
         /// <summary>
-        /// REQUIRED. The list of possible responses as they are returned from executing this operation.
+        /// The list of possible responses as they are returned from executing this operation.
         /// </summary>
         public OpenApiResponses? Responses { get; set; } = [];
```
Youssef1313, line 80: What about

https://github.com/microsoft/OpenAPI.NET/blob/24469a7216516789af227349e5bdbf5aceb63703/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L206

https://github.com/microsoft/OpenAPI.NET/blob/24469a7216516789af227349e5bdbf5aceb63703/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L319

Should they be using WriteOptionalObject instead?

baywet: Ahh, here is another clue, [it used to be required in v2](https://spec.openapis.org/oas/v2.0.html#fixed-fields-4), as well as [v3](https://spec.openapis.org/oas/v3.0.4.html#fixed-fields-7), but [was relaxed in 3.1](https://spec.openapis.org/oas/v3.1.2.html#fixed-fields-7)

baywet: And no, we shouldn't change this. The specification requires the property to be present with a non null value. WriteRequiredObject will write an empty object if the value is null. (`{}`)

MIchaelMainer (approved):

src/Microsoft.OpenApi/Models/OpenApiOperation.cs
Youssef1313: I'm extremely confused. If the specification requires the property to be present with a non null value, why are we making the C# property here nullable?

baywet: versions 2.0, and 3.0 required the property to be present, and the lib adds a placeholder empty object when it's null.
versions 3.1 and 3.2 do not require the property to be present.

Youssef1313 (approved):

Issue #2369, opened 2025-06-02
Add all "Annotations" to OpenApiSchemaReference


desjoerd:
Currently we can only set `description` as an annotation on an OpenApiSchemaReference. With OpenApi 3.1 and Json Schema 2020-12 it's allowed to apply all annotations on a reference.

Annotations: https://json-schema.org/understanding-json-schema/reference/annotations#annotations
Spec: https://json-schema.org/draft/2020-12/json-schema-core#section-7.7.1.1
Meta data schema: [json-schema.org/draft/2020-12/meta/meta-data](https://json-schema.org/draft/2020-12/meta/meta-data)

## Why

When mapping nested classes to OpenApi in ASP.NET Core it uses comments to apply annotations like title, description, examples. But this is currently limited to only apply a description for a reference. In the [example](https://json-schema.org/understanding-json-schema/reference/annotations#annotations
Spec: https://json-schema.org/draft/2020-12/json-schema-core#section-7.7.1.1) provided in the spec it would be awesome to be able to set different metadata on properties (which become references).

## Links
Related issue: https://github.com/dotnet/aspnetcore/issues/61965

baywet:
Hi @desjoerd 
Thank you for using the SDK and for reaching out.

Here are the annotations that are already supported today:
- summary
- description

Here is what I can see is missing from the spec/documentation [metadata](https://json-schema.org/draft/2020-12/meta/meta-data):
- default
- title
- deprecated
- readOnly
- writeOnly
- examples

Am I missing anything?

baywet:
Instructions for copilot:
1. Create a new `OpenApiSchemaReferenceInformation` that derives from `OpenApiReference`
2. In the `OpenApiSchemaReferenceInformation` type, add the following fields: default, title, deprecated, readOnly, writeOnly, examples
3. Add an additional generic parameter to `BaseOpenApiReferenceHolder` called U, with a constraint to derive from `OpenApiReference`
4. Update all existing usage of `BaseOpenApiReferenceHolder` to pass `OpenApiReference` as a value for that parameter
5. Update `BaseOpenApiReferenceHolder` Reference property to use that new generic parameter
6. Update the fields default, title, deprecated, readOnly, writeOnly, examples in  `OpenApiSchemaReference` to behave like the summary field
7. Update the serialization and deserialization code to account for the `OpenApiSchemaReferenceInformation` type and its additional fields
8. Add the required unit tests

desjoerd:
[...]

PR #2896, opened 2026-06-19
fix: preserve JSON Schema 2020-12 keyword siblings on $ref schemas for OAS 3.1+

### LINKED ISSUE #2895: Bug: OpenAPI 3.1 `$ref` siblings (`$defs`, `$dynamicAnchor`, `$id`) are silently dropped when parsing
## Describe the bug

When parsing an OpenAPI 3.1+ document, `OpenApiV31Deserializer.LoadSchema` short-circuits as soon as it sees a `$ref`: it returns an `OpenApiSchemaReference` and never runs `ParseMap` over the remaining sibling keywords. The only sibling data captured on that path is whatever `Reference.SetMetadataFromJsonObject` extracts, which (via `JsonSchemaReference.SetAdditional31MetadataFromMapNode`) is a hard-coded allow-list of annotations — `title`, `deprecated`, `readOnly`, `writeOnly`, `default`, `examples`, `extensions`. Every other sibling keyword — `$defs`, `$dynamicAnchor`, `$dynamicRef`, `$id`, `$anchor`, `$vocabulary`, `$comment`, and the rest of the JSON Schema 2020-12 vocabulary — is never parsed into the object model at all. The `OpenApiSchemaReference` accessors then delegate to `Target`, so callers observe `Target`'s values (or null) instead of the sibling declarations present in the source document.

The keywords in question are JSON Schema 2020-12 keywords. OpenAPI 3.0 does not recognize them, and — independently — the 3.0 spec requires `$ref` siblings to be ignored, so dropping them on the 3.0 path is correct. The bug is specific to the 3.1+ deserializer: 3.1 / JSON Schema 2020-12 makes `$ref` a regular keyword whose siblings are fully valid, yet the 3.1 parser still skips them when `$ref` is present.

Related prior issues:
- #2369 — "Add all 'Annotations' to OpenApiSchemaReference" (closed; addressed `description` and a subset of annotation siblings but not the JSON Schema 2020-12 keyword siblings below)
- #2763 — "OpenApiSchemaReference allows Description alongside `$ref` producing non-spec-compliant output in OAS 3.0" (separate concern about annotation-sibling behavior across 3.0 vs 3.1 — explicitly out of scope for this issue)

This issue is about JSON Schema 2020-12 keyword siblings being **dropped** when they should be **preserved**, on 3.1+ documents.

## OpenApi File To Reproduce

```yaml
openapi: 3.1.0
info:
  title: Sibling preservation repro
  version: 1.0.0
paths: {}
components:
  schemas:
    Target:
      type: object
      properties:
        name:
          type: string
    Referencing:
      $ref: '#/components/schemas/Target'
      description: Sibling description
      $dynamicAnchor: anchor
      $defs:
        sibling:
          $dynamicAnchor: inner
          $ref: '#/components/schemas/Target'
```

### Steps to reproduce

```csharp
using Microsoft.OpenApi;
using Microsoft.OpenApi.Reader;

var settings = new OpenApiReaderSettings();
settings.AddYamlReader();

var yaml = """
openapi: 3.1.0
info:
  title: Sibling preservation repro
  version: 1.0.0
paths: {}
components:
  schemas:
    Target:
      type: object
      properties:
        name:
          type: string
    Referencing:
      $ref: '#/components/schemas/Target'
      description: Sibling description
      $dynamicAnchor: anchor
      $defs:
        sibling:
          $dynamicAnchor: inner
          $ref: '#/components/schemas/Target'
""";

using var stream = new MemoryStream(System.Text.Encoding.UTF8.GetBytes(yaml));
var result = await OpenApiDocument.LoadAsync(stream, "yaml", settings);
var referencing = result.Document.Components!.Schemas["Referencing"];

Console.WriteLine($"Type:                  {referencing.GetType().Name}");
Console.WriteLine($"Is OpenApiSchemaReference: {referencing is OpenApiSchemaReference}");
Console.WriteLine($"Description (sibling): {referencing.Description ?? "(null)"}");
Console.WriteLine($"DynamicAnchor (sibling): {referencing.DynamicAnchor ?? "(null)"}");
Console.WriteLine($"Definitions count:     {referencing.Definitions?.Count ?? 0}");
if (referencing.Definitions != null)
    foreach (var kvp in referencing.Definitions)
        Console.WriteLine($"  $defs.{kvp.Key}: anchor={kvp.Value.DynamicAnchor ?? "(null)"}");
```

Verified against `Microsoft.OpenApi` `3.7.0` and `Microsoft.OpenApi.YamlReader` `3.7.0`.

## Expected behavior

Per OpenAPI 3.1 / JSON Schema 2020-12, `$ref` is no longer a special keyword that overrides siblings. All sibling keywords MUST be preserved on the referencing schema and accessible through the object model:

```
Type:                  OpenApiSchemaReference
Is OpenApiSchemaReference: True
Description (sibling): Sibling description
DynamicAnchor (sibling): anchor
Definitions count:     1
  $defs.sibling: anchor=inner
```

## Actual behavior

```
Type:                  OpenApiSchemaReference
Is OpenApiSchemaReference: True
Description (sibling): Sibling description   ← preserved (per #2369)
DynamicAnchor (sibling): (null)              ← LOST
Definitions count:     0                      ← LOST (delegates to Target.Definitions)
```

The sibling values aren't merely hidden by delegation — they were never parsed. `OpenApiV31Deserializer.LoadSchema` returns an `OpenApiSchemaReference` immediately on `$ref` (`Reader/V31/OpenApiSchemaDeserializer.cs:452`) and skips `ParseMap` entirely, so the fixed-field entries that recognize `$defs` / `$dynamicAnchor` / `$id` / `$anchor` / `$vocabulary` / `$comment` (lines 24-50) never execute. `OpenApiSchemaReference`'s accessors then delegate to `Target`, which is why `Target`'s (absent) values surface instead.

## Screenshots/Code Snippets

The deeper impact is observable on the JSON Schema generics pattern (Pattern B from the `$dynamicRef` ecosystem). The fixture below is the canonical "reusable paginated response" pattern, validator-backed by Hyperjump:

```yaml
components:
  schemas:
    User: { type: object, properties: { id: { type: string } } }
    PaginatedTemplate:
      $defs:
        itemType:
          $dynamicAnchor: itemType
          not: {}                                  # placeholder slot
      type: object
      properties:
        items:
          type: array
          items:
            $dynamicRef: '#itemType'              # template slot
    PaginatedUserResponse:
      $ref: '#/components/schemas/PaginatedTemplate'
      $defs:
        itemType:
          $dynamicAnchor: itemType
          $ref: '#/components/schemas/User'       # binding for User
```

After parsing:
- `PaginatedUserResponse` is an `OpenApiSchemaReference` with `Target = PaginatedTemplate`.
- `PaginatedUserResponse.Definitions` returns `PaginatedTemplate.Definitions` (the placeholder slot), not the User-bound sibling.
- There is no API path to reach the User-bound `$defs.itemType` entry. The dynamic-scope binding is unreachable.

This blocks Kiota (and any other downstream tool using `Microsoft.OpenApi`) from emitting `class PaginatedUserResponse : PaginatedTemplate<User>` from this fixture.

### Spec reference

- **OpenAPI 3.1**: "The OpenAPI Schema Object is a superset of the JSON Schema Specification Draft 2020-12." — https://spec.openapis.org/oas/v3.1.1#schema-object
- **JSON Schema 2020-12 §4.7**: `$ref` is a regular applicator keyword; siblings are no longer ignored. The 3.0 "When a `$ref` is used, all other properties SHALL be ignored" rule does not appear in 3.1.
- **JSON Schema 2020-12 §7.7**: `$dynamicAnchor` / `$dynamicRef` / `$defs` are first-class keywords that participate in dynamic scope resolution alongside `$ref`.

### Why this matters

The JSON Schema generics pattern (a template schema with `$dynamicRef` slots bound by sibling `$defs` in referencing schemas) is one of two primary real-world uses of `$dynamicRef`. It cannot be implemented on top of `Microsoft.OpenApi` 3.7.0 because the binding information is dropped at parse time.

A cross-generator compatibility matrix is tracked at https://github.com/aqeelat/openapi-dynamicref-adoption-tracker — multiple tools are affected by this pattern, and `Microsoft.OpenApi` is the upstream blocker for all .NET-based tools (Kiota today; potentially others).

### Environment

- `Microsoft.OpenApi` 3.7.0
- `Microsoft.OpenApi.YamlReader` 3.7.0
- .NET SDK 10.0.301
- Verified on macOS arm64; behavior is platform-independent

## Willingness to contribute a PR

I will open a PR soon.

---
*This issue was drafted with assistance from AI tooling. The submitter is responsible for reviewing and validating the contents before submission.*

aqeelat:
## Problem

`OpenApiV31Deserializer.LoadSchema` short-circuits on `$ref` before `ParseMap`, so JSON Schema 2020-12 keyword siblings (`$defs`, `$dynamicAnchor`, `$dynamicRef`, `$id`, `$schema`, `$anchor`, `$vocabulary`, `$comment`) were never parsed into the object model. This made Pattern B (generic template + `$dynamicRef` binding via sibling `$defs`) unimplementable for any tool built on `Microsoft.OpenApi`.

Fixes #2895.

## Solution

Mirrors the #2369 annotation-sibling pattern across four coordinated changes:

1. **Parser extraction** — `SetAdditional31MetadataFromMapNode` extracts scalar/collection siblings; `$defs` is parsed in `LoadSchema` (needs `LoadSchema` for nested schema materialization).
2. **Storage** — 8 new properties on `JsonSchemaReference` (`$id` stored as `SchemaId` to avoid collision with `BaseOpenApiReference.Id`).
3. **Accessor overrides** — `OpenApiSchemaReference` getters changed from `Target?.X` to `Reference.X ?? Target?.X`.
4. **Serialization** — `SerializeAdditionalV3XProperties` emits the new siblings alongside `$ref`.

Version-safe by call-site separation: `SetAdditional31MetadataFromMapNode` is only reachable from V31/V32 `LoadSchema`, never V3. All 1639 existing tests pass unchanged.

## Scope

Intentionally narrow — only JSON Schema 2020-12 structural keywords (`$`-prefixed). Application keywords (`type`, `properties`, `required`, etc.) still delegate to `Target`. Full sibling compliance is a separate concern with backward-compat implications.

Empty `$defs: {}` / `$vocabulary: {}` siblings do not suppress the target's values (collection assigned only when non-empty).

## Test coverage

12 new tests across V31 and V32:
- Parse: keyword siblings (`$dynamicAnchor` + `$defs`), scalar siblings (`$id`, `$schema`, `$comment`, `$anchor`, `$dynamicRef`), `$vocabulary`, `allOf`-binding variant
- Serialize round-trip: keyword siblings + scalar siblings


aqeelat:

src/Microsoft.OpenApi/Models/JsonSchemaReference.cs
```diff
@@ -76,24 +117,50 @@ public JsonSchemaReference(JsonSchemaReference reference) : base(reference)
...
+        writer.WriteProperty(OpenApiConstants.Comment, Comment);
+        writer.WriteOptionalMap(OpenApiConstants.Vocabulary, Vocabulary, (w, s) => w.WriteValue(s));
+        if (version == OpenApiSpecVersion.OpenApi3_1)
+        {
+            writer.WriteOptionalMap(OpenApiConstants.Defs, Definitions, (w, s) => s.SerializeAsV31(w));
+        }
+        else
+        {
+            writer.WriteOptionalMap(OpenApiConstants.Defs, Definitions, (w, s) => s.SerializeAsV32(w));
+        }
```
baywet: why do you need to use the version instead of the callback?
this is brittle and we might forget to update things here in future versions

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
```diff
@@ -452,6 +452,31 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
             var result = new OpenApiSchemaReference(reference.Item1, hostDocument, reference.Item2);
             result.Reference.SetMetadataFromJsonObject(jsonObject);
             result.Reference.SetJsonPointerPath(pointer, nodeLocation);
+
+            // Parse $defs sibling — requires LoadSchema for nested schema materialization,
+            // so it cannot be done inside SetAdditional31MetadataFromMapNode.
+            if (jsonObject.TryGetPropertyValue(OpenApiConstants.Defs, out var defsNode) && defsNode is JsonObject defsObj)
```
baywet: we could refactor that to accept a callback. Assuming the SetMetadata method is internal

test/Microsoft.OpenApi.Readers.Tests/V31Tests/OpenApiSchemaTests.cs
```diff
@@ -935,5 +935,412 @@ public void DeserializeFalseSchemaParsesAsNotEmptySchema()
...
+            var result = await OpenApiDocument.LoadAsync(stream, "yaml", SettingsFixture.ReaderSettings);
+            var referencing = result.Document.Components!.Schemas["Referencing"];
+
+            // Assert — siblings are preserved on the OpenApiSchemaReference
+            referencing.Should().BeOfType<OpenApiSchemaReference>();
+            referencing.Description.Should().Be("Sibling description");
+            referencing.DynamicAnchor.Should().Be("anchor");
+            referencing.Definitions.Should().NotBeNull();
+            referencing.Definitions!.Should().ContainKey("sibling");
+            referencing.Definitions["sibling"].DynamicAnchor.Should().Be("inner");
```
baywet: refactor to use the assert and not the fluent API. Long term we should get rid of this library.

src/Microsoft.OpenApi/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,17 @@
...
+Microsoft.OpenApi.JsonSchemaReference.Definitions.set -> void
+Microsoft.OpenApi.JsonSchemaReference.DynamicAnchor.get -> string?
+Microsoft.OpenApi.JsonSchemaReference.DynamicAnchor.set -> void
+Microsoft.OpenApi.JsonSchemaReference.DynamicRef.get -> string?
+Microsoft.OpenApi.JsonSchemaReference.DynamicRef.set -> void
+Microsoft.OpenApi.JsonSchemaReference.Schema.get -> System.Uri?
+Microsoft.OpenApi.JsonSchemaReference.Schema.set -> void
+Microsoft.OpenApi.JsonSchemaReference.SchemaId.get -> string?
+Microsoft.OpenApi.JsonSchemaReference.SchemaId.set -> void
+Microsoft.OpenApi.JsonSchemaReference.Vocabulary.get -> System.Collections.Generic.IDictionary<string!, bool>?
```
baywet, line 16: In JSON schema 2020-12 only annotation keywords are allowed next to $ref.
$anchor, $dynamicRef, $dynamicAnchor, $defs, $schema $id and $vocabulary are all core vocabulary keywords.
The only exception is $comment that's technically core vocabulary but allowed anyway.
The only fields that implement this behaviour (target vs reference value) are the annotation keywords (default, readOnly....)

baywet (changes requested):
Thanks for the contribution!
A couple of comments to help this move forward

src/Microsoft.OpenApi/PublicAPI.Unshipped.txt
baywet: Actually, after digging some more, it seems that:
- [$defs is allowed next two $ref for the whole thing to work](https://www.learnjsonschema.com/2020-12/core/dynamicref/)
- so is $id
- so is $schema

I couldn't find any additional evidence for $anchor, $dynamicAnchor or $vocabulary

baywet: actually I went ahead and had additional discussions with both @handrews and @darrelmiller (thank you both). My understanding was wrong.
Any JSON Schema keyword can appear next to $ref, some might not be applicable (not do anything) due to their nature, some might be contradictory when appearing both in the reference and the referenced schema, in which case, it's up to the application to define what the behaviour should be.

In the case where both are present, we have a precedent of returning the reference value for annotation keywords, so being consistent here makes sense. Then the application can check the reference and the target values, and compare them, if the difference is important to the application.

Any keyword in JSON schema exposed as a property in the [IOpenAPISchema interface](https://github.com/microsoft/OpenAPI.NET/blob/7345caac686dde49e97e36b2509f9fb674c825ca/src/Microsoft.OpenApi/Models/Interfaces/IOpenApiSchema.cs#L11) as well as in the [IOpenApiSchemaMissingProperties](https://github.com/microsoft/OpenAPI.NET/blob/7345caac686dde49e97e36b2509f9fb674c825ca/src/Microsoft.OpenApi/Models/Interfaces/IOpenApiSchemaMissingProperties.cs#L13) should have:
- a corresponding setter in the JsonSchemaReference type, and serialization code to match (use x-jsonschema prefix for OAI < 3.1)
- a corresponding getter in the JsonSchemaReference type, and the deserialization code to match (use x-jsonschema prefix for OAI < 3.1)
- a corresponding setter in OpenApiSchemaReference, that maps to the reference corresponding setter
- a corresponding getter, which returns the reference value when present, otherwise the target value
- cloning and unit test code like you've started

Now, I understand this is significant work, if you want to focus on the properties you've identified first, and punt the other properties to an additional issue, let me know.

src/Microsoft.OpenApi/Models/JsonSchemaReference.cs
aqeelat: Good catch. Done — replaced the version parameter with an `Action<IOpenApiWriter, IOpenApiSerializable>` callback.

src/Microsoft.OpenApi/Models/JsonSchemaReference.cs
```diff
@@ -76,24 +117,50 @@ public JsonSchemaReference(JsonSchemaReference reference) : base(reference)
...
+        writer.WriteProperty(OpenApiConstants.Anchor, Anchor);
+        writer.WriteProperty(OpenApiConstants.DynamicRef, DynamicRef);
+        writer.WriteProperty(OpenApiConstants.DynamicAnchor, DynamicAnchor);
+
         // Additional schema metadata annotations in 3.1
         writer.WriteOptionalObject(OpenApiConstants.Default, Default, (w, d) => w.WriteAny(d));
         writer.WriteProperty(OpenApiConstants.Title, Title);
```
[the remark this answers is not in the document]
aqeelat: Good catch. This is pre-existing — the hardcoded `OpenApi3_1` predates this PR. Will address in a follow-up.

test/Microsoft.OpenApi.Readers.Tests/V31Tests/OpenApiSchemaTests.cs
aqeelat: Good catch. Done — converted all new tests to `Assert.*`.

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
aqeelat: Good catch. Done — `$defs` parsing stays in the deserializer where `LoadSchema` is available, and the serialization version branching is replaced with a callback.

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
```diff
@@ -452,6 +452,31 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
+                    finally
+                    {
+                        context.EndObject();
+                    }
+                }
+                if (defs.Count > 0)
+                {
+                    result.Reference.Definitions = defs;
+                }
+            }
```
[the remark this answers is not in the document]
aqeelat: Good catch. Done — added `context.StartObject($defs)` before the loop, wrapped each entry in try/catch (`OpenApiReaderException` / `OpenApiException` → `context.Diagnostic.Errors`), mirroring the `ParseField` pattern.

src/Microsoft.OpenApi/Reader/V32/OpenApiSchemaDeserializer.cs
```diff
@@ -452,6 +452,31 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
+                    finally
+                    {
+                        context.EndObject();
+                    }
+                }
+                if (defs.Count > 0)
+                {
+                    result.Reference.Definitions = defs;
+                }
+            }
```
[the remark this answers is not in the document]
aqeelat: Good catch. Done — same fix applied to V32.

src/Microsoft.OpenApi/PublicAPI.Unshipped.txt
aqeelat: I'd like to focus on the current set of properties for this PR and file a follow-up issue for the remaining IOpenApiSchema / IOpenApiSchemaMissingProperties properties. Expanding to full sibling compliance (properties, required, type, etc.) is a larger change with backward-compat implications that warrants its own discussion. Let me know if that works.

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
baywet: Sorry, I wasn't specific enough here, pushed an update.

baywet (commented):

baywet (dismissed):

src/Microsoft.OpenApi/PublicAPI.Unshipped.txt
baywet: Opened #2903 as a follow up.

baywet (dismissed):

aqeelat:
@baywet Happy to help.
Can you give me a few minutes to test this against my kiota branch?

baywet:
@aqeelat sure thing! I'm running the performance tests to update the results anyway. Plus we now need a third approver since I made changes.

peombwa (approved):

aqeelat:
@baywet thank you for the fast turnaround. When do you think this will be released? Is it possible to get 3.7.1 today?

PR #2913, opened 2026-06-27
feat(schema): resolve bare $dynamicRef via $dynamicAnchor index

### LINKED ISSUE #2911: Feature: Resolve $dynamicRef against $dynamicAnchor in the schema reference resolution pipeline
## Background

JSON Schema 2020-12 defines `$dynamicRef` / `$dynamicAnchor` for recursive and extensible schema resources. This enables patterns where a base schema can recursively refer to the active derived schema:

```yaml
components:
  schemas:
    BaseCategory:
      $dynamicAnchor: category
      type: object
      properties:
        children:
          type: array
          items:
            $dynamicRef: '#category'
    LocalizedCategory:
      $dynamicAnchor: category
      allOf:
        - $ref: '#/components/schemas/BaseCategory'
```

When `LocalizedCategory` is the active schema, `children.items` should resolve to `LocalizedCategory`, not `BaseCategory`.

## Current State

After #2896, JSON Schema 2020-12 keywords such as `$dynamicRef`, `$dynamicAnchor`, and `$defs` are parsed, preserved, and serialized, including as siblings on `$ref` schemas.

However, `$dynamicRef` is not currently part of schema reference resolution:

- `BaseOpenApiReferenceHolder.Target` resolves `$ref` through `OpenApiDocument.ResolveReference`.
- `OpenApiWorkspace` indexes schemas by component path and `$id`, but not by `$dynamicAnchor`.
- `JsonNodeHelper.GetReferencePointer` only checks `$ref`.
- A schema with `$dynamicRef` but no `$ref` is deserialized as a plain `OpenApiSchema` with `DynamicRef` set, so it does not enter the `OpenApiSchemaReference.Target` resolution path.

## Design Questions

There are two separate decisions to make.

### 1. How should bare `$dynamicRef` schemas participate in resolution?

Options:

1. Treat `$dynamicRef` similarly to `$ref` during schema loading, producing a schema reference type whose target resolution uses `$dynamicAnchor`.
2. Keep `$dynamicRef` as metadata on `OpenApiSchema`, but add a separate dynamic-ref resolution path outside `Target`.

Option 1 integrates with the existing reference model but requires `Target` to resolve context-dependently. Option 2 keeps `$dynamicRef` resolution explicit and doesn't conflate it with static `$ref` resolution.

### 2. What scope should resolution support?

A document-level `$dynamicAnchor` lookup handles the simple case where only one schema declares an anchor.

Full JSON Schema 2020-12 behavior requires dynamic-scope resolution: when multiple resources declare the same `$dynamicAnchor`, the outermost active schema resource wins. That requires context that the current `Target` / `Workspace.ResolveReference` pipeline does not carry today.

| Aspect | Document-scoped lookup | Dynamic-scope resolution |
|--------|---------------------------|---------------------------|
| Simple recursion (single anchor) | Works | Works |
| Generic types (multiple anchors, same name) | **Fails** — no context | Works |
| Implementation size | Small (index + lookup) | Large (scope plumbing + walker changes) |
| Risk to existing `Target` semantics | Low (additive) | Medium-high (new resolution path) |
| Precedent in repo | `$id` registration pattern | `LoopDetector`, `ResolveRecursiveTarget` |
| Order-dependency | None | Inherent (documented in kiota) |
| Solves the pure-`$dynamicRef` blocker? | Needs deserializer change | Needs deserializer change |

## Related

- #2896
- #2895
- #2903
- microsoft/kiota#7815
- microsoft/kiota#7817

aqeelat:
# Pull Request

## Description

Implements document-scoped `$dynamicRef` resolution per JSON Schema 2020-12 §8.2.3.2. Bare `$dynamicRef` schemas (no `$ref`) now deserialize as `OpenApiSchemaReference` whose `Target` resolves via per-document `$dynamicAnchor` and `$anchor` registries in `OpenApiWorkspace`.

## Resolution order in `Target`

1. `$dynamicAnchor` index — single candidate resolves automatically
2. `$anchor` fallback — when zero `$dynamicAnchor` candidates exist (per §8.2.3.2)
3. `null` — when ambiguous (multiple candidates need dynamic-scope tracking)

## Type of Change

## Public APIs for consumers tracking dynamic scope

- `GetDynamicAnchorCandidates(doc, anchorName)` — returns all candidate schemas
- `ResolveDynamicAnchorInContext(contextSchema, anchorName)` — resolves against a specific schema's `$defs`

## `$ref` sibling behavior

This PR relies on the existing OpenAPI.NET 3.1/3.2 behavior where `$ref` schema siblings are preserved on `JsonSchemaReference` and surfaced through `OpenApiSchemaReference`'s existing reference-first property getters.

This PR does not introduce that behavior. It extends the same storage model to bare `$dynamicRef` schemas so sibling keywords are preserved while the schema participates in reference resolution.

See #2919 for a broader discussion on whether the reference-first getter model should be revised for assertion keywords.

## Anchor index coverage

Registries are populated by recursively walking the entire document tree: component schemas, reusable component definitions (parameters, responses, request bodies, headers, callbacks, path items, media types), inline schemas (paths, operations, webhooks), and all nested subschema locations (`$defs`, `properties`, `items`, `allOf`, `if/then/else`, etc.).

## Open question

When multiple schemas declare the same `$dynamicAnchor`, `Target` returns `null` (ambiguous). The spec requires the outermost candidate in the dynamic evaluation scope. Automatic resolution would require threading evaluation context through `Target` (e.g., `AsyncLocal<Stack>`). Currently consumers handle this via `GetDynamicAnchorCandidates` + `ResolveDynamicAnchorInContext`. Feedback welcome on whether the library should handle this automatically or leave it to consumers.

## Testing

- [x] All existing tests pass (except 7 pre-existing `Serialize_DoesNotMutateDom` failures on clean main)

46 dynamic-ref tests across V31 and V32 covering: bare `$dynamicRef` deserialization, resolution to anchors in all schema locations, ambiguity handling, `$anchor` fallback, `$dynamicAnchor` precedence, round-trip serialization, sibling preservation, context-aware resolution, OM-built document, and `$ref` regression.

`Microsoft.OpenApi.Tests`: 1149 passed. `Microsoft.OpenApi.Readers.Tests`: 535 passed, 7 pre-existing failures.

## Checklist

## Versions applicability

## Related

- Fixes #2911
- Builds on #2896 (preserve JSON Schema 2020-12 keyword siblings on `$ref` schemas)
- microsoft/kiota#7817 tracks the consumer-side use case


aqeelat:
@baywet sorry for the mega PR. Most of the changes are tests and mechanical changes. However, if you want, I can split it into multiple PRs if that will make it easier for you to review.

baywet:
Thanks for the contribution!

No worries, I'll take the time to review during the week.

Looking back at the reference mechanisms that are implemented throughout the library, I realized we had a specification/documentation page that never got published on release. Took the time to clean it up a little and it's now available here.

https://learn.microsoft.com/en-us/openapi/openapi.net/references-openapi

Let us know what you think!

src/Microsoft.OpenApi/Models/JsonSchemaReference.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.OpenApi;
 /// Schema reference information that includes metadata annotations from JSON Schema 2020-12.
 /// This class extends OpenApiReference to provide schema-specific metadata override capabilities.
 /// </summary>
-public class JsonSchemaReference : OpenApiReferenceWithDescription
+public class JsonSchemaReference : OpenApiReferenceWithDescription, IOpenApiSchemaMissingProperties
```
baywet: what was the motivation for adding this interface?
I think this is going to add confusion, essentially, only OpenApiSchemaReference and OpenApiSchema should provide that service to the consumers. The only reason why this class is public and not internal is because it's a generic type argument for another public class.

test/Microsoft.OpenApi.Readers.Tests/V31Tests/OpenApiDynamicRefTests.cs
```diff
@@ -0,0 +1,1270 @@
...
+
+        var doc = await LoadDocumentAsync(yaml);
+
+        var tree = doc.Components.Schemas["Tree"];
+        var next = tree.Properties["next"];
+        var reference = Assert.IsType<OpenApiSchemaReference>(next);
+        Assert.True(reference.Reference.IsDynamicRefOnly);
+        // External target is not loaded in this workspace, so resolution returns null rather than
+        // falling back to the local Tree anchor.
+        Assert.Null(reference.Target);
```
baywet, line 585: could we update this unit test to provide a second document and ensure the resolution of externals also works?

baywet (commented):
Thanks for the contribution!

In addition to the comments I've left, I think we have good coverage of the deserialization scenarios, and some coverage of the serialization scenarios (at least for basic properties).
In think that one key aspect that is missing from the tests is: people using the object model to build documents. For "regular refs" components registration is required. I'd like to see more tests demonstrating that a document built from OM can resolve dynamic refs without requiring a re-parsing.

src/Microsoft.OpenApi/Models/JsonSchemaReference.cs
aqeelat: Reverted. I added it to share an `EnumerateMissingPropertiesChildren` helper between the two `EnumerateChildren` overloads in `OpenApiWorkspace`, but you're right that it adds confusion. Restored the separate overload that reads `JsonSchemaReference` properties directly.

test/Microsoft.OpenApi.Readers.Tests/V31Tests/OpenApiDynamicRefTests.cs
aqeelat: Good point. External `$dynamicRef` resolution (cross-document via URI) requires finding the target document by URI and looking up its anchor registry. The current `Target` override only resolves fragment-only refs (`#node`) against the local document. I can add cross-document support in this PR if you want it now, or track it as a follow-up. Let me know which you prefer.

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
```diff
@@ -467,6 +468,25 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
+                    if (!string.Equals(name, OpenApiConstants.DynamicRef, StringComparison.Ordinal))
+                    {
+                        schema.UnrecognizedKeywords ??= new Dictionary<string, JsonNode>(StringComparer.Ordinal);
+                        schema.UnrecognizedKeywords[name] = value;
+                    }
+                });
+            result.Reference.ApplySchemaMetadata(referenceMetadata, jsonObject);
+            result.Reference.IsDynamicRefOnly = true;
+            return result;
+        }
```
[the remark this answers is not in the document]
aqeelat: Fixed. Added `result.Reference.SetJsonPointerPath(dynamicPointer, nodeLocation)` before returning, mirroring the `$ref` branch.

src/Microsoft.OpenApi/Reader/V32/OpenApiSchemaDeserializer.cs
```diff
@@ -467,6 +468,25 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
+                    if (!string.Equals(name, OpenApiConstants.DynamicRef, StringComparison.Ordinal))
+                    {
+                        schema.UnrecognizedKeywords ??= new Dictionary<string, JsonNode>(StringComparer.Ordinal);
+                        schema.UnrecognizedKeywords[name] = value;
+                    }
+                });
+            result.Reference.ApplySchemaMetadata(referenceMetadata, jsonObject);
+            result.Reference.IsDynamicRefOnly = true;
+            return result;
+        }
```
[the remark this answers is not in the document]
aqeelat: Fixed. Same change applied to V32.

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
```diff
@@ -288,6 +446,189 @@ internal bool RegisterComponent<T>(string location, T component)
...
+        /// <param name="hostDocument">The document whose anchors to search.</param>
+        /// <param name="anchorName">The bare anchor name (without leading <c>#</c>).</param>
+        /// <returns>All candidate schemas, or an empty list if none.</returns>
+        public IReadOnlyList<IOpenApiSchema> GetDynamicAnchorCandidates(OpenApiDocument hostDocument, string anchorName)
+        {
+            if (_dynamicAnchorRegistryByDocument.TryGetValue(hostDocument, out var anchors) &&
+                anchors.TryGetValue(anchorName, out var candidates))
+                return candidates;
+            return [];
+        }
```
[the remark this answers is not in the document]
aqeelat: Fixed. Now returns `candidates.AsReadOnly()`.

test/Microsoft.OpenApi.Readers.Tests/V31Tests/OpenApiDynamicRefTests.cs
```diff
@@ -0,0 +1,1270 @@
...
+        Assert.NotNull(reference.Target);
+        Assert.Same(doc.Components.Schemas["Foo"], reference.Target);
+    }
+
+    [Fact]
+    public async Task DynamicRefWithSiblingsPreservesSiblings()
+    {
+        // A $dynamicRef alongside structural schema keywords must not drop the siblings. The object
+        // is parsed as a normal OpenApiSchema (preserving maxProperties and properties) rather than
+        // being reduced to a bare reference that loses them.
```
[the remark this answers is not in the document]
aqeelat: Fixed. Updated comment to reflect that the object is parsed as an `OpenApiSchemaReference` with siblings preserved via `ApplySchemaMetadata`.

src/Microsoft.OpenApi/Reader/JsonNodeHelper.cs
```diff
@@ -167,6 +167,39 @@
...
+        }
+
+        /// <summary>
+        /// Extracts the bare anchor name from a $dynamicRef value.
+        /// Handles fragment-only (#meta), absolute-URI (https://example.com#meta), and bare (meta) forms.
+        /// Returns null for null/empty input. Returns empty string for bare "#" (root reference).
+        /// </summary>
+        public static string? ExtractDynamicAnchorName(string? dynamicRef)
+        {
+            if (string.IsNullOrEmpty(dynamicRef) || dynamicRef is null) return null;
```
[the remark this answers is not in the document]
aqeelat: Fixed. Removed the redundant `|| dynamicRef is null` and used `dynamicRef!` on the subsequent dereference. The redundant check was a workaround for netstandard2.0 nullable flow analysis; the null-forgiving operator achieves the same without the dead conditional.

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
```diff
@@ -213,6 +217,160 @@
...
+        /// </summary>
+        private void RegisterComponentAnchors(OpenApiDocument document)
+        {
+            var components = document.Components;
+            if (components is null) return;
+
+            if (components.Parameters is not null)
+                foreach (var parameter in components.Parameters.Values)
+                    if (parameter is not null)
+                        RegisterParameterAnchors(document, parameter);
```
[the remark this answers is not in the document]
aqeelat: Acknowledged. These `foreach + if (x is not null)` patterns match the existing `RegisterComponents` code style throughout the file. Converting to `.Where()` would add LINQ iterator allocations in a one-time registration path for no runtime benefit. Leaving as-is for consistency with the existing codebase.

aqeelat:
@baywet CodeQL is suggesting we convert the `foreach + if (x is not null)` patterns in `OpenApiWorkspace` to `.Where()` and `FirstOrDefault`. The trade-off: `.Where()` reads cleaner but allocates an iterator per call site; `foreach + if` iterates directly with zero allocations. Negligible in a one-time registration path like `RegisterComponents`. Should we convert for readability or keep as-is?

baywet:
> @baywet CodeQL is suggesting we convert the `foreach + if (x is not null)` patterns in `OpenApiWorkspace` to `.Where()` and `FirstOrDefault`. The trade-off: `.Where()` reads cleaner but allocates an iterator per call site; `foreach + if` iterates directly with zero allocations. Negligible in a one-time registration path like `RegisterComponents`. Should we convert for readability or keep as-is?

Let's convert for consistency (we have patterns like those throughout the codebase). Allocation wise, the costliest part is the DOM itself and associated strings, the iterators are short lived and negligeable in comparison. And we've made significant performance improvements over the past year or so to have a better margin here.

test/Microsoft.OpenApi.Readers.Tests/V31Tests/OpenApiDynamicRefTests.cs
baywet: please implement that as part of this pull request so we have good coverage.

aqeelat: Implemented. `Target` now resolves URI-based `$dynamicRef` values by finding the target document via `FindDocumentByBaseUri` in the workspace and looking up the anchor in that document's registry. Added `AbsoluteDynamicRefResolvesAcrossDocuments` test with two documents in one workspace verifying cross-document resolution. Also kept `AbsoluteDynamicRefResolvesToLocalAnchorWhenExternalNotLoaded` to verify null is returned when the external document isn't loaded.

aqeelat:
Done. Converted all `foreach + if (x is not null)` patterns to `.Where()` throughout `OpenApiWorkspace.cs`, and replaced the `ResolveDynamicAnchorInContext` loop with `FirstOrDefault`. Pushed in the latest force-push.

All review comments should be addressed now. Do you think this can be merged today?

aqeelat:
@copilot fix failing test and address the other review items

baywet:
> copilot fix failing test and address the other review items

@aqueelat, I believe this doesn't work on forks. (I had the same problem with aspnetcore recently)

test/Microsoft.OpenApi.Readers.Tests/V32Tests/OpenApiDocumentSerializationTests .cs
```diff
@@ -28,6 +28,7 @@ public async Task Serialize_DoesNotMutateDom(OpenApiSpecVersion version)
             // Act: Serialize using System.Text.Json
             var options = new JsonSerializerOptions
             {
+                ReferenceHandler = System.Text.Json.Serialization.ReferenceHandler.Preserve,
```
baywet: I'm curious about why we've had to turn this on?

src/Microsoft.OpenApi/Models/JsonSchemaReference.cs
```diff
@@ -335,6 +335,13 @@ public class JsonSchemaReference : OpenApiReferenceWithDescription
     /// </summary>
     public IOpenApiSchema? Else { get; set; }
 
+    /// <summary>
+    /// Indicates whether this reference was created from a bare $dynamicRef (no $ref).
+    /// When true, serialization emits $dynamicRef instead of $ref, and Target resolution
+    /// uses the $dynamicAnchor index rather than the $ref URI lookup.
+    /// </summary>
+    internal bool IsDynamicRefOnly { get; set; }
```
baywet: The problem I can see with this flag is if someone initializes the reference from the object model (so not deserializing an existing document), it's not being set. So they'll end up with a "broken" document. We need to account for that from the object model constructor/setters. We don't want to make that flag public.
So either the flag need to be calculated from the getter (no field/setter) or the object model needs to set it in that scenario, and we need unit tests to demonstrate that.

test/Microsoft.OpenApi.Readers.Tests/V32Tests/OpenApiDocumentSerializationTests .cs
aqeelat: The sample documents (`docWith31properties.json`, `docWith32properties.json`) contain recursive `$dynamicRef`:

```json
{
  "$dynamicAnchor": "addressDef",
  "properties": {
    "address": { "$dynamicRef": "#addressDef" }
  }
}
```

After this PR, bare `$dynamicRef` deserializes as `OpenApiSchemaReference`. Its convenience getters (`Properties`, `Items`, etc.) delegate to `Target` per the existing `Reference.X ?? Target?.X` contract. For recursive refs, `Target` resolves to the ancestor schema, creating a cycle that STJ's default serializer can't handle (max depth 64).

`ReferenceHandler.Preserve` lets STJ handle cycles via `$id`/`$ref` metadata. The test still detects mutations because both snapshots (before/after native serialization) use the same format.

This is a direct consequence of making bare `$dynamicRef` participate in reference resolution as `OpenApiSchemaReference` (#2911).

baywet: ah thanks for the context! But doesn't that insert $id and $ref that we don't want in the resulting output?

src/Microsoft.OpenApi/Models/JsonSchemaReference.cs
aqeelat: Good point. I replaced the field/setter with a computed getter:

```csharp
internal bool IsDynamicRefOnly
    => !string.IsNullOrEmpty(DynamicRef)
       && ReferenceV3 is string refV3
       && !refV3.StartsWith("#/components/", StringComparison.OrdinalIgnoreCase);
```

The deserializer calls `SetJsonPointerPath` with the raw `$dynamicRef` value (e.g. `#node`), which sets `ReferenceV3` to a non-component path. So `IsDynamicRefOnly` is now derived from existing state — no flag to forget.

For object model construction, users need to call `SetJsonPointerPath` after setting `DynamicRef`. I updated `DynamicRefResolvesInObjectModelBuiltDocument` to demonstrate this. Changes are in the latest uncommitted diff — pushing shortly.

test/Microsoft.OpenApi.Readers.Tests/V32Tests/OpenApiDocumentSerializationTests .cs
aqeelat: You're right — `ReferenceHandler.Preserve` adds `$id`/`$ref` artifacts that don't belong. Replaced it with a custom `JsonConverter<IOpenApiSchema>` that routes schema serialization through the native OpenAPI writer, avoiding reflection on the computed getters that create the cycle.

Root cause: the sample documents contain recursive `$dynamicRef` (e.g. `$dynamicAnchor: addressDef` with `$dynamicRef: #addressDef` pointing back to the parent). After this PR, bare `$dynamicRef` deserializes as `OpenApiSchemaReference` whose convenience getters delegate to `Target`, creating a cycle that STJ's default serializer can't handle.

Options considered:
1. `[JsonIgnore]` on `Target` — cycle is through ~45 computed getters, not `Target` directly
2. Existing `OpenApiSchemaJsonConverter` — typed as `JsonConverter<OpenApiSchema>`, not `IOpenApiSchema`. Changing the type parameter breaks existing tests because STJ doesn't route concrete types through interface-typed converters at the root level.
3. `MaxDepth = 256` — infinite cycle, no depth limit helps
4. `ReferenceHandler.Preserve` — works but pollutes output with `$id`/`$ref`
5. Custom `JsonConverter<IOpenApiSchema>` — routes schemas through native writer, no artifacts, non-schema properties still compared via STJ reflection

Went with #5. The converter is shared between V31/V32 test files:
```csharp
internal sealed class OpenApiSchemaInterfaceJsonConverter(
    Action<IOpenApiWriter, IOpenApiSchema> serialize) : JsonConverter<IOpenApiSchema>
{
    public override IOpenApiSchema Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        => throw new NotSupportedException();

    public override void Write(Utf8JsonWriter writer, IOpenApiSchema value, JsonSerializerOptions options)
    {
        using var tw = new StringWriter();
        var ow = new OpenApiJsonWriter(tw);
        serialize(ow, value);
        writer.WriteRawValue(tw.ToString(), skipInputValidation: true);
    }
}
```

Registered as `new OpenApiSchemaInterfaceJsonConverter(static (w, s) => s.SerializeAsV31(w))`.

baywet:
@aqeelat merging #2918 means this one is conflicting now. Would you mind solving the conflicts before I give it another look please?

aqeelat:
@baywet Conflicts resolved — rebased on `origin/main`. 

One observation from the merge: `RegisterSchemaIdentifiers` (#2918) and `RegisterAnchors` (this PR) both recursively walk the same schema tree for each component schema — once to register `$id`/`$anchor` as component URIs, once to register `$dynamicAnchor`/`$anchor` in the anchor registry. The walks visit the same nodes but serve different registries with different reference-handling strategies (`RegisterSchemaIdentifiers` skips references; `RegisterAnchors` reads from the `JsonSchemaReference` directly).

I left them as separate calls for now. Merging into one pass would require:
- Adding a `document` parameter to all `RegisterSchemaIdentifiers` overloads
- Adding reference-specific anchor registration before the existing `OpenApiSchemaReference` skip
- Adding reference-specific child enumeration via a separate `EnumerateChildren(JsonSchemaReference)` overload (since `JsonSchemaReference` is not `IOpenApiSchema`)

Happy to do that if you prefer a single walk, or leave as-is since the double walk is O(2n) and doesn't show up in benchmarks.

baywet:
@aqeelat thanks for handling the conflicts, and for pointing this out. Let's merge the crawling, it *should* reduce the amount of code we have and improve the performance.

aqeelat:
@baywet can we do it in a different PR?

src/Microsoft.OpenApi/Models/JsonSchemaReference.cs
```diff
@@ -335,6 +335,17 @@ public class JsonSchemaReference : OpenApiReferenceWithDescription
...
+    /// <summary>
+    /// Indicates whether this reference represents a bare $dynamicRef (no $ref to a component).
+    /// When true, serialization emits $dynamicRef instead of $ref, and Target resolution
+    /// uses the $dynamicAnchor index rather than the $ref URI lookup.
+    /// Computed from whether $dynamicRef is set and ReferenceV3 does not point to a component path.
+    /// </summary>
+    internal bool IsDynamicRefOnly
+        => !string.IsNullOrEmpty(DynamicRef)
+           && ReferenceV3 is string refV3
+           && !refV3.StartsWith("#/components/", StringComparison.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
aqeelat: Not actionable. Per JSON Schema 2020-12, `$dynamicAnchor` names MUST start with a letter and contain only `[A-Za-z0-9-_]` — they cannot contain `/`. So `$dynamicRef: "#/components/schemas/Foo"` has a fragment that can never match a `$dynamicAnchor`. The spec (§8.2.3.2) says `$dynamicRef` falls back to `$ref` behavior when no matching `$dynamicAnchor` is found. The computed getter correctly returns `false` for this case, treating it as a regular `$ref`, which matches the spec's fallback behavior.

src/Microsoft.OpenApi/Services/OpenApiWorkspace.cs
```diff
@@ -359,6 +509,208 @@ private void RegisterSchemaIdentifier(IOpenApiSchema? schema, Uri baseUri, ISet<
...
+        public static IOpenApiSchema? ResolveDynamicAnchorInContext(IOpenApiSchema? contextSchema, string anchorName)
+        {
+            if (contextSchema is null || string.IsNullOrEmpty(anchorName)) return null;
+
+            if (contextSchema.DynamicAnchor is string a && a.Equals(anchorName, StringComparison.Ordinal))
+                return contextSchema;
+
+            return contextSchema.Definitions?.Values.FirstOrDefault(def =>
+                def.DynamicAnchor is string da && da.Equals(anchorName, StringComparison.Ordinal));
+        }
```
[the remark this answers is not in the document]
aqeelat: Fixed in f506268c. `ResolveDynamicAnchorInContext` now special-cases `OpenApiSchemaReference` to read from `Reference.DynamicAnchor`/`Reference.Definitions` directly, without falling through to Target. Added test `ResolveDynamicAnchorInContextDoesNotReadFromTargetForReferenceHolders` that verifies a bare `$ref` without authored `$defs` does not resolve anchors from the Target schema.

src/Microsoft.OpenApi/Models/References/OpenApiSchemaReference.cs
```diff
@@ -35,6 +35,51 @@ private OpenApiSchemaReference(OpenApiSchemaReference schema) : base(schema)
...
+                        && hostDocument.Workspace is { } workspace)
+                    {
+                        // Determine which document to resolve against
+                        OpenApiDocument? targetDoc = hostDocument;
+                        if (!Microsoft.OpenApi.Reader.JsonNodeHelper.IsFragmentOnlyDynamicRef(Reference.DynamicRef))
+                        {
+                            // URI-based ref: find the external document
+                            var docUri = Microsoft.OpenApi.Reader.JsonNodeHelper.ExtractDocumentUri(Reference.DynamicRef);
+                            targetDoc = docUri is not null ? workspace.FindDocumentByBaseUri(docUri) : null;
+                        }
```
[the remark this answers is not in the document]
aqeelat: Known limitation, tracked as #2928. Relative URI resolution in `$dynamicRef` is not yet implemented — `ExtractDocumentUri` returns the raw string and `FindDocumentByBaseUri` requires `UriKind.Absolute`. The fix requires resolving relative URIs against the host document's `BaseUri` before calling `FindDocumentByBaseUri`. Out of scope for this PR.

aqeelat:
@baywet I forgot to mention that I already did the merged crawl in [`77c976c`](https://github.com/microsoft/OpenAPI.NET/pull/2913/commits/77c976c3c32b3125ade9d876c7d494d6dd598a73)

I have another PR to work on to address https://github.com/microsoft/OpenAPI.NET/pull/2913#discussion_r3547356433. I'll stack it on top of this one.

The kiota changes are ready locally but require we release a new version with this PR.

baywet (dismissed):

aqeelat:
@baywet I just pushed the copilot review fixes.

I also noticed there are a couple of tests not mirrored between V31 and V32. I will include them in the next PR.

baywet (approved):

PR #2915, opened 2026-06-29
feat: add JsonConverter for OpenApiSchema System.Text.Json serialization

### LINKED ISSUE #2299: OpenApiSchema System.Text.Json (De)Serialization
Default System.Text.Json of OpenApiJsonSchema is verbose and unsatisfying.

A simple way to fix it, probably via a JsonConverter, either by default or to configure on AddMVC json options.

Otherwise, OpenApiSchema could support direct serialization to a JsonWriter:

**Additional context**
Turns out I did it in my projects, and I also just really wanted to know if this was appropriate, or I missed something available
```c#

internal class SchemaConverter : System.Text.Json.Serialization.JsonConverter<OpenApiSchema>
{
	private static readonly ConcurrentDictionary<int, JsonSerializerOptions> _jsonSerializerCache = new();
	public override OpenApiSchema Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
	{
		var correctedOptions = _jsonSerializerCache.GetOrAdd(options.GetHashCode(), _ =>
		{
			var optionsCopy = new JsonSerializerOptions(options);
			optionsCopy.Converters.Remove(options.Converters.First(c => c is SchemaConverter));
			return optionsCopy;
		});

		return JsonSerializer.Deserialize<OpenApiSchema>(ref reader, correctedOptions)!;
	}

	public override void Write(Utf8JsonWriter writer, OpenApiSchema value, JsonSerializerOptions options)
	{
		using var memoryStream = new MemoryStream();
		using (var textWriter = new StreamWriter(memoryStream, leaveOpen: true))
		{
			value.SerializeAsV31(new OpenApiJsonWriter(textWriter));
			textWriter.Flush();
		}

		memoryStream.Position = 0;
		writer.WriteRawValue(memoryStream.ToArray());
	}
}


```

Mahdigln:
## Description
Adds `OpenApiSchemaJsonConverter`, a `System.Text.Json.JsonConverter<OpenApiSchema>` that produces clean OpenAPI wire format output instead of the default verbose reflection-based serialization.

## Type of Change

## Related Issue(s)
Closes #2299

## Changes Made
- Added `OpenApiSchemaJsonConverter` in `src/Microsoft.OpenApi/Converters/`
- Supports all OpenAPI versions: 2.0, 3.0, 3.1 (default), 3.2
- Added `Read` (deserialization) support in addition to `Write` (serialization)
- Updated `PublicAPI.Unshipped.txt` with new public API surface

## Testing
- [x] All existing tests pass

## Checklist

## Versions applicability
- [x] My change applies to the version 2.X of the library, if so PR link:
- [x] My change applies to the version 3.X of the library, if so PR link:

## Additional Notes
Usage example:
```csharp
var options = new JsonSerializerOptions();
options.Converters.Add(new OpenApiSchemaJsonConverter());

// Serialize
var json = JsonSerializer.Serialize(schema, options);

// Deserialize
var schema = JsonSerializer.Deserialize<OpenApiSchema>(json, options);

// Target a specific OpenAPI version
options.Converters.Add(new OpenApiSchemaJsonConverter(OpenApiSpecVersion.OpenApi3_0));

### CODE REVIEW

File: test/Microsoft.OpenApi.Tests/Converters/OpenApiSchemaJsonConverterTests.cs [hunk 1]
```diff
@@ -0,0 +1,147 @@
...
+            {
+                Type = JsonSchemaType.String,
+                Description = "A simple string"
+            };
+
+            var json = JsonSerializer.Serialize(schema, _optionsV31);
+
+            using var doc = JsonDocument.Parse(json);
+            doc.RootElement.GetProperty("type").GetString().Should().Be("string");
+            doc.RootElement.GetProperty("description").GetString().Should().Be("A simple string");
```
baywet: let's use the assert API. I thought we had gotten rid of FluentAssertions?

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// </code>
+    /// </remarks>
+    public sealed class OpenApiSchemaJsonConverter : JsonConverter<OpenApiSchema>
+    {
+        private readonly OpenApiSpecVersion _version;
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="OpenApiSchemaJsonConverter"/> targeting OpenAPI 3.1.
+        /// </summary>
+        public OpenApiSchemaJsonConverter() : this(OpenApiSpecVersion.OpenApi3_1) { }
```
baywet: The default should be 3.2 like it is for serialization helpers. (maybe all those should rely on a common constant?)

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// Deserializes a bare JSON Schema object into an <see cref="OpenApiSchema"/> by temporarily
+        /// embedding it in a minimal OpenAPI 3.1 document for parsing.
+        /// </remarks>
+        public override OpenApiSchema? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            using var document = JsonDocument.ParseValue(ref reader);
+            var schemaJson = document.RootElement.GetRawText();
+
+            var wrapper = string.Concat(
+                "{\"openapi\":\"3.1.0\",\"info\":{\"title\":\"temp\",\"version\":\"0.0.0\"},",
```
baywet: this is going to break down based on the version that's passed in the constructor.
The templates should either be version specific or the constructor should guard against versions that are not supported.

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// </summary>
+    /// <remarks>
+    /// <para>Register this converter via <see cref="JsonSerializerOptions.Converters"/>:</para>
+    /// <code>
+    /// var options = new JsonSerializerOptions();
+    /// options.Converters.Add(new OpenApiSchemaJsonConverter());
+    /// var json = JsonSerializer.Serialize(schema, options);
+    /// </code>
+    /// </remarks>
+    public sealed class OpenApiSchemaJsonConverter : JsonConverter<OpenApiSchema>
```
baywet, line 26: in theory this could be expanded to any model from this library (operations, path items, etc)

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+        public override OpenApiSchema? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            using var document = JsonDocument.ParseValue(ref reader);
+            var schemaJson = document.RootElement.GetRawText();
+
+            var wrapper = string.Concat(
+                "{\"openapi\":\"3.1.0\",\"info\":{\"title\":\"temp\",\"version\":\"0.0.0\"},",
+                "\"components\":{\"schemas\":{\"schema\":", schemaJson, "}}}");
+
+            var result = OpenApiDocument.Parse(wrapper);
```
baywet: why not use the model factory instead of creating an empty document?

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+            IOpenApiSchema? schema = null;
+            result.Document?.Components?.Schemas?.TryGetValue("schema", out schema);
```
baywet: This might break if any reference is used, the unit tests should at least account for this.

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        /// <inheritdoc/>
+        /// <remarks>
+        /// Deserializes a bare JSON Schema object into an <see cref="OpenApiSchema"/> by temporarily
+        /// embedding it in a minimal OpenAPI 3.1 document for parsing.
+        /// </remarks>
+        public override OpenApiSchema? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            using var document = JsonDocument.ParseValue(ref reader);
+            var schemaJson = document.RootElement.GetRawText();
```
baywet: If this ends up in a hot path, it'll lead to a lot of string allocations

baywet (commented):
Thanks for the contribution!

Let's a couple of comments to help get this merged

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
Mahdigln: Good point! Generalizing to other models like operations and path items would be a great follow-up PR. Keeping this focused on OpenApiSchema for now since that's the use case from the issue

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
Mahdigln: Acknowledged. Using ArrayBufferWriter<byte> instead of MemoryStream + StreamWriter would reduce allocations. Happy to address this as a follow-up if needed

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
baywet: sure! this easily could be refactored without introducing breaking changes. Resolving

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
baywet: I don't think this was fundamentally addressed. We should at least have unit tests to prevent regressions in the references scenarios.

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
baywet: I think we could even pass the JsonNode directly to the factory if we do get a JsonNode.

src/Microsoft.OpenApi/Converters/OpenApiSchemaJsonConverter.cs
```diff
@@ -52,18 +53,9 @@ public OpenApiSchemaJsonConverter(OpenApiSpecVersion version)
...
-            using var stream = new MemoryStream(schemaBytes);
-            var schema = OpenApiModelFactory.Load<OpenApiSchema>(
-                stream,
-                _version,
-                OpenApiConstants.Json,
-                new OpenApiDocument(),
-                out _);
-
-            return schema;
+            var jsonNode = JsonNode.Parse(ref reader);
```
baywet: let's throw an explicit exception instead of using null forgiving here. If the json node is null, something went wrong and we should provide the feedback to the end user.

baywet (approved):

peombwa (approved):

PR #2918, opened 2026-06-29
feat/external anchors and relative ids

baywet:
related https://github.com/MicrosoftDocs/openapi-docs/pull/167 and #2913


baywet:
@copilot update the performance reports

test/Microsoft.OpenApi.Readers.Tests/V32Tests/RelativeReferenceTests.cs
```diff
@@ -172,6 +172,123 @@ public async Task ParseExternalComponentSubschemaJsonSchemaReferenceWorks()
...
+paths:
+  /person/{id}:
+    get:
+      responses:
+        '200':
+          description: ok
+          content:
+            application/json:
+              schema:
+                $ref: 'OAS-schemas.yaml#address'
```
Youssef1313, line 198: Is this equivalent to `OAS-schemas.yaml#/components/schemas/person/address`?

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -639,7 +639,11 @@ private static string ConvertByteArrayToString(byte[] hash)
                 string relativePath;
                 var referenceV3 = !string.IsNullOrEmpty(reference.ReferenceV3) ? reference.ReferenceV3! : string.Empty;
 
-                if (!string.IsNullOrEmpty(referenceV3) && IsSubComponent(referenceV3))
+                if (reference.Type is ReferenceType.Schema && useExternal && TryGetPlainNameFragment(referenceV3, out var plainNameFragment))
```
Youssef1313, line 642: `TryGetPlainNameFragment` returns true only when the `#` is not followed by `/`.

Could we still be "external" document but have the `/`? How does that work?

test/Microsoft.OpenApi.Readers.Tests/V32Tests/RelativeReferenceTests.cs
baywet: in this specific context, yes

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
baywet: I'm not sure I'm following, could you give examples of what you believe works, doesn't, and should work?

Youssef1313: At this line, if `referenceV3` is `file.yaml#/a/b/c`, I think `TryGetPlainNameFragment` will be false. Is that expected?

This is essentially a question of the impact of the `!parts[1].StartsWith('/')` check in `TryGetPlainNameFragment`

baywet: > If present, the value of this keyword MUST be a string and MUST start with a letter ([A-Za-z]) or underscore ("_"), followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), and periods (".")

https://json-schema.org/draft/2020-12/json-schema-core#name-defining-location-independe

Since the $anchor keyword value CAN ONLY contain those characters, the fragment value MUST match for the resolution to work. So it CAN'T contain /. The check is here to differentiate between an anchor based resolution or  a JSON pointer based one (block below).

Hopefully that makes sense?

Youssef1313 (approved):
I'm not familiar on everything but giving a sign-off (thought the approval won't be counting anyways as I'm not a maintainer in this repo)

peombwa (approved):

PR #2920, opened 2026-07-01
Fix handling of nullable enums for 3.0

Youssef1313:
Previously, this scenario wasn't producing a `type` and wasn't producing a null value in the enum array.

This fixes it so that they are produced.

In the future, we could also avoid adding a `oneOf` with a single element in this scenario.

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -554,7 +554,7 @@ namespace Microsoft.OpenApi
                     var clonedToMutateEnum = element.CreateShallowCopy();
                     if (clonedToMutateEnum is OpenApiSchema { Enum: { } existingEnum } concreteCloned)
                     {
-                        concreteCloned.Enum = [.. existingEnum, null!];
+                        concreteCloned.Enum = [.. existingEnum, JsonNullSentinel.JsonNull];
                         callback(writer, clonedToMutateEnum);
                     }
                     else
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -1100,23 +1100,21 @@ namespace Microsoft.OpenApi
                     }
                     else if (schema.Enum is { Count: > 0 })
                     {
-                        foreach (var enumValue in schema.Enum)
+                        foreach (var enumValue in schema.Enum.Where(x => x is not null))
                         {
-                            if (enumValue is not null)
+                            var currentType = enumValue.GetValueKind() switch
                             {
-                                var currentType = enumValue.GetValueKind() switch
-                                {
-                                    JsonValueKind.Array => JsonSchemaType.Array,
-                                    JsonValueKind.String => JsonSchemaType.String,
-                                    JsonValueKind.Number => JsonSchemaType.Number,
-                                    JsonValueKind.True or JsonValueKind.False => JsonSchemaType.Boolean,
-                                    JsonValueKind.Null => (JsonSchemaType)0,
-                                    _ => JsonSchemaType.Object,
-                                };
+                                JsonValueKind.Array => JsonSchemaType.Array,
+                                JsonValueKind.String => JsonSchemaType.String,
+                                JsonValueKind.Number => JsonSchemaType.Number,
+                                JsonValueKind.True or JsonValueKind.False => JsonSchemaType.Boolean,
+                                JsonValueKind.Null => (JsonSchemaType)0,
+                                _ => JsonSchemaType.Object,
+                            };
 
-                                commonType |= currentType;
-                            }
+                            commonType |= currentType;
                         }
+
                         commonType |= JsonSchemaType.String;
                     }
                 }
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -521,6 +521,7 @@ namespace Microsoft.OpenApi
             IList<IOpenApiSchema>? effectiveOneOf = OneOf;
             IList<IOpenApiSchema>? effectiveAnyOf = AnyOf;
             bool hasNullInComposition = false;
+            bool hasOneOfNullAndSingleEnumWith3_0 = false;
             JsonSchemaType? inferredType = null;
 
             if (version == OpenApiSpecVersion.OpenApi3_0)
@@ -531,6 +532,9 @@ namespace Microsoft.OpenApi
                 (effectiveAnyOf, var inferredAnyOf, var nullInAnyOf) = ProcessCompositionForNull(AnyOf);
                 hasNullInComposition |= nullInAnyOf;
                 inferredType = inferredAnyOf ?? inferredType;
+
+                hasOneOfNullAndSingleEnumWith3_0 = nullInOneOf && effectiveOneOf is { Count: 1 } &&
+                    effectiveOneOf[0].Enum is { Count: > 0 };
             }
 
             // type
@@ -543,7 +547,21 @@ namespace Microsoft.OpenApi
             writer.WriteOptionalCollection(OpenApiConstants.AnyOf, effectiveAnyOf, callback);
 
             // oneOf
-            writer.WriteOptionalCollection(OpenApiConstants.OneOf, effectiveOneOf, callback);
+            if (hasOneOfNullAndSingleEnumWith3_0 &&
+                effectiveOneOf![0] is OpenApiSchema { Enum.Count: > 0 } singleEffectiveOneOf)
+            {
+                writer.WriteRequiredCollection(OpenApiConstants.OneOf, effectiveOneOf, (writer, element) =>
+                {
+                    var clonedToMutateEnum = (OpenApiSchema)((OpenApiSchema)element).MemberwiseClone();
+                    clonedToMutateEnum.Enum = [.. clonedToMutateEnum.Enum!, null!];
+                    callback(writer, clonedToMutateEnum);
+                });
+            }
+            else
+            {
+                writer.WriteOptionalCollection(OpenApiConstants.OneOf, effectiveOneOf, callback);
+            }
+
 
             // not
             writer.WriteOptionalObject(OpenApiConstants.Not, Not, callback);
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -531,6 +532,9 @@ private void SerializeInternal(IOpenApiWriter writer, OpenApiSpecVersion version
                 (effectiveAnyOf, var inferredAnyOf, var nullInAnyOf) = ProcessCompositionForNull(AnyOf);
                 hasNullInComposition |= nullInAnyOf;
                 inferredType = inferredAnyOf ?? inferredType;
+
+                hasOneOfNullAndSingleEnumWith3_0 = nullInOneOf && effectiveOneOf is { Count: 1 } &&
+                    effectiveOneOf[0].Enum is { Count: > 0 };
```
DeagleGross, line 537: should this be `count > 1` because enum values will have `null` and at least one enum value? Maybe I am wrong

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -543,7 +547,21 @@ private void SerializeInternal(IOpenApiWriter writer, OpenApiSpecVersion version
             writer.WriteOptionalCollection(OpenApiConstants.AnyOf, effectiveAnyOf, callback);
 
             // oneOf
-            writer.WriteOptionalCollection(OpenApiConstants.OneOf, effectiveOneOf, callback);
+            if (hasOneOfNullAndSingleEnumWith3_0 &&
+                effectiveOneOf![0] is OpenApiSchema { Enum.Count: > 0 } singleEffectiveOneOf)
+            {
+                writer.WriteRequiredCollection(OpenApiConstants.OneOf, effectiveOneOf, (writer, element) =>
+                {
+                    var clonedToMutateEnum = (OpenApiSchema)((OpenApiSchema)element).MemberwiseClone();
```
DeagleGross: Not familiar with the code here - but can we attempt to avoid doing clone + rewrite here for perf reasons? Can u please try if that is possible to inject null before or not?

DeagleGross (approved):

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: It's valid (and more correct) to not have `null` in the enum array. We are dealing with something like:

```json
"oneOf": [
  {
    "type": "null"
  },
  {
    "enum": [
      // ...
    ]
  }
]
```

The enum array could have a single element, and in all reasonable practical use cases, it must **not** have null, because then null becomes invalid as it matches both child schemas.

Note that the above shape is draft 2020-12, which isn't what we emit for OpenAPI 3.0. However, the OpenApiSchema **model** is designed more towards the 2020-12 draft.

That's why we can't emit the thing as-is and we have to do lots of special casing for versions. So, we emit the exact shape above that I mentioned for 3.1 and 3.2 which use 2020-12. But the same OpenApiSchema model serializes differently for 3.0 because it's not really using 2020-12 draft and the above shape isn't valid for OpenAPI 3.0.

That's at least how I'm understanding things currently.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: I didn't want to mutate objects that are provided by users as it's risky. I think the json serialization cost is likely much higher than this clone.

Youssef1313: Mutating it also will technically lead to wrong OpenApiSchema, this is mostly an internal hack during serialization only and for 3.0 only.

In theory, users could use the same OpenApiSchema and serialize it to 3.0 and use the same instance to serialize it later to 3.1.

If we mutated it, we will break it for users when they serialize for the newer version.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -1070,10 +1088,17 @@ private static (IList<IOpenApiSchema>? effective, JsonSchemaType? inferredType,
...
                 {
-                    commonType |= schema.Type.GetValueOrDefault() & ~JsonSchemaType.Null;
+                    if (schema.Type.HasValue)
+                    {
+                        commonType |= schema.Type.Value & ~JsonSchemaType.Null;
+                    }
+                    else if (schema.Enum is { Count: > 0 })
+                    {
+                        commonType |= JsonSchemaType.String;
+                    }
```
baywet, line 1119: enum members are not necessarily strings, they can be a discrete set of numbers as well, or boolean. Constraining to string here will lead to impossible to validate schemas

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
baywet: this cast is unsafe, it might be a reference. Use type assertions instead

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1853,6 +1856,80 @@ public void DeserializeContainsExtensionsInV3AssignsContainsProperties()
...
+        }
+
+        [Theory]
+        [InlineData(OpenApiSpecVersion.OpenApi3_1)]
+        [InlineData(OpenApiSpecVersion.OpenApi3_2)]
+        public async Task SerializeNullableEnumWith3_1_And_Later(OpenApiSpecVersion version)
+        {
+            var schema = CreateNullableEnumSchema();
+            var result = await schema.SerializeAsJsonAsync(version);
+            Assert.Equal("""
```
baywet: use JSON node deep equals instead in case properties ordering shifts around we don't get noise.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1853,6 +1856,80 @@ public void DeserializeContainsExtensionsInV3AssignsContainsProperties()
...
+                    JsonValue.Create("B")
+                }
+            });
+            return schema;
+        }
+
+        private enum MyEnum
+        {
+            A, B
+        }
```
baywet: is this used anywhere?

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1853,6 +1856,80 @@ public void DeserializeContainsExtensionsInV3AssignsContainsProperties()
...
+            Assert.Equal("""
+                {
+                  "type": "string",
+                  "oneOf": [
+                    {
+                      "enum": [
+                        "A",
+                        "B",
+                        null
+                      ]
```
baywet, line 1880: in theory the enum definition can be moved up (no one of). But the more permutations we do, the more we risk loosing other keywords in the process.

baywet (changes requested):

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: Good point.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: It should be safe. There is already a type check above `effectiveOneOf![0] is OpenApiSchema`.

We are dealing here with a case where we have a collection of single element, and we already checked that the single element is OpenApiSchema.

I'll refactor it to be easier to read and avoid MemberwiseClone anyways as we have already a clone constructor.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
Youssef1313: Nope, removing.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
Youssef1313: Yup, I noted that in the PR description as a possibility to do in the future.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: Fixed.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: Refactored so that it is more straightforward to see and understand.

baywet (dismissed):

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
baywet: valid yes. More correct, not really. Each keyword is evaluated independently, the instance needs to validate one of the keywords. And enums are effectively oneOf consts. We could argue it's easier to ready though.

Youssef1313: @baywet

What I mean is something like:

```
"oneOf": [
  {
    "type": "null"
  },
  {
    "enum": [
      "A",
      "B",
      null
    ]
  }
]
```

In this case, null matches both child schemas. And so, oneOf fails and `null` won't be considered valid.

See https://www.jsonschemavalidator.net/s/bdAKMMTl

So my point is that a schema like the above is pointless.

Youssef1313:
@baywet Is there anything remaining to get this merged? Would we have a new version released before .NET 11 Preview 7 so that we try to get that in aspnetcore repo and possibly build some fixes on our side on top of this?

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
baywet: ah thanks for the additional context here. Yes I didn't think of "both are valid, so oneOf will fail". Resolving

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -543,7 +547,27 @@ private void SerializeInternal(IOpenApiWriter writer, OpenApiSpecVersion version
...
             // oneOf
-            writer.WriteOptionalCollection(OpenApiConstants.OneOf, effectiveOneOf, callback);
+            if (hasOneOfNullAndSingleEnumWith3_0)
+            {
+                writer.WriteRequiredCollection(OpenApiConstants.OneOf, effectiveOneOf!, (writer, element) =>
+                {                    
+                    var clonedToMutateEnum = element.CreateShallowCopy();
+                    if (clonedToMutateEnum is OpenApiSchema { Enum: { } existingEnum } concreteCloned)
+                    {
+                        concreteCloned.Enum = [.. existingEnum, null!];
```
baywet:
```suggestion
                        concreteCloned.Enum = [.. existingEnum, JsonNullSentinel.JsonNull];
```

baywet (approved):

Issue #2903, opened 2026-06-26
Add missing keywords support for OpenAPIReference (and JsonSchemaRefernece


baywet:
actually I went ahead and had additional discussions with both @handrews and @darrelmiller (thank you both). My understanding was wrong.
Any JSON Schema keyword can appear next to $ref, some might not be applicable (not do anything) due to their nature, some might be contradictory when appearing both in the reference and the referenced schema, in which case, it's up to the application to define what the behaviour should be.

In the case where both are present, we have a precedent of returning the reference value for annotation keywords, so being consistent here makes sense. Then the application can check the reference and the target values, and compare them, if the difference is important to the application.

Any keyword in JSON schema exposed as a property in the [IOpenAPISchema interface](https://github.com/microsoft/OpenAPI.NET/blob/7345caac686dde49e97e36b2509f9fb674c825ca/src/Microsoft.OpenApi/Models/Interfaces/IOpenApiSchema.cs#L11) as well as in the [IOpenApiSchemaMissingProperties](https://github.com/microsoft/OpenAPI.NET/blob/7345caac686dde49e97e36b2509f9fb674c825ca/src/Microsoft.OpenApi/Models/Interfaces/IOpenApiSchemaMissingProperties.cs#L13) should have:
- a corresponding setter in the JsonSchemaReference type, and serialization code to match (use x-jsonschema prefix for OAI < 3.1)
- a corresponding getter in the JsonSchemaReference type, and the deserialization code to match (use x-jsonschema prefix for OAI < 3.1)
- a corresponding setter in OpenApiSchemaReference, that maps to the reference corresponding setter
- a corresponding getter, which returns the reference value when present, otherwise the target value
- cloning and unit test code like you've started

Now, I understand this is significant work, if you want to focus on the properties you've identified first, and punt the other properties to an additional issue, let me know.

_Originally posted by @baywet in https://github.com/microsoft/OpenAPI.NET/pull/2896#discussion_r3454882031_

PR #2921, opened 2026-07-02
fix(schema): apply additive JSON Schema semantics to $ref sibling getters

### LINKED ISSUE #2919: Clarify or revise OpenApiSchemaReference semantics for $ref sibling keywords
## Problem

OpenAPI.NET currently preserves OpenAPI 3.1/3.2 `$ref` schema siblings on `JsonSchemaReference` and exposes them through `OpenApiSchemaReference` convenience getters using reference-first fallback behavior:

```csharp
Reference.X ?? Target?.X
```

This behavior appears to be intentional. In #2903, @baywet wrote after discussions with @handrews and @darrelmiller that:

> Any JSON Schema keyword can appear next to `$ref` ... some might be contradictory when appearing both in the reference and the referenced schema, in which case, it's up to the application to define what the behaviour should be.

And:

> In the case where both are present, we have a precedent of returning the reference value for annotation keywords, so being consistent here makes sense. Then the application can check the reference and the target values, and compare them, if the difference is important to the application.

That explains why #2906 made the schema keyword getters reference-first across all keyword types.

The concern is that this convenience behavior can be easy to misread for JSON Schema assertion keywords.

In JSON Schema 2020-12, `$ref` siblings are evaluated additively. They are not overrides of the referenced schema.

Example:

```yaml
components:
  schemas:
    Target:
      type: object
      minProperties: 5
    Referencing:
      $ref: '#/components/schemas/Target'
      minProperties: 2
```

Both constraints apply. The sibling `minProperties: 2` does not relax the target's `minProperties: 5`.

With the current object model:

```csharp
schemaRef.MinProperties;           // 2
schemaRef.Reference.MinProperties; // 2
schemaRef.Target.MinProperties;    // 5
```

This can make assertion keywords appear to override target constraints, even though JSON Schema evaluation treats them compositionally.

## Question

Should `OpenApiSchemaReference` getters continue to expose reference-first convenience behavior for all schema keywords, or should the library distinguish more explicitly between:

- authored sibling keywords on `JsonSchemaReference`
- resolved target schema values via `Target`
- an evaluation-oriented/composed schema view

### Option 1: Keep current behavior and document it

Keep `Reference.X ?? Target?.X` for all schema keyword getters.

Document that these getters are object-model convenience accessors, not a JSON Schema evaluation/effective-schema view. Consumers that care about conflicts or additive constraints should inspect both `schemaRef.Reference` and `schemaRef.Target`.

### Option 2: Split annotation and assertion getter behavior

Keep reference-first getters for annotation-style keywords such as:

- `Description`
- `Title`
- `Default`
- `Deprecated`
- `ReadOnly`
- `WriteOnly`
- `Examples`
- `Extensions`

Return target values for assertion/structural keywords such as:

- `Type`
- `Format`
- `MinProperties`
- `MaxProperties`
- `Pattern`
- `Required`
- `Properties`
- `AllOf`
- `OneOf`
- `AnyOf`
- `Not`
- `Items`
- `If`
- `Then`
- `Else`

This is more validation-oriented, but it breaks the uniform getter rule from #2903 and creates setter/getter asymmetry. For example, setting `schemaRef.Type` stores the sibling value on `Reference`, but reading `schemaRef.Type` would return the target value.

### Option 3: Keep current getters and add an explicit evaluation/composition API later

Keep the current object-model behavior for compatibility and consistency.

Add a future API such as `GetEffectiveSchema()` or similar for consumers that need JSON Schema evaluation semantics. That API could preserve additive sibling assertions compositionally, for example by representing inlined references with `allOf` rather than flattening sibling assertions as overrides.

## Related Concern: Inline Serialization

`CopyReferenceAsTargetElementWithOverrides()` currently creates an inline `OpenApiSchema` from the reference view. Because the reference view is reference-first, inline serialization can flatten target-plus-sibling state in a way that looks like sibling assertions override target assertions.

A semantically safer inline representation would preserve composition, for example:

```yaml
allOf:
  - <inlined target schema>
minProperties: 2
```

This may be better handled separately from the getter behavior.

## Related

- #2913 extends the existing sibling-preservation model to bare `$dynamicRef` schemas.
- #2911 tracks `$dynamicRef` / `$dynamicAnchor` resolution.

aqeelat:
## Summary

Following the discussion in #2919 with @handrews and @baywet, this PR is now documentation-only.

## Changes

Adds XML documentation to `OpenApiSchemaReference` clarifying that convenience getters return authored sibling keyword values from `Reference` before falling back to resolved `Target` values. These getters are object-model conveniences and do not represent JSON Schema evaluation semantics.

## Rationale

The reference-first getter behavior (`Reference.X ?? Target?.X`) was an explicit design decision (#2903). Per #2919:
- Assertion keywords alongside `$ref` are additive in JSON Schema evaluation, not overrides
- The library does not perform validation — splitting getter behavior per keyword type creates setter/getter asymmetry and keyword-specific edge cases
- The safest direction is to keep consistent reference-first getters and document the contract explicitly

Consumers that need precise JSON Schema semantics should inspect both `Reference` and `Target`, or use a JSON Schema validator.

Fixes #2919

### DIFF
--- a/src/Microsoft.OpenApi/Models/References/OpenApiSchemaReference.cs
+++ b/src/Microsoft.OpenApi/Models/References/OpenApiSchemaReference.cs
@@ -9,7 +9,11 @@ namespace Microsoft.OpenApi
 {
 #pragma warning disable CS0618
     /// <summary>
-    /// Schema reference object
+    /// Schema reference object.
+    /// Convenience getters return <c>$ref</c>-sibling keyword values authored on
+    /// <see cref="IOpenApiReferenceHolder{V}.Reference"/> before falling back to resolved values from <see cref="Target"/>.
+    /// These getters are object-model conveniences and do not represent JSON Schema
+    /// evaluation semantics.
     /// </summary>
     public class OpenApiSchemaReference : BaseOpenApiReferenceHolder<OpenApiSchema, IOpenApiSchema, JsonSchemaReference>, IOpenApiSchema, IOpenApiSchemaMissingProperties, IOpenApiSchemaWithUnevaluatedProperties, IOpenApiExtensible
     {


handrews:
I'm not entirely sure the `additionalProperties` and `unevaluatedProperties` behaviors are correct.  Both of those have behavior that is highly dependent on their exact schema location.  In particular, `additionalProperties` can _only_ be evaluated in terms of its immediate adjacent keywords in the same schema object.  And for `unevaluatedProperties`, the behavior of placing in in the `$ref` source vs target is very different.  Based purely on the comments (I haven't had time to try to understand the code- this is not an ecosystem with which I am familiar) this also seems to treat both as booleans, when either could also be a non-trivial schema of some sort.

baywet (dismissed):

baywet (approved):

PR #2933, opened 2026-07-09
Handle nullability more accurately

Youssef1313:
My previous PR was trying to handle nullability for enums specifically. This PR is more generalized to handle nulls more correctly as it was still buggy for non-enums.

In addition, the previous fix for enums didn't also fix all enum scenarios. It only fixed it when we are given "oneOf (null, schemaWithEnum)". But if we are given a schema reference instead, we are not going to add "null" to the reference and we will produce the wrong thing.

I'll leave comments on some specific updated tests to explain.

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -514,28 +514,17 @@ namespace Microsoft.OpenApi
                 : Enum;
             writer.WriteOptionalCollection(OpenApiConstants.Enum, enumValue, (nodeWriter, s) => nodeWriter.WriteAny(s));
 
-            // Handle oneOf/anyOf with null type for v3.0 downcast
-            bool hasNullInComposition = false;
-            JsonSchemaType? inferredType = null;
-
             if (version == OpenApiSpecVersion.OpenApi3_0)
             {
-                (var inferredOneOf, var nullInOneOf) = ProcessCompositionForNull(OneOf);
-                hasNullInComposition |= nullInOneOf;
-                inferredType = inferredOneOf ?? inferredType;
-                (var inferredAnyOf, var nullInAnyOf) = ProcessCompositionForNull(AnyOf);
-                hasNullInComposition |= nullInAnyOf;
-                inferredType = inferredAnyOf ?? inferredType;
-
                 // If we have a schema that's only just { "type": "null" }, we serialize it as enum with null value.
-                if (Type == JsonSchemaType.Null && OneOf is not { Count: > 0 } && AnyOf is not { Count: > 0 })
+                if (Type == JsonSchemaType.Null && OneOf is not { Count: > 0 } && AnyOf is not { Count: > 0 } && Enum is not { Count: > 0 })
                 {
                     writer.WriteOptionalCollection(OpenApiConstants.Enum, s_singleNullElementList, (nodeWriter, s) => nodeWriter.WriteAny(s));
                 }
             }
 
             // type
-            var serializedTypeProperty = TrySerializeTypeProperty(writer, version, inferredType);
+            var serializedTypeProperty = TrySerializeTypeProperty(writer, version);
 
             // allOf
             writer.WriteOptionalCollection(OpenApiConstants.AllOf, AllOf, callback);
@@ -587,7 +576,7 @@ namespace Microsoft.OpenApi
                 //
                 // If the user explicitly set IsNullable to true, we serialize it even if redundant.
                 // But if **we** are inferring it (from oneOf/anyOf), we don't serialize it when it's redundant.
-                SerializeNullable(writer, version, hasNullInComposition);
+                SerializeNullable(writer, version);
             }
 
             // discriminator
@@ -1037,9 +1026,9 @@ namespace Microsoft.OpenApi
             }
         }
 
-        private void SerializeNullable(IOpenApiWriter writer, OpenApiSpecVersion version, bool hasNullInComposition = false)
+        private void SerializeNullable(IOpenApiWriter writer, OpenApiSpecVersion version)
         {
-            if (HasNullType || hasNullInComposition)
+            if (HasNullType)
             {
                 switch (version)
                 {
@@ -1050,61 +1039,6 @@ namespace Microsoft.OpenApi
                         writer.WriteProperty(OpenApiConstants.Nullable, true);
                         break;
                 }
-            }
-        }
-
-        /// <summary>
-        /// Processes a composition (oneOf or anyOf) for null types, filtering out null schemas and inferring common type.
-        /// </summary>
-        /// <param name="composition">The list of schemas in the composition.</param>
-        /// <returns>A tuple with the effective list, inferred type, and whether null is present in composition.</returns>
-        private static (JsonSchemaType? inferredType, bool hasNullInComposition)
-            ProcessCompositionForNull(IList<IOpenApiSchema>? composition)
-        {
-            if (composition is null || !composition.Any(static s => s.Type is JsonSchemaType.Null))
-            {
-                // Nothing to patch
-                return (null, false);
-            }
-
-            var nonNullSchemas = composition
-                .Where(static s => s.Type is null or not JsonSchemaType.Null)
-                .ToList();
-
-            if (nonNullSchemas.Count > 0)
-            {
-                JsonSchemaType commonType = 0;
-
-                foreach (var schema in nonNullSchemas)
-                {
-                    if (schema.Type.HasValue)
-                    {
-                        commonType |= schema.Type.Value & ~JsonSchemaType.Null;
-                    }
-                    else if (schema.Enum is { Count: > 0 })
-                    {
-                        foreach (var enumValue in schema.Enum.Where(x => x is not null))
-                        {
-                            var currentType = enumValue.GetValueKind() switch
-                            {
-                                JsonValueKind.Array => JsonSchemaType.Array,
-                                JsonValueKind.String => JsonSchemaType.String,
-                                JsonValueKind.Number => JsonSchemaType.Number,
-                                JsonValueKind.True or JsonValueKind.False => JsonSchemaType.Boolean,
-                                JsonValueKind.Null => (JsonSchemaType)0,
-                                _ => JsonSchemaType.Object,
-                            };
-
-                            commonType |= currentType;
-                        }
-                    }
-                }
-
-                return (commonType == 0 ? null : commonType, true);
-            }
-            else
-            {
-                return (null, true);
             }
         }
 
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -517,7 +517,11 @@ namespace Microsoft.OpenApi
             if (version == OpenApiSpecVersion.OpenApi3_0)
             {
                 // If we have a schema that's only just { "type": "null" }, we serialize it as enum with null value.
-                if (Type == JsonSchemaType.Null && OneOf is not { Count: > 0 } && AnyOf is not { Count: > 0 } && Enum is not { Count: > 0 })
+                if (Type == JsonSchemaType.Null &&
+                    OneOf is not { Count: > 0 } &&
+                    AnyOf is not { Count: > 0 } &&
+                    AllOf is not { Count: > 0 } &&
+                    Enum is not { Count: > 0 })
                 {
                     writer.WriteOptionalCollection(OpenApiConstants.Enum, s_singleNullElementList, (nodeWriter, s) => nodeWriter.WriteAny(s));
                 }
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -111,12 +111,18 @@ namespace Microsoft.OpenApi
         /// <inheritdoc />
         public JsonSchemaType? Type { get; set; }
 
+        /// <summary>
+        /// This property is internal and is only used to mark the "nullable" state when doing deserialization.
+        /// It's only set during deserialization to flag whether we encountered "nullable" or not.
+        /// And it's read:
+        /// 1. During FinalizeDeserialization to determine whether the Type should be changed to include null or not.
+        /// 2. When serializing nullable, if IsNullable was false and IsNullableFromDeserialization was true, we will still serialize "nullable".
+        /// </summary>
+        internal bool IsNullableFromDeserialization { get; set; }
+
         // x-nullable is filtered out by deserializers, but keep the check here in case it gets added from user code.
         internal bool IsNullable
-        {
-            get => field || HasTrueNullableExtension || (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null));
-            set => field = value;
-        }
+            => HasTrueNullableExtension || (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null));
 
         private bool HasTrueNullableExtension
             => Extensions is not null &&
@@ -798,11 +804,11 @@ namespace Microsoft.OpenApi
         {
             if (HasTrueNullableExtension)
             {
-                IsNullable = true;
                 Extensions!.Remove(OpenApiConstants.NullableExtension);
+                IsNullableFromDeserialization = true;
             }
 
-            if (IsNullable && Type is not null && Type != 0)
+            if (IsNullableFromDeserialization && Type is not null && Type != 0)
             {
                 Type |= JsonSchemaType.Null;
             }
@@ -1065,7 +1071,7 @@ namespace Microsoft.OpenApi
 
         private void SerializeNullable(IOpenApiWriter writer, OpenApiSpecVersion version, bool hasNullInComposition = false)
         {
-            if (IsNullable || hasNullInComposition)
+            if (IsNullable || IsNullableFromDeserialization || hasNullInComposition)
             {
                 switch (version)
                 {
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -114,12 +114,6 @@ namespace Microsoft.OpenApi
         private bool HasNullType
             => Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null);
 
-        private bool HasTrueNullableExtension
-            => Extensions is not null &&
-                Extensions.TryGetValue(OpenApiConstants.NullableExtension, out var nullExtRawValue) &&
-                nullExtRawValue is JsonNodeExtension { Node: JsonNode jsonNode } &&
-                jsonNode.GetValueKind() is JsonValueKind.True;
-
         /// <inheritdoc />
         public string? Const { get; set; }
 
@@ -788,18 +782,6 @@ namespace Microsoft.OpenApi
 
             // extensions
             writer.WriteExtensions(Extensions, OpenApiSpecVersion.OpenApi2_0);
-        }
-
-        internal void FinalizeDeserialization(OpenApiSpecVersion version)
-        {
-            if (version is OpenApiSpecVersion.OpenApi2_0 or OpenApiSpecVersion.OpenApi3_0 && HasTrueNullableExtension)
-            {
-                Extensions!.Remove(OpenApiConstants.NullableExtension);
-                if (Type is not null && Type != 0)
-                {
-                    Type |= JsonSchemaType.Null;
-                }
-            }
         }
 
         private void WriteFormatProperty(IOpenApiWriter writer)
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -111,19 +111,12 @@ namespace Microsoft.OpenApi
         /// <inheritdoc />
         public JsonSchemaType? Type { get; set; }
 
-        /// <summary>
-        /// This property is internal and is only used to mark the "nullable" state when doing deserialization.
-        /// It's only set during deserialization to flag whether we encountered "nullable" or not.
-        /// And it's read:
-        /// 1. During FinalizeDeserialization to determine whether the Type should be changed to include null or not.
-        /// 2. When serializing nullable, if IsNullable was false and IsNullableFromDeserialization was true, we will still serialize "nullable".
-        /// </summary>
-        internal bool IsNullableFromDeserialization { get; set; }
+        internal bool Nullable { get; set; }
+
+        private bool NullableOrHasNullType
+            => Nullable || (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null));
 
         // x-nullable is filtered out by deserializers, but keep the check here in case it gets added from user code.
-        internal bool IsNullable
-            => HasTrueNullableExtension || (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null));
-
         private bool HasTrueNullableExtension
             => Extensions is not null &&
                 Extensions.TryGetValue(OpenApiConstants.NullableExtension, out var nullExtRawValue) &&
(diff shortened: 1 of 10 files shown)


test/Microsoft.OpenApi.Readers.Tests/V3Tests/OpenApiSchemaTests.cs
```diff
@@ -532,8 +532,9 @@ public async Task SerializeSchemaWithNullableShouldSucceed()
         [Fact]
         public async Task SerializeSchemaWithOnlyNullableShouldSucceed()
         {
+            // NOTE: nullable property has no effect on the schema if type is not specified, so it is omitted in the serialized output.
             // Arrange
-            var expected = @"nullable: true";
+            var expected = @"{ }";
```
Youssef1313, line 537: Both behaviors are technically okay. But it's better to avoid emitting nullable when the spec says that it has no effect. This is less misleading for people reading the document and no familiar with the spec details.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -961,6 +961,11 @@ public async Task SerializeOneOfWithNullAsV3ShouldUseNullableAsync()
                 {
                   "type": "string",
                   "oneOf": [
+                    {
```
Youssef1313, line 963: The behavior here was previously buggy.

Even though "nullable" was set to true. A null value would have been still rejected because we won't match any of the "oneOf" schemas.

nullable only widens the "type" when present, but if the schema is still constrained to not allow null, a null would be rejected.

https://spec.openapis.org/oas/v3.0.4.html#fixed-fields-20

> This keyword only takes effect if `type` is explicitly defined within the same Schema Object. A `true` value indicates that both `null` values and values of the type specified by `type` are allowed. ***Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of `null` as a value***. A `false` value leaves the specified or default `type` unmodified. The default value is `false`.

The new behavior is more aligned with the object model given by the user.

The user specified a oneOf child schema with Type = null, and we serialize the semantically equivalent schema that's valid for OpenAPI 3.0 (which doesn't use draft 2020-12 and so can't use the straightforward `{ "type": "null" }` approach.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -110,12 +112,17 @@ public string? ExclusiveMinimum
         public JsonSchemaType? Type { get; set; }
 
         // x-nullable is filtered out by deserializers, but keep the check here in case it gets added from user code.
-        private bool IsNullable =>
-            (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null)) ||
-            Extensions is not null &&
-            Extensions.TryGetValue(OpenApiConstants.NullableExtension, out var nullExtRawValue) &&
-            nullExtRawValue is JsonNodeExtension { Node: JsonNode jsonNode } &&
-            jsonNode.GetValueKind() is JsonValueKind.True;
+        internal bool IsNullable
```
Youssef1313: When we are parsing a 3.0 document, we might encounter "nullable" property.

We might not know (yet) if this nullable property should have any effect or not (because we don't know if a future type property will be encountered or not).

As a result, we shouldn't be modifying `OpenApiSchema.Type` right away as soon as we see `"nullable": true` when deserializing.

To address this, `IsNullable` is made settable internally.

We will set it without modifying the type to indicate that `"nullable": true` was encountered while deserializing a document.

Then, at the end of the deserialization when we got all the information, we add "null" to `OpenApiSchema.Type` if nullable was set **and** we have a non-null type.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -518,56 +525,36 @@ private void SerializeInternal(IOpenApiWriter writer, OpenApiSpecVersion version
             writer.WriteOptionalCollection(OpenApiConstants.Enum, enumValue, (nodeWriter, s) => nodeWriter.WriteAny(s));
 
             // Handle oneOf/anyOf with null type for v3.0 downcast
-            IList<IOpenApiSchema>? effectiveOneOf = OneOf;
-            IList<IOpenApiSchema>? effectiveAnyOf = AnyOf;
```
Youssef1313, line 522: For correctness, we are no longer altering/modifying the oneOf/anyOf.
If we find oneOf/anyOf that has a child schema with null type, we will serialize it as enum with a single null element, which is valid for OpenAPI 3.0.

src/Microsoft.OpenApi/Reader/V2/OpenApiSchemaDeserializer.cs
```diff
@@ -268,16 +268,18 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
 
             ParseMap(jsonObject, schema, _openApiSchemaFixedFields, _openApiSchemaPatternFields, hostDocument, context);
 
-            if (schema.Extensions is not null && schema.Extensions.ContainsKey(OpenApiConstants.NullableExtension))
```
Youssef1313, line 271: ContainsKey here was wrong. The key could exist with "false" value. So I'm getting the value of the key and acting according to it.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -110,12 +112,17 @@ public string? ExclusiveMinimum
...
-            (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null)) ||
-            Extensions is not null &&
-            Extensions.TryGetValue(OpenApiConstants.NullableExtension, out var nullExtRawValue) &&
-            nullExtRawValue is JsonNodeExtension { Node: JsonNode jsonNode } &&
-            jsonNode.GetValueKind() is JsonValueKind.True;
+        internal bool IsNullable
+        {
+            get => field || HasNullableExtension || (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null));
+            set => field = value;
+        }
```
baywet: I'm not in favour of introducing a field. This means we have 2 things to keep in sync (the field, the type). Those getting out of sync for one reason or another will lead to bugs down the road.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -110,12 +112,17 @@ public string? ExclusiveMinimum
...
+
+        private bool HasNullableExtension
+            => Extensions is not null &&
+                Extensions.TryGetValue(OpenApiConstants.NullableExtension, out var nullExtRawValue) &&
+                nullExtRawValue is JsonNodeExtension { Node: JsonNode jsonNode } &&
+                jsonNode.GetValueKind() is JsonValueKind.True;
```
baywet: The name does not capture the fact that it also checks for the value being true.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -110,12 +112,17 @@ public string? ExclusiveMinimum
...
-            Extensions.TryGetValue(OpenApiConstants.NullableExtension, out var nullExtRawValue) &&
-            nullExtRawValue is JsonNodeExtension { Node: JsonNode jsonNode } &&
-            jsonNode.GetValueKind() is JsonValueKind.True;
+        internal bool IsNullable
+        {
+            get => field || HasNullableExtension || (Type.HasValue && Type.Value.HasFlag(JsonSchemaType.Null));
+            set => field = value;
+        }
+
+        private bool HasNullableExtension
```
baywet: we should make this internal, so you can reuse it in deserializers instead of duplicating the logic

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -995,14 +987,14 @@ private void SerializeAsV2(
             writer.WriteEndObject();
         }
 
-        private void SerializeTypeProperty(IOpenApiWriter writer, OpenApiSpecVersion version, JsonSchemaType? inferredType = null)
+        private bool SerializeTypeProperty(IOpenApiWriter writer, OpenApiSpecVersion version, JsonSchemaType? inferredType = null)
```
baywet: should this be name TrySerialize... if it doesn't always serialize?

baywet (changes requested):
Thanks for the contribution!

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: @baywet Would you be okay trying to split this into two properties? One that's settable by deserializers, and another that combines everything and is read-only?

Or how do you suggest to approach this?

See also https://github.com/microsoft/OpenAPI.NET/pull/2933#discussion_r3550969244 for why it would be wrong to attempt to set the type right away.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: I de-duplicated the logic in a different way. Kept this property private and de-duplicated the whole block of code in an internal method.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
baywet: I think this is confusing because of the "upgrade" path. Let's acknowledged that conversions are lossy by nature, and that we do everything we can to enable roundtrips with as good fidelity as possible.

## on roundtrips and upgrades

```yaml
nullable: true
type: string
```

should roundtrip in 3.0, but should also lead to 

```yaml
type: ["string", "null"]
```

In 3.1/3.2

## on the presence of nullable without type

[the OpenAPI 3.0 specification states](https://spec.openapis.org/oas/v3.0.4.html#fixed-fields-20)

> This keyword only takes effect if type is explicitly defined within the same Schema Object

It is NOT incorrect to have nullable: true with no type, it only means that keyword/field is meaningless.

While the library ensures the document is structured properly, it does not enforce any specific semantics (other than the logical conversions).

## on v2 serialization

v2 is a 12 years old standard, v3: 9 years old. While those still have a large user base due to some tooling never upgrading (some of it becoming abandonware), I don't believe we should over optimize for those. Especially if that comes at the potential expense of the more modern versions.

We do provide the nullable as an extension to provide more context to the consumer when something was downgraded. But honestly, I doubt that a tool that has not taken the time to upgrade in 9 years at this point will take the time to implement that extension. Yes we should roundtrip that extension when possible to avoid losses, but that'd a much lower priority.

## on v3.1/3.2

[the specification clearly states](https://json-schema.org/draft/2020-12/json-schema-validation#name-type) that having just `type: null` is perfectly fine.

## conclusion

With all that in mind. I don't want to see "multiple states" on the schema object that need to be kept in sync. Those schema objects can be initialized from deserialization, but also from the object model, and can be mutated afterwards. It'd be too easy for the state to become "out of sync" on one of the paths because the caller does not (and shouldn't know) the implementation details.

For deserialization: 
- When we encounter `nullable: true` in v2, or `x-nullable: true` in v3, we should set the type to "add null" (preserving any previous type value) as one of the values. (and drop the extension in v2 to avoid any confusion)
- When deserializing the type, it should preserve the type null value that might have been set before, this way we guard against any order of those properties.
- (for v3.1, 3.2, we should deserialize the type only)

For serialization:
- If the type contains null in v2, emit the `x-nullable: true` extension.
- If the type contains null in v3, emit the `nullable: true` extension.
- For any other version, just emit the type.

For the object model:
- we should only expose the object model as defined in 3.2, and have fields (as in state) for the things defined in 3.2.

Doing anything else with any/one/all of is a slippery slope, especially when we talk about upgrades etc. We should not be doing any of that.
This also assumes that we're serializing a "null" value for enum correctly, and not silently dropping it.

I hope this clarifies the cleanup at hand. Let me know what you think.

Youssef1313: @baywet Does it look better after the last refactoring?

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
```diff
@@ -385,15 +385,7 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
-            if (schema.Extensions is not null && schema.Extensions.ContainsKey(OpenApiConstants.NullableExtension))
-            {
-                if (schema.Type.HasValue)
-                    schema.Type |= JsonSchemaType.Null;
-                else
-                    schema.Type = JsonSchemaType.Null;
-
-                schema.Extensions.Remove(OpenApiConstants.NullableExtension);
-            }
+            schema.FinalizeDeserialization();
```
baywet: why are we implementing reading the extension at all in this version? it should only read nullable. The extension should only apply to v2

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
```diff
@@ -222,10 +225,7 @@ internal static partial class OpenApiV3Deserializer
                 {
                     if (bool.TryParse(n.GetScalarValue(), out var parsed) && parsed)
                     {
-                        if (o.Type.HasValue)
-                            o.Type |= JsonSchemaType.Null;
-                        else
-                            o.Type = JsonSchemaType.Null;
+                        o.IsNullableFromDeserialization = true;
```
baywet: we shouldn't maintain a state on the object for that. Revert this change. (original code before the PR)

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -109,13 +111,24 @@ public string? ExclusiveMinimum
...
         public JsonSchemaType? Type { get; set; }
 
+        /// <summary>
+        /// This property is internal and is only used to mark the "nullable" state when doing deserialization.
+        /// It's only set during deserialization to flag whether we encountered "nullable" or not.
+        /// And it's read:
+        /// 1. During FinalizeDeserialization to determine whether the Type should be changed to include null or not.
+        /// 2. When serializing nullable, if IsNullable was false and IsNullableFromDeserialization was true, we will still serialize "nullable".
+        /// </summary>
+        internal bool IsNullableFromDeserialization { get; set; }
```
baywet: we shouldn't have this field/state

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
```diff
@@ -332,14 +333,7 @@ internal static partial class OpenApiV31Deserializer
...
                 {
-                    var nullable = bool.Parse(value);
-                    if (nullable) // if nullable, convert type into an array of type(s) and null
-                    {
-                        if (o.Type.HasValue)
-                            o.Type |= JsonSchemaType.Null;
-                        else
-                            o.Type = JsonSchemaType.Null;
-                    }
+                    o.IsNullableFromDeserialization = bool.Parse(value);
```
baywet: revert this change (code from before the PR)

src/Microsoft.OpenApi/Reader/V31/OpenApiSchemaDeserializer.cs
```diff
@@ -476,15 +470,7 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
-        if (schema.Extensions is not null && schema.Extensions.ContainsKey(OpenApiConstants.NullableExtension))
-        {
-            if (schema.Type.HasValue)
-                schema.Type |= JsonSchemaType.Null;
-            else
-                schema.Type = JsonSchemaType.Null;
-
-            schema.Extensions.Remove(OpenApiConstants.NullableExtension);
-        }
+        schema.FinalizeDeserialization();
```
baywet: we shouldn't parse the extension in 3.1

src/Microsoft.OpenApi/Reader/V32/OpenApiSchemaDeserializer.cs
```diff
@@ -332,14 +333,7 @@ internal static partial class OpenApiV32Deserializer
...
                 {
-                    var nullable = bool.Parse(value);
-                    if (nullable) // if nullable, convert type into an array of type(s) and null
-                    {
-                        if (o.Type.HasValue)
-                            o.Type |= JsonSchemaType.Null;
-                        else
-                            o.Type = JsonSchemaType.Null;
-                    }
+                    o.IsNullableFromDeserialization = bool.Parse(value);
```
baywet: revert this change (before the PR)

src/Microsoft.OpenApi/Reader/V32/OpenApiSchemaDeserializer.cs
```diff
@@ -476,15 +470,7 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
-        if (schema.Extensions is not null && schema.Extensions.ContainsKey(OpenApiConstants.NullableExtension))
-        {
-            if (schema.Type.HasValue)
-                schema.Type |= JsonSchemaType.Null;
-            else
-                schema.Type = JsonSchemaType.Null;
-
-            schema.Extensions.Remove(OpenApiConstants.NullableExtension);
-        }
+        schema.FinalizeDeserialization();
```
baywet: we shouldn't parse the extension in 3.1

baywet (commented):
Getting closer (I'm only focusing my review on schema object field, and deserialization so far)

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
Youssef1313: @baywet This is pre-existing behavior, I think. I'm not sure why it existed.

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
Youssef1313: I'm not following how I should approach fixing the current bugs in the area then.

Basically, we cannot just say "we have nullable set to true, we we are going to expand the Type with JsonSchemaType.Null". That in itself was a bug.

src/Microsoft.OpenApi/Reader/V32/OpenApiSchemaDeserializer.cs
Youssef1313: @baywet This is a pre-existing behavior, I think, right?

baywet: yes, and I think it was an error to have it in the first place.

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
baywet: yes, and I think it was a mistake to begin with. Added here #1820

Youssef1313: There is a test asserting the current behavior: `LoadSchemaWithNullableExtensionAsV31Works`.
I can change the test and behavior though.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: > When we encounter nullable: true in v2, or x-nullable: true in v3, we should set the type to "add null" (preserving any previous type value) as one of the values. (and drop the extension in v2 to avoid any confusion)

I'm assuming you swapped the version for where nullable and x-nullable belong. But when we encounter it, I do **not** think we should add "null" to the type right away. We might not know **yet** if the user has "type" defined in the first place. What I want to make sure is, if I have the following two schemas in 3.0:

```json
{
  "nullable": true,
  "type": "string"
}
```

and

```json
{
  "nullable": true
}
```

In the first schema, I expect the OpenApiSchema.Type to be `JsonSchemaType.String | JsonSchemaType.Null`, but for the second, OpenApiSchema.Type should remain `null`.

So, we don't always just add `JsonSchemaType.Null` to the Type, as it's incorrect, because nullable only has a meaning when `type` is explicitly defined, and so deserialization should take that into account.

Youssef1313: If we want to avoid the additional state/field in OpenApiSchema object model, we can instead use the extensions dictionary during deserialization to have an internal extension that we "encountered" nullable. And in FinalizeDeserialization, we always remove the internal extension, and at that point we know whether or not we should add `JsonSchemaType.Null`.

baywet:
To recap our call, and previous discussions

There shouldn't be any field/state for nullability on the OpenAPISchema model, only in the deserialization phases.
We shouldn't be "moving schemas arounds" (except for v2, because that behaviour has been here for 10+ years at this point).

## OpenAPI 3.1/3.2

serialization and deserialization should be symmetrical, serialize type null as part of the type property, array or single.

## OpenAPI v3.0

### serialization

Serialize type null (no other types) as enum null because nullable true by itself is ignored and the resulting schema would validate anything.
Serialize type null + something as type something, nullable true

### deserialization

Deserialize type something to something
Deserialized nullable true to a marker in the metadata dictionary.

Once both properties are deserialized:
- if nullable true is by itself, drop it, because it only applies when a type is also provided
- if nullable true is with a type something, add null to the type array 

## Swagger 2.0

### serialization

always serialize type null (with or without something else) as x-nullable: true. That's because :
- one/anyOf is not supported by this version of the standard
- we're already picking the first one/anyOf entry, that bechaviour has been here for a long time, we shouldn't change it
- enum null can't be composed, is therefore useless, so we should diverge from OpenAPI 3.0 behaviour.

### Deserialization

Do the same mark and resolve behaviour as V3.0 for consistency, but for the x-nullable extension.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -517,57 +514,26 @@ private void SerializeInternal(IOpenApiWriter writer, OpenApiSpecVersion version
...
-                hasNullInComposition |= nullInOneOf;
-                inferredType = inferredOneOf ?? inferredType;
-                (effectiveAnyOf, var inferredAnyOf, var nullInAnyOf) = ProcessCompositionForNull(AnyOf);
-                hasNullInComposition |= nullInAnyOf;
-                inferredType = inferredAnyOf ?? inferredType;
-
-                hasOneOfNullAndSingleEnumWith3_0 = nullInOneOf && effectiveOneOf is { Count: 1 } &&
-                    effectiveOneOf[0].Enum is { Count: > 0 };
+                // If we have a schema that's only just { "type": "null" }, we serialize it as enum with null value.
+                if (Type == JsonSchemaType.Null && OneOf is not { Count: > 0 } && AnyOf is not { Count: > 0 } && Enum is not { Count: > 0 })
```
baywet: should this also include all of?

src/Microsoft.OpenApi/Reader/V2/OpenApiSchemaDeserializer.cs
```diff
@@ -268,14 +268,15 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
+            var extensions = schema.Extensions;
+            if (extensions?.TryGetValue(OpenApiConstants.NullableExtension, out var nullExtRawValue) == true &&
```
baywet: use metadata instead of extensions

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: Yup.

src/Microsoft.OpenApi/Reader/V2/OpenApiSchemaDeserializer.cs
```diff
@@ -210,6 +210,29 @@ internal static partial class OpenApiV2Deserializer
...
+                        }
+                        else
+                        {
+                            // We mirror the same behavior of OpenAPI 3.0 "nullable" which is a type modifier.
+                            // It only has effect if there is a "Type" set.
+                            // Given that we don't have a Type set yet, we want to keep track of it until the end of deserialization.
+                            // If, at a later point during deserialization, Type was set, we apply nullable to it.
+                            // Otherwise, we throw it away.
+                            o.Metadata ??= new Dictionary<string, object>();
+                            o.Metadata["encounteredNullable"] = true;
```
baywet: can we use a constant to avoid duplication?

baywet (approved):

PR #2936, opened 2026-07-13
Differentiate unset value from null value in OpenApiSchema.Const

### LINKED ISSUE #2935: OpenApiSchema.Const cannot differentiate between "null" and unset value.
When serializing an OpenApiSchema object as 3.1+, the Const property is ignored completely if it's set to null. However, 'null' value is allowed by the draft 2020-12 spec.

There should be two concepts, first is "unset value" and second is "value set to null".

In addition to that, when serializing Const as 2.0 or 3.0, the same bug exist, but also an empty string value is considered as "unset". Empty string is always clearly a value that should be serialized and should never be treated as "unset".

I imagine this might need a new public API like `OpenApiUnsetValues.UnsetString` so users can do:

```csharp
var schema = new OpenApiSchema();
schema.Const = null; // serializes to null.
schema.Const = OpenApiUnsetValues.UnsetString; // doesn't serialize.
```

The default value of `Const` will be `OpenApiUnsetValues.UnsetString`.

I'm not sure if there are any other properties having similar issues as well.

@baywet any thoughts?

Youssef1313:
Fixes #2935


Youssef1313:
Benchmark is reporting a perf regression. @baywet I assume it's because we are introducing a new field (bool) to the objects which is increasing the object size? Should it be fine?

baywet (dismissed):
Thanks for the contribution! Would you mind updating the performance reports please?

Youssef1313:
@copilot Can you update the performance report?

baywet (approved):

PR #2938, opened 2026-07-13
PERF: Reduce boxing allocations in GetScalarValue

Youssef1313:
`GetScalarValue` can be dealing with a JsonNode that's wrapping a JsonElement. Doing `GetValue<object>` will return a boxed JsonElement, which is unnecessary allocation. Then we could also force creating a string when we deal with boolean in many cases.

This PR extracts a separate helper for bool values, which is more efficient and is more straightforward to use.

And for the string-based GetScalarValue, try to get the value as string right away first to avoid JsonElement boxing.

### DIFF
--- a/src/Microsoft.OpenApi/Reader/JsonNodeHelper.cs
+++ b/src/Microsoft.OpenApi/Reader/JsonNodeHelper.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Text.Json;
 using System.Text.Json.Nodes;
 
 namespace Microsoft.OpenApi.Reader
@@ -169,6 +170,51 @@ namespace Microsoft.OpenApi.Reader
             }
 
             return Convert.ToString(scalarNode.GetValue<object>(), CultureInfo.InvariantCulture);
+        }
+
+        public static bool GetScalarBoolValue(this JsonNode? node)
+        {
+            var scalarNode = node is JsonValue value ? value : throw new OpenApiException("Expected scalar value.");
+            return scalarNode.GetValue<bool>();
+        }
+
+        public static int GetScalarIntValue(this JsonNode? node)
+        {
+            var scalarNode = node is JsonValue value && value.GetValueKind() == JsonValueKind.Number ? value : throw new OpenApiException("Expected numeric scalar value.");
+
+            if (scalarNode.TryGetValue<int>(out var intValue))
+            {
+                return intValue;
+            }
+            else if (scalarNode.TryGetValue<long>(out var longValue))
+            {
+                return (int)longValue;
+            }
+            else if (scalarNode.TryGetValue<decimal>(out var decimalValue))
+            {
+                return (int)decimalValue;
+            }
+
+            return Convert.ToInt32(scalarNode.GetValue<object>());
+        }
+
+        public static uint GetScalarUIntValue(this JsonNode? node)
+        {
+            var scalarNode = node is JsonValue value && value.GetValueKind() == JsonValueKind.Number ? value : throw new OpenApiException("Expected numeric scalar value.");
+            if (scalarNode.TryGetValue<uint>(out var uintValue))
+            {
+                return uintValue;
+            }
+            else if (scalarNode.TryGetValue<ulong>(out var ulongValue))
+            {
+                return (uint)ulongValue;
+            }
+            else if (scalarNode.TryGetValue<decimal>(out var decimalValue))
+            {
+                return (uint)decimalValue;
+            }
+
+            return Convert.ToUInt32(scalarNode.GetValue<object>());
         }
 
         public static string? GetReferencePointer(this JsonObject jsonObject)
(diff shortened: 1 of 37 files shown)


src/Microsoft.OpenApi/Reader/V32/OpenApiV32Deserializer.cs
```diff
@@ -144,10 +144,9 @@ private static IOpenApiExtension LoadExtension(string name, JsonNode node, Parsi
             return node.GetScalarValue();
         }
 
-        private static bool? LoadBool(JsonNode node)
+        private static bool LoadBool(JsonNode node)
         {
-            var value = node.GetScalarValue();
-            return value is not null ? bool.Parse(value) : null;
```
baywet, line 150: here we're changing the behaviour. We're assigning false to values that are nullable instead of null, aren't we?

src/Microsoft.OpenApi/Reader/JsonNodeHelper.cs
```diff
@@ -159,9 +160,63 @@ public static Dictionary<string, HashSet<T>> CreateArrayMap<T>(this JsonNode? no
         {
             var scalarNode = node is JsonValue value ? value : throw new OpenApiException("Expected scalar value.");
 
+            // It's much more efficient to call scalarNode.TryGetValue<string> than to call scalarNode.GetValue<object>() and then convert to string.
+            // When asking for "object" type, if scalarNode is JsonValueOfElement (internal type in STJ), we will get back
+            // a boxed JsonElement (paying the cost of unnecessary boxing allocation), and we then call JsonElement.ToString.
+            // So, we first check if we can get the string value directly, and only if that fails, we fallback to the expensive code.
```
baywet, line 166: I think that one reason why we had to do that was to avoid mangling like default, numbers with formats, etc... While I do appreciate the performance improvements, I don't want it to come at the price of a regression.

src/Microsoft.OpenApi/Reader/JsonNodeHelper.cs
```diff
@@ -159,9 +160,63 @@ public static Dictionary<string, HashSet<T>> CreateArrayMap<T>(this JsonNode? no
...
+
+            if (scalarNode.TryGetValue<int>(out var intValue))
+            {
+                return intValue;
+            }
+            else if (scalarNode.TryGetValue<long>(out var longValue))
+            {
+                return (int)longValue;
+            }
+            else if (scalarNode.TryGetValue<decimal>(out var decimalValue))
```
baywet: can we get other numeric types?

baywet (commented):
Thanks for the contribution!

src/Microsoft.OpenApi/Reader/V32/OpenApiV32Deserializer.cs
Youssef1313: This is used for `$vocabulary`. https://json-schema.org/draft/2020-12/json-schema-core#section-8.1.2

> The values of the object properties MUST be booleans.

src/Microsoft.OpenApi/Reader/JsonNodeHelper.cs
Youssef1313: `TryGetValue<string>` is never going to throw or do something strange. It will return true if the node really represents a string. For example, if you have any numbers in JSON like `"maximum": 10`, `TryGetValue<string>` will return false because the node doesn't represent a string. And then we fallback to the original implementation.

I done it this way explicitly because I saw that some numerics are read raw as strings that way.

src/Microsoft.OpenApi/Reader/JsonNodeHelper.cs
Youssef1313: Decimal in most cases will be coming through the YML library being used to convert YML to JSON.
I'm not sure if we can get other numeric types but I haven't seen it when trying to run the test suite. However, to reduce any risks, I kept the `GetValue<object>` fallback.

Edit: It's not the library. It's our code to handle YML. I tried to adjust it so that it handles numeric in a more correct way.

src/Microsoft.OpenApi/Reader/V32/OpenApiV32Deserializer.cs
baywet: ah this is only used at that place (vocab), I thought it was used multiple places and changed the semantics by replacing the default from null to false. Let's drop the method all together and inline it (both for 3.1 and 3.2)

baywet: (same for laod string above)

src/Microsoft.OpenApi/Reader/JsonNodeHelper.cs
baywet: This is what I'm referring to https://github.com/microsoft/OpenAPI.NET/commit/c17a87e8d90eb5b4ad71f45615fc5df1ddd314cd

I think the object convert was introduced to avoid formatting issues for defaults and other properties when using types like dates and such.

PR #2952, opened 2026-07-14
fix: validate required properties of security scheme before serialization

### LINKED ISSUE #1843: Incorrect default value for OpenApiSecurityScheme.Location when using "http" type with "bearer" scheme
**Describe the bug**
When defining an OpenApiSecurityScheme of type http with the scheme bearer, the default value for Location is set to Query instead of Header. According to the OpenAPI Specification, header should be the implied default when the type is http and the scheme is bearer.

**OpenApi File To Reproduce**
```yaml
openapi: 3.0.0
info:
  title: Minimal API
  version: 1.0.0
paths:
  /example:
    get:
      summary: Example endpoint
      security:
        - Bearer: []
      responses:
        '200':
          description: Successful response
components:
  securitySchemes:
    Bearer:
      type: http
      scheme: bearer
 ```

**Expected behavior**
The default value for OpenApiSecurityScheme.Location should be Header when the type is http and the scheme is bearer, aligning with the OpenAPI Specification’s default behavior.

**Additional context**
This issue causes incorrect behavior when generating clients or code based on the OpenAPI definition, as the security token is expected to be sent as a query parameter instead of the Authorization header.

baywet:
fixes #1843

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiSecurityScheme.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSecurityScheme.cs
@@ -97,6 +97,7 @@ namespace Microsoft.OpenApi
             Action<IOpenApiWriter, IOpenApiSerializable> callback)
         {
             Utils.CheckArgumentNull(writer);
+            EnsureRequiredPropertiesAreSet(version);
 
             writer.WriteStartObject();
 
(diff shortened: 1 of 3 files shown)


peombwa (approved):

baywet:

baywet:

baywet:

PR #2960, opened 2026-07-15
Do not ignore multiple types when serializing to 3.0

### LINKED ISSUE #2939: Consider representing multiple types in 3.0 using anyOf or oneOf
See https://github.com/dotnet/aspnetcore/issues/66304 for context.

Youssef1313:
Fixes #2939

- The change to `ToSingleIdentifier` is only a simplification to make it easier to read.
- Type serialization is now split to two path to be easier to read, especially that 3.0 needs more parameters/info than 2.0.
- The behavior of 2.0 is maintained as-is, no change there
- Previously, in 3.0, we ignored when `Type` had multiple values (e.g, String and Integer). This is now handled using `oneOf` or `anyOf` (whichever isn't used by the current schema already). If both are used, we ignore as it we used to in the past.
- The general idea here is:
    1. We take the `Type`.
    2. We remove `Null` from it (if it exists).
    3. If, after removing null, we have a single type, we simply emit it, and conditionally emit nullable: true depending on whether null was existing. This is the same as the existing behavior.
    4. If, after removing null, we have more than one type, we create a oneOf/anyOf with children schemas handling each type individually.
- The reason around all that is that because type only supports a single value when using OpenApi 3.0.

For context: https://spec.openapis.org/oas/v3.0.4.html#json-schema-keywords

> The following keywords are taken from the JSON Schema definition but their definitions were adjusted to the OpenAPI Specification.
> type - Value MUST be a string. Multiple types via an array are not supported.


test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1968,6 +1968,482 @@ public async Task SerializeNullableTypeWith3_1_And_Later(OpenApiSpecVersion vers
             Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result)));
         }
 
+        [Fact]
+        public async Task SerializeMultipleTypesWith3_0_NotIncludingNull_NotIncludingOneOfOrAnyOf()
+        {
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String | JsonSchemaType.Integer,
+            };
```
Youssef1313, line 2139: Before my change, this was resulting in an empty schema, which will match anything.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1968,6 +1968,482 @@ public async Task SerializeNullableTypeWith3_1_And_Later(OpenApiSpecVersion vers
...
+
+            Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result)));
+        }
+
+        [Fact]
+        public async Task SerializeMultipleTypesWith3_0_IncludingNullAndTwoTypes_IncludingAnyOfButNotOneOf()
+        {
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String | JsonSchemaType.Integer | JsonSchemaType.Null,
```
Youssef1313, line 2732: Before my change, the type was ignored, and only anyOf was emitted.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1968,6 +1968,482 @@ public async Task SerializeNullableTypeWith3_1_And_Later(OpenApiSpecVersion vers
...
+
+            Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result)));
+        }
+
+        [Fact]
+        public async Task SerializeMultipleTypesWith3_0_NotIncludingNull_IncludingAnyOfButNotOneOf()
+        {
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String | JsonSchemaType.Integer,
```
Youssef1313, line 2239: Before my change, the type was ignored and only anyOf was emitted.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1968,6 +1968,482 @@ public async Task SerializeNullableTypeWith3_1_And_Later(OpenApiSpecVersion vers
...
+
+            Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result)));
+        }
+
+        [Fact]
+        public async Task SerializeMultipleTypesWith3_0_IncludingNullAndTwoTypes_IncludingOneOfButNotAnyOf()
+        {
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String | JsonSchemaType.Integer | JsonSchemaType.Null,
```
Youssef1313, line 2661: Before my change, the type was ignored and only oneOf was emitted.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1968,6 +1968,482 @@ public async Task SerializeNullableTypeWith3_1_And_Later(OpenApiSpecVersion vers
...
+
+            Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result)));
+        }
+
+        [Fact]
+        public async Task SerializeMultipleTypesWith3_0_NotIncludingNull_IncludingOneOfButNotAnyOf()
+        {
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String | JsonSchemaType.Integer,
```
Youssef1313, line 2175: Before my change, only oneOf was emitted.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1968,6 +1968,482 @@ public async Task SerializeNullableTypeWith3_1_And_Later(OpenApiSpecVersion vers
...
+
+            Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result)));
+        }
+
+        [Fact]
+        public async Task SerializeMultipleTypesWith3_0_IncludingNullAndTwoTypes_NotIncludingOneOfOrAnyOf()
+        {
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String | JsonSchemaType.Integer | JsonSchemaType.Null,
```
Youssef1313, line 2617: Before my change, an empty schema was generated, allowing anything to match.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1968,6 +1968,497 @@ public async Task SerializeNullableTypeWith3_1_And_Later(OpenApiSpecVersion vers
...
+
+            Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result)));
+        }
+
+        [Fact]
+        public async Task SerializeMultipleTypesWith3_0_NotIncludingNull_IncludingBothOneOfAndAnyOf()
+        {
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String | JsonSchemaType.Integer,
```
Youssef1313, line 2303: In this test, the behavior is the same as in the past. We are not respecting Type, and I can't see an easy way to respect it.

It's technically possible to emit "allOf" that contains a schema representing the types as "oneOf" or "anyOf" schema, and then another schema that wraps the existing anyOf/oneOf. But that's a major re-write I didn't want to introduce.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaTests.cs
```diff
@@ -1968,6 +1968,497 @@ public async Task SerializeNullableTypeWith3_1_And_Later(OpenApiSpecVersion vers
...
+
+            Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expected), JsonNode.Parse(result)));
+        }
+
+        [Fact]
+        public async Task SerializeMultipleTypesWith3_0_IncludingNullAndTwoTypes_IncludingBothOneOfAndAnyOf()
+        {
+            var schema = new OpenApiSchema()
+            {
+                Type = JsonSchemaType.String | JsonSchemaType.Integer | JsonSchemaType.Null,
```
Youssef1313, line 2803: Same here. This wasn't (and is still not) respected. We could respect it but it will be a more bigger re-write.

baywet (dismissed):
I believe this kind of normalization should not be done during serialization, but rather through a walker which would be called before serialization.

See #808 and other related issues.

Case in point, we've had to revert some of the normalization done for null/nullable (to oneOf) in #2933

Youssef1313:
> I believe this kind of normalization should not be done during serialization, but rather through a walker which would be called before serialization.

Given the investigations I did for #2967, I think it's fine for us to continue ensuring, as much as possible, that we produce semantically equivalent document for all versions. The cases presented in #2967 fall into two groups:

- user bug (e.g, the Swashbuckle bug, or the JsonApiDotNetCore bug)
- deserialization bugs on our side, which we can simply fix.

I don't believe there is a need for a new public API to do the semantic transformations, and I don't think all transformations we might want to do in future will be representable in the object model. So we will end up with logic scattered between dedicated walkers and in the serialization itself. We shouldn't need to pay additional performance cost for deep cloning the whole document as well.

In all cases, we can always decide to move the logic to a walker in the future, if really necessary.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -1006,31 +1006,106 @@ private void SerializeAsV2(
...
+        }
 
-            if (typeToUse is null)
+        /// <summary>
+        /// Tries to serialize the "type" property for OpenAPI v3 and later versions.
+        /// </summary>
+        /// <returns>
+        /// true if the Type was serializable using "type" property, and false if
+        /// it serialized using anyOf/oneOf or if it couldn't be serialized at all.
+        /// </returns>
```
gavinbarron: Returns void, not bool

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -1006,31 +1006,106 @@ private void SerializeAsV2(
...
+                        writer.WriteOptionalCollection(OpenApiConstants.OneOf, ConstructChildSchemasForTypes(arrayWithoutNull, hasNull), callback);
                         return;
                     }
-                    break;
-                default:
-                    WriteUnifiedSchemaType(typeToUse.Value, writer);
+                }
+                else
+                {
+                    writer.WriteProperty(OpenApiConstants.Type, arrayWithoutNull[0].ToSingleIdentifier());
```
gavinbarron, line 1069: There's a potential IndexOutOfBoundsException here given that this branch is entered when there are 0 or 1 elements in the `arrayWithoutNull` 
This was previously guarded via `typeWithoutNull != 0`

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaV30CompatibilityTests.cs
```diff
@@ -118,28 +118,34 @@ public async Task NullableTypeAsV3ShouldRoundTripCorrectly()
         }
 
         [Fact]
-        public async Task SerializeMultipleNonNullTypesAsV3OmitsType()
+        public async Task SerializeMultipleNonNullTypesAsV3DoesNotOmitType()
```
gavinbarron, line 121: It seems we're missing a test case here for types with constraints

e.g. `{ "anyOf": [ {"type":"string","format":"email","maxLength":10}, {"type":"integer","minimum":0} ] }`

We should ensure that those constraints are correctly handled.

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
```diff
@@ -420,7 +420,47 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
                 schema.Type = JsonSchemaType.Null;
             }
 
+            if (schema.Type is null)
+            {
+                if (schema.AnyOf is not null &&
+                    schema.AnyOf.All(childSchema => childSchema is OpenApiSchema schema && DoesSchemaRepresentSingleType(schema)))
```
gavinbarron: **nit:** shadowing here hurts readability
```suggestion
                    schema.AnyOf.All(s => s is OpenApiSchema child && DoesSchemaRepresentSingleType(child)))
```

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
```diff
@@ -420,7 +420,47 @@ public static IOpenApiSchema LoadSchema(JsonNode node, OpenApiDocument hostDocum
...
+                {
+                    JsonSchemaType types = GetAllTypes(schema.AnyOf);
+                    schema.AnyOf = null;
+                    schema.Type = types;
+                }
+                else if (schema.OneOf is not null &&
+                    schema.OneOf.All(childSchema => childSchema is OpenApiSchema schema && DoesSchemaRepresentSingleType(schema)))
```
gavinbarron: **nit**

```suggestion
                    schema.OneOf.All(s=> s is OpenApiSchema child && DoesSchemaRepresentSingleType(child)))
```

gavinbarron (commented):
a few small nits, and one additional test case we need to ensure works as intended

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: Fixed.

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
Youssef1313: Fixed.

test/Microsoft.OpenApi.Tests/Models/OpenApiSchemaV30CompatibilityTests.cs
Youssef1313: There was a bug in deserialization indeed. Fixed now.

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
Youssef1313: Fixed.

src/Microsoft.OpenApi/Reader/V3/OpenApiSchemaDeserializer.cs
Youssef1313: Fixed.

gavinbarron (approved):

PR #2965, opened 2026-07-17
Route NuGet restore through the CFS central package feed

gavinbarron:
Routes CI package restore through the authenticated `GraphDeveloperExperiences_Public` Azure Artifacts feed (network isolation / CFS compliance).

- Adds `NuGetAuthenticate@1` and generates a `nuget.config` with `<clear/>` + the central feed as the only source.

Note: requires a pipeline run to validate against the feed.

### DIFF
--- a/.azure-pipelines/ci-build.yml
+++ b/.azure-pipelines/ci-build.yml
@@ -79,6 +79,20 @@ extends:
             checkLatest: true
 
         # Build the Product project
+        - task: NuGetAuthenticate@1
+          displayName: 'Authenticate to Azure Artifacts'
+
+        - pwsh: |
+            @"
+            <?xml version="1.0" encoding="utf-8"?>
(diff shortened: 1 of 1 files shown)


.azure-pipelines/ci-build.yml
```diff
@@ -79,6 +79,20 @@ extends:
             checkLatest: true
 
         # Build the Product project
+        - task: NuGetAuthenticate@1
```
baywet, line 82: Shouldn't this be after setting the feed? So it can use that information to auth.

gavinbarron: Strangely not for NuGet. This sets up the credential provider independently of the specific feed, which works as the feed is from inside the same ADO tenant.

baywet (approved):

PR #2970, opened 2026-07-20
fix: serialize examples as extension in v2/v3

### LINKED ISSUE #2969: OpenApiSchema plural examples are not serialized for OpenAPI 2.0/3.0
## Summary

`OpenApiSchema.Examples` currently serializes only for OpenAPI 3.1+ as the JSON Schema `examples` keyword. For OpenAPI 2.0 and 3.0, plural schema examples are omitted instead of being preserved as a JSON Schema compatibility extension.

## Expected behavior

Schema-level example fields should serialize as follows:

- `example` singular serializes as `example` for OpenAPI 2.0, 3.0, and 3.1+.
- `examples` plural serializes as `examples` for OpenAPI 3.1+.
- `examples` plural serializes as `x-jsonschema-examples` for OpenAPI 2.0 and 3.0.

## Current behavior

- `example` singular is serialized as `example` in all versions.
- `examples` plural is written from `OpenApiSchema.WriteJsonSchemaKeywords`, which is only called for OpenAPI 3.1+.
- The OpenAPI 3.0 JSON Schema compatibility path writes other `x-jsonschema-*` keywords, but does not write `Examples`.
- The OpenAPI 2.0 schema serialization path also does not write `Examples`.

## Notes

The existing `x-examples` constant/behavior appears to apply to parameter/response example maps, not schema-level JSON Schema `examples`. Schema-level plural examples should use `x-jsonschema-examples` when down-converting to OpenAPI 2.0 or 3.0.

baywet:
fixes #2969

### DIFF
--- a/src/Microsoft.OpenApi/Models/OpenApiConstants.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiConstants.cs
@@ -806,6 +806,11 @@ namespace Microsoft.OpenApi
         public const string ExamplesExtension = "x-examples";
 
         /// <summary>
+        /// Extension: x-jsonschema-examples
+        /// </summary>
+        public const string JsonSchemaExamplesExtension = "x-jsonschema-examples";
+
+        /// <summary>
         /// Field: version3_0_0
         /// </summary>
         public static readonly Version version3_0_0 = new(3, 0, 0);
--- a/src/Microsoft.OpenApi/Models/OpenApiConstants.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiConstants.cs
@@ -771,6 +771,11 @@ namespace Microsoft.OpenApi
         public const string OAuth2MetadataUrl = "oauth2MetadataUrl";
 
         /// <summary>
+        /// Extension: x-oai-oauth2-metadata-url
+        /// </summary>
+        internal const string OAuth2MetadataUrlExtension = "x-oai-oauth2-metadata-url";
+
+        /// <summary>
         /// Field: OpenIdConnectUrl
         /// </summary>
         public const string OpenIdConnectUrl = "openIdConnectUrl";
--- a/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
+++ b/src/Microsoft.OpenApi/Models/OpenApiSchema.cs
@@ -746,6 +746,7 @@ namespace Microsoft.OpenApi
             writer.WriteOptionalObject(OpenApiConstants.IfExtension, If, callback);
             writer.WriteOptionalObject(OpenApiConstants.ThenExtension, Then, callback);
             writer.WriteOptionalObject(OpenApiConstants.ElseExtension, Else, callback);
+            writer.WriteOptionalCollection(OpenApiConstants.JsonSchemaExamplesExtension, Examples, (nodeWriter, s) => nodeWriter.WriteAny(s));
         }
 
         internal void WriteAsItemsProperties(IOpenApiWriter writer)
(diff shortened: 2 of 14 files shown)


src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -746,6 +746,7 @@ private void WriteV3CompatibilityKeywords(IOpenApiWriter writer, Action<IOpenApi
             writer.WriteOptionalObject(OpenApiConstants.IfExtension, If, callback);
             writer.WriteOptionalObject(OpenApiConstants.ThenExtension, Then, callback);
             writer.WriteOptionalObject(OpenApiConstants.ElseExtension, Else, callback);
+            writer.WriteOptionalCollection(OpenApiConstants.JsonSchemaExamplesExtension, Examples, (nodeWriter, s) => nodeWriter.WriteAny(s));
```
Youssef1313: Could we write the first element of the array as `example` as best effort to remain following the 2.0 and 3.0 way?

baywet: do you mean in case example doesn't have something already?

Youssef1313: Yup. I think the following behavior is best for 2.0/3.0

- `Example` not set, and `Examples` has a single example -> serialize that single example in `example` property. Don't serialize the extension.
- `Example` not set, and `Examples` has more than a single example -> serialize the first example in `example` property, and serialize the extension to cover the remaining examples.

In addition, I'd expect the `Example` property to be marked `Obsolete` to reflect it being deprecated.

https://spec.openapis.org/oas/v3.2.0.html#fixed-fields-20

> Deprecated: The example field has been deprecated in favor of the JSON Schema examples keyword. Use of example is discouraged, and later versions of this specification may remove it.

baywet: Let's tackle the field being marked as obsolete in a separate PR for now.

@copilot implement the following additional serialization updates for v3.0 and 2.0 when example is not set:
- if examples has a single item, serialize that as example property, do not serialize any extension
- if examples has multiple items, serialize the first one as example property, serialize the remaining ones as the extension

baywet: Here is the issue for the obsolete stuff https://github.com/microsoft/OpenAPI.NET/issues/2972

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
```diff
@@ -746,6 +749,7 @@ private void WriteV3CompatibilityKeywords(IOpenApiWriter writer, Action<IOpenApi
             writer.WriteOptionalObject(OpenApiConstants.IfExtension, If, callback);
             writer.WriteOptionalObject(OpenApiConstants.ThenExtension, Then, callback);
             writer.WriteOptionalObject(OpenApiConstants.ElseExtension, Else, callback);
+            writer.WriteOptionalCollection(OpenApiConstants.JsonSchemaExamplesExtension, GetCompatibilityExamplesExtension(), (nodeWriter, s) => nodeWriter.WriteAny(s));
```
Youssef1313, line 752: Is this going to write the extension for `>= 3.1` ?

src/Microsoft.OpenApi/Models/OpenApiConstants.cs
```diff
@@ -805,6 +805,11 @@ public static class OpenApiConstants
         /// </summary>
         public const string ExamplesExtension = "x-examples";
 
+        /// <summary>
+        /// Extension: x-jsonschema-examples
+        /// </summary>
+        public const string JsonSchemaExamplesExtension = "x-jsonschema-examples";
```
Youssef1313, line 821: Do we need to handle this new extension when **deserializing**?

baywet: I think we have inconsistent implementations here across those extensions for deserialization. Some do, some do not. In this specific instance, I'm ok with reading from the extension with older versions of the specification.

@copilot implement that change

src/Microsoft.OpenApi/Models/OpenApiSchema.cs
baywet: the method we're in only gets called for v3.0 and v2.0

Youssef1313 (approved):
Aside from handling the extension during deserialization, this looks good to me 🎉 🎉

peombwa (approved):

PR #2985, opened 2026-07-22
feat: add SerializeAsJsonAsync overloads accepting OpenApiJsonWriterS…

### LINKED ISSUE #2798: Support ability to return compact JSON format when calling SerializeAsJsonAsync
I would like to obtain a compact json format when calling OpenApiSerializableExtensions.SerializeAsJsonAsync

That could be done by adding an optional OpenApiJsonWriterSettings parameter to the SerializeAsJsonAsync method (with Terse = true if compact format needed)

Currently i can parse and re-serialize json to obtain a minified format but it would be better to just be able to have the right format immediately.

**Additional context**
I'm able to propose a pull request if needed.

Mahdigln:
## Description
Add `SerializeAsJsonAsync` and `SerializeAsync` overloads that accept `OpenApiJsonWriterSettings`, enabling callers to produce compact (minified) JSON output without having to parse and re-serialize the result.

## Type of Change

## Related Issue(s)
Closes #2798

## Changes Made
- Add `SerializeAsJsonAsync<T>(Stream, OpenApiSpecVersion, OpenApiJsonWriterSettings)` overload
- Add `SerializeAsJsonAsync<T>(OpenApiSpecVersion, OpenApiJsonWriterSettings)` overload returning `string`
- Add `SerializeAsync<T>(OpenApiSpecVersion, string, OpenApiWriterSettings?)` overload returning `string`
- Register the three new signatures in `PublicAPI.Unshipped.txt`

## Testing
- [x] All existing tests pass

Tests cover:
- `Terse = true` via stream overload produces compact JSON
- `Terse = true` via string overload produces compact JSON
- `Terse = false` via string `SerializeAsync` overload produces pretty-printed JSON

## Checklist

## Versions applicability

## Additional Notes
The three new overloads intentionally omit the `CancellationToken` parameter to avoid RS0026/RS0027 analyzer violations that arise from adding optional-parameter overloads alongside existing shipped ones. Callers who need both custom settings and cancellation support can use the existing `SerializeAsync(Stream, OpenApiSpecVersion, string, OpenApiWriterSettings?, CancellationToken)` overload directly.

`#pragma warning disable RS0027` is applied to four existing shipped methods where the analyzer fires a false positive: the new overloads use distinct types (`OpenApiJsonWriterSettings` vs `CancellationToken`) so no call-site ambiguity exists.


src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -21,12 +21,28 @@ public static class OpenApiSerializableExtensions
...
 
+        /// <summary>
+        /// Serialize the <see cref="IOpenApiSerializable"/> to the Open API document (JSON) using the given stream, specification version and settings.
+        /// </summary>
+        /// <typeparam name="T">the <see cref="IOpenApiSerializable"/></typeparam>
+        /// <param name="element">The Open API element.</param>
+        /// <param name="stream">The output stream.</param>
+        /// <param name="specVersion">The Open API specification version.</param>
+        /// <param name="settings">Settings controlling JSON output, including <see cref="OpenApiJsonWriterSettings.Terse"/> for compact formatting.</param>
+        public static Task SerializeAsJsonAsync<T>(this T element, Stream stream, OpenApiSpecVersion specVersion, OpenApiJsonWriterSettings settings)
```
baywet: this should also accept a cancellation token

baywet (changes requested):
Thanks for the contribution!

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -21,12 +21,29 @@ public static class OpenApiSerializableExtensions
...
+        /// <summary>
+        /// Serialize the <see cref="IOpenApiSerializable"/> to the Open API document (JSON) using the given stream, specification version and settings.
+        /// </summary>
+        /// <typeparam name="T">the <see cref="IOpenApiSerializable"/></typeparam>
+        /// <param name="element">The Open API element.</param>
+        /// <param name="stream">The output stream.</param>
+        /// <param name="specVersion">The Open API specification version.</param>
+        /// <param name="settings">Settings controlling JSON output, including <see cref="OpenApiJsonWriterSettings.Terse"/> for compact formatting.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        public static Task SerializeAsJsonAsync<T>(this T element, Stream stream, OpenApiSpecVersion specVersion, OpenApiJsonWriterSettings settings, CancellationToken cancellationToken)
```
baywet: the cancellation token should be optional

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -142,15 +161,33 @@ public static Task SerializeAsync<T>(this T element, IOpenApiWriter writer, Open
...
+        /// Serializes the <see cref="IOpenApiSerializable"/> to the Open API document as a string in JSON format using the given settings.
+        /// </summary>
+        /// <typeparam name="T">the <see cref="IOpenApiSerializable"/></typeparam>
+        /// <param name="element">The Open API element.</param>
+        /// <param name="specVersion">The Open API specification version.</param>
+        /// <param name="settings">Settings controlling JSON output, including <see cref="OpenApiJsonWriterSettings.Terse"/> for compact formatting.</param>
+        public static Task<string> SerializeAsJsonAsync<T>(
+            this T element,
+            OpenApiSpecVersion specVersion,
+            OpenApiJsonWriterSettings settings)
```
baywet: missing the cancellation token here

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -195,5 +234,30 @@ public static async Task<string> SerializeAsync<T>(
...
+        /// <typeparam name="T">the <see cref="IOpenApiSerializable"/></typeparam>
+        /// <param name="element">The Open API element.</param>
+        /// <param name="specVersion">The Open API specification version.</param>
+        /// <param name="format">Open API document format.</param>
+        /// <param name="settings">Provide configuration settings for controlling writing output.</param>
+        public static async Task<string> SerializeAsync<T>(
+            this T element,
+            OpenApiSpecVersion specVersion,
+            string format,
+            OpenApiWriterSettings? settings)
```
baywet: missing the cancellation token here

test/Microsoft.OpenApi.Tests/Extensions/OpenApiSerializableExtensionsTests.cs
```diff
@@ -90,4 +91,60 @@ public async Task UsesTheTerseOutputInformationFromSettingsNoSettings()
...
+        {
+            Name = "param1",
+            In = ParameterLocation.Query,
+            Schema = new OpenApiSchema { Type = JsonSchemaType.String }
+        };
+
+        var settings = new OpenApiJsonWriterSettings { Terse = true };
+
+        using var stream = new MemoryStream();
+        await parameter.SerializeAsJsonAsync(stream, OpenApiSpecVersion.OpenApi3_1, settings, CancellationToken.None);
```
baywet: we should be passing the test context cancellation token instead

baywet (changes requested):

src/Microsoft.OpenApi/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,6 @@
 #nullable enable
+const Microsoft.OpenApi.OpenApiConstants.JsonSchemaExamplesExtension = "x-jsonschema-examples" -> string!
+const Microsoft.OpenApi.OpenApiConstants.OaiLicenseIdentifier = "x-oai-license-identifier" -> string!
```
baywet:
```suggestion
```

baywet:
This will be pending on #3006, sorry I thought we already did the upgrade. Please wait for the PR to merge, and then rebase your PR.

baywet:
@Mahdigln can you please merge main into your branch?

Mahdigln:
> Mahdigln can you please merge main into your branch?

Done! I've merged the latest vnext into my branch and pushed the changes.

Mahdigln:
> This will be pending on #3006, sorry I thought we already did the upgrade. Please wait for the PR to merge, and then rebase your PR.

No problem! I'll wait for #3006 to merge and then rebase.

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
```diff
@@ -21,12 +21,31 @@ public static class OpenApiSerializableExtensions
         /// <param name="stream">The output stream.</param>
         /// <param name="specVersion">The Open API specification version.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
+#pragma warning disable RS0027 // The settings overload below has the same parameter count but different types; no ambiguity exists.
```
baywet: those suppressions are binary safe, but they are not source safe for a specific edge case.

```csharp
await doc.SerializeAsJsonAsync(version, default);
await doc.SerializeAsync(version, OpenApiConstants.Json, default);
await doc.SerializeAsJsonAsync(stream, version, default);
```

Would now fail to compile because of the ambiguity. We'll need to review the approach here.

baywet (changes requested):

src/Microsoft.OpenApi/Extensions/OpenApiSerializableExtensions.cs
baywet: I've noodled on the problem for the last 15 minutes or so. Also came across [this documentation](https://github.com/dotnet/roslyn/blob/main/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md)

Given the current state of things, I don't think we have a ton of options:
- Add the settings as the LAST parameter, that's ugly but should work. And we can clean it up in the next major version.
- Use a different method name. Ugly as well, but should work. We could even mark the existing methods as obsolete for now. And then we could clean up the API surface in the next major version.
- Conditionally add those overloads to net8+ only, and add [the OverloadResolutionPriorityAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.overloadresolutionpriorityattribute?view=net-10.0) on the older methods. And clean all of this up in the next major version.
- (not merging this pull request at all, unless we get overwhelming feedback this is needed, and leave the implementation to the next major version).

Thoughts?

baywet: For option 3, @youssef1313 correctly pointed out the attribute is only available starting with net9.0, so we'd need to add a new target to the project (most likely net10.0) and make the conditional compilation use that. This increases the cost of that solution (not only in terms of maintenance, but for the consumers when pulling down the package, etc...)

Mahdigln: I went with option 1 — placed `CancellationToken` before `settings` so callers must pass it explicitly. This removes all ambiguity. The new signatures would be:

```csharp
SerializeAsJsonAsync(Stream stream, OpenApiSpecVersion specVersion, CancellationToken cancellationToken, OpenApiJsonWriterSettings settings)

SerializeAsJsonAsync(OpenApiSpecVersion specVersion, CancellationToken cancellationToken, OpenApiJsonWriterSettings settings)

SerializeAsync(OpenApiSpecVersion specVersion, string format, CancellationToken cancellationToken, OpenApiWriterSettings? settings)
```

Does this work for you?

baywet: I think that's probably the least worst option. Thank you for making the changes!

Can you also add a `// TODO: ` comment to fix that on the next major release please?

I'll also ask a colleague's input to make sure "it's ok" from an API design perspective.

Youssef1313: It violates https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068, not sure if you're okay with such violation.

baywet: Note, we could create the overloads as experimental until the next major version so people who take a dependency on it know it'll change.

baywet: @youssef1313 Thank you for the additional information.

So our ONLY *clean* option would be to use a different method name at this point?

PR #3025, opened 2026-08-15
feat(document): add GetOperationById to search operations across path…

### LINKED ISSUE #1654: Get operation by ID or ref

I have an `OpenApiLink` object, which has either an `OperationId` or an `OperationRef`, but I can't find a way to access the `Operation` object that these refer to.

Either an `Operation` property on the `OpenApiLink` object that holds the "resolved" operation, or methods like `openApiDocument.GetOperationById(string operationId)` and `openApiDocument.GetOperationByRef(string operationRef)`.

For operation IDs, constructing a dictionary of operations, keyed by operation ID, and using that. I don't know about operation refs though.

**Additional context**

n/a

Mahdigln:
## Description

Adds a `GetOperationById(string operationId)` method to `OpenApiDocument` that searches for an operation by its `operationId` across both `Paths` and `Webhooks`.

Closes #1654

## Type of Change

## Related Issue(s)

Closes #1654

## Changes Made

- Added `GetOperationById(string operationId)` method to `OpenApiDocument`
- Method searches both `Paths` and `Webhooks` to cover the full document
- Resolves operations through `OpenApiPathItemReference` (when the reference is resolved)
- Added `GetOperationById` entry to `PublicAPI.Unshipped.txt`

## Testing

- [x] All existing tests pass

Tests cover:
- Returns matching operation from `Paths`
- Returns matching operation from `Webhooks`
- Returns `null` when not found
- Case-sensitive matching (`getUser` ≠ `GetUser`)
- Resolves operation through a `$ref` path item
- Returns first match when duplicate `operationId` exists (spec violation)
- Skips unresolved path item references gracefully
- Throws `ArgumentNullException` for null or empty `operationId`

## Checklist

## Versions applicability

## Additional Notes

`GetOperationByRef` (resolving `operationRef` via JSON Pointer) is not included in this PR and can be addressed in a follow-up.


src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -857,6 +857,33 @@ static bool AddToDictionary<TValue>(IDictionary<string, TValue> dict, string key
...
+
+            var allPathItems = Webhooks is not null
+                ? Paths.Values.Concat(Webhooks.Values)
+                : Paths.Values;
+
+            foreach (var pathItem in allPathItems)
+            {
+                if (pathItem.Operations is not null)
+                {
+                    foreach (var operation in pathItem.Operations.Values)
```
baywet: this is running at o(n) (n being the number of operations) and going to be really slow for larger API descriptions. Could you look into optimizing the search please?

baywet (changes requested):
Thanks for the contribution!

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
Mahdigln: Thanks for the feedback! To address the O(n) concern, I'm thinking of building a lazy-initialized Dictionary<string, OpenApiOperation> index on the first call to GetOperationById, making subsequent lookups O(1). The trade-off is that the cache becomes stale if Paths or Webhooks are mutated after the index is built — but since OpenApiDocument is typically read-only after parsing, this seems acceptable.

Does this approach work for you, or do you have a different optimization in mind?

baywet: That should work, but before we read into that index, we should be able to correlate whether the sources have changed since the last time we built the index and rebuild it if required.

Mahdigln: Thanks for the clarification! To implement change detection, I need to understand how to track mutations to Paths and Webhooks. A few options I'm considering:

Count-based: Compare Paths.Count and Webhooks?.Count — simple but misses replacements (same count, different items)
Version counter: Increment a counter whenever Paths or Webhooks is mutated — but OpenApiPaths doesn't expose change notifications today
Wrap collections: Replace OpenApiPaths with an observable/versioned collection that notifies the document on change

Do you have a preferred approach, or is there existing infrastructure in the codebase I should leverage?

baywet: have you considered storing the hash code for those values next to the index. And ahead of querying the index, if the current hash codes don't match, recompute + update the values. ??

Mahdigln: Before implementing, one question: how should the hash be computed for Paths and Webhooks? Should I use the existing GetHashCodeAsync on the document, or compute something lighter like combining Paths.Count, Webhooks?.Count, and the hash codes of the keys?

baywet: since we're doing a simple comparison in memory, can't we simply use [GetHashCode](https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-10.0) ?

Mahdigln: I want to flag a potential issue: GetHashCode() on Dictionary/OpenApiPaths uses reference-based hashing in .NET, so it won't detect mutations to the existing collection (e.g. adding or removing a path). It would only detect if Paths or Webhooks is replaced with a new object entirely. Is this level of detection sufficient, or do we need content-aware change detection?

baywet: right, but couldn't we aggregate the hash over the entries? using a comparer for example?

Mahdigln: I see two possible approaches for the hash:

1. **Hash over path keys only** (O(1) per lookup after first build): aggregate hash from `Paths` keys and `Webhooks` keys. Detects path additions/removals but misses operationId changes within an existing path item.

2. **Hash over keys + operationIds** (still O(n) per lookup): more thorough change detection but essentially the same complexity as the original linear search, just with caching benefit for repeated calls on an unchanged document.

Which approach do you prefer?

baywet: Thank you for the additional information.

Right, because one could mutate the operation id without mutating the dictionaries, which gets us back to square 1 in terms of lookup complexity to build the index.

Thanks for getting to the bottom of this discussion with me. Resolving.

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -857,6 +857,33 @@ static bool AddToDictionary<TValue>(IDictionary<string, TValue> dict, string key
...
+        /// <summary>
+        /// Finds an operation in the document by its operation ID.
+        /// </summary>
+        /// <param name="operationId">The operation ID to search for.</param>
+        /// <returns>The matching <see cref="OpenApiOperation"/>, or <see langword="null"/> if not found.</returns>
+        public OpenApiOperation? GetOperationById(string operationId)
+        {
+            Utils.CheckArgumentNullOrEmpty(operationId);
+
+            var allPathItems = Webhooks is not null
```
baywet: This is going to lead to the assignment of a new collection on each lookup. Could we instead:
1. move the looping to a dedicated private method
2. loop over both collections instead of assembling an aggregate

?

Mahdigln: Done, both changes applied

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -857,6 +857,42 @@ static bool AddToDictionary<TValue>(IDictionary<string, TValue> dict, string key
...
+        }
+
+        private IEnumerable<OpenApiOperation> GetAllOperations()
+        {
+            foreach (var pathItem in Paths.Values)
+            {
+                if (pathItem.Operations is null) continue;
+                foreach (var operation in pathItem.Operations.Values)
+                    yield return operation;
+            }
```
baywet: this is duplicative with what we have below. My initial comment was about making a method to reuse for that aspect.

src/Microsoft.OpenApi/Models/OpenApiDocument.cs
```diff
@@ -857,6 +857,43 @@ static bool AddToDictionary<TValue>(IDictionary<string, TValue> dict, string key
...
+
+        private static IEnumerable<OpenApiOperation> GetOperationsFromPathItems(IDictionary<string, IOpenApiPathItem> pathItems)
+        {
+            foreach (var pathItem in pathItems.Values)
+            {
+                if (pathItem.Operations is null) continue;
+                foreach (var operation in pathItem.Operations.Values)
+                    yield return operation;
+            }
+        }
```
baywet, line 887:
```suggestion
        public OpenApiOperation? GetOperationById(string operationId)
        {
            Utils.CheckArgumentNullOrEmpty(operationId);

            return GetOperationByIdFromPathItems(Paths, operationId) ??
                (Webhooks is not null ? 
                    GetOperationByIdFromPathItems(Webhooks, operationId) : null);
        }

        private static OpenApiOperation? GetOperationByIdFromPathItems(IDictionary<string, IOpenApiPathItem> pathItems, string operationId)
        {
            foreach (var pathItem in pathItems.Values)
            {
                if (pathItem.Operations is null) continue;
                foreach (var operation in pathItem.Operations.Values)
                {
                    if (string.Equals(operation.OperationId, operationId, StringComparison.Ordinal))
                        return operation;
                }
            }
            return null;
        }
```

Sorry if I wasn't clear with all my comments, this is what I meant: single looping sequence for each of the collections, no new collection assignment, no duplication of the filtering logic.

baywet (approved):

Mahdigln:
> Thanks for the contribution!

Thank you for the thorough review and guidance, Vincent! I learned a lot from the feedback.

PR #3040, opened 2026-08-19
fix: harden yaml parsing (v1 port)

gavinbarron:
port of #3027 to v1


src/Microsoft.OpenApi.Readers/InputLimitTextReader.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+            else if (value <= 0x7FF)
+            {
+                AddBytes(2);
+            }
+            else
+            {
+                AddBytes(3);
+            }
+        }
```
baywet, line 108: I can't say I'm a big fan of all the magic numbers here with no doc comment at the top of the method. This is going to be difficult to maintain in case anything is wrong in here.

src/Microsoft.OpenApi.Readers/YamlHelper.cs
```diff
@@ -2,32 +2,50 @@
...
 
-        public static YamlNode ParseYamlString(string yamlString)
+        public static YamlNode ParseYamlString(
+            string yamlString,
+            ParsingContext context = null)
         {
-            var reader = new StringReader(yamlString);
-            var yamlStream = new YamlStream();
-            yamlStream.Load(reader);
+            var settings = new OpenApiReaderSettings();
```
baywet: can't we get the settings from the callers instead of instantiating a new one here?

src/Microsoft.OpenApi/Validations/Rules/OpenApiSchemaRules.cs
```diff
@@ -4,6 +4,8 @@
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Properties;
 using System.Collections.Generic;
+using System.Linq;
```
baywet: I think changes to this class are a leak from the other PR in the works. Revert for now?

test/Microsoft.OpenApi.Tests/Validations/OpenApiSchemaValidationTests.cs
```diff
@@ -323,5 +323,238 @@ public void ValidateOneOfSchemaPropertyNameContainsPropertySpecifiedInTheDiscrim
             //Assert
             errors.Should().BeEmpty();
         }
+
+        [Fact]
+        public void ValidateSchemaDiscriminatorReturnsExistingErrorForSelfCycleWithoutValidDiscriminator()
```
baywet: same here, came from the other PR

baywet (changes requested):

src/Microsoft.OpenApi.Readers/InputLimitTextReader.cs
gavinbarron: good callout, fixed

test/Microsoft.OpenApi.Tests/Validations/OpenApiSchemaValidationTests.cs
gavinbarron: removed for now

src/Microsoft.OpenApi/Validations/Rules/OpenApiSchemaRules.cs
gavinbarron: removed for now

baywet (dismissed):

baywet (approved):

PR #3042, opened 2026-08-19
fix: circular ref guard

gavinbarron:
port of #3039 to v1


src/Microsoft.OpenApi/Validations/Rules/OpenApiSchemaRules.cs
```diff
@@ -89,28 +91,20 @@ public static class OpenApiSchemaRules
         /// between other schemas which may satisfy the payload description.</param>
         public static bool ValidateChildSchemaAgainstDiscriminator(OpenApiSchema schema, string discriminatorName)
         {
-            if (!schema.Required?.Contains(discriminatorName) ?? false)
+            if (discriminatorName == null)
```
baywet:
```suggestion
            if (string.IsNullOrEmpty(discriminatorName))
```

baywet (approved):

PR #3044, opened 2026-08-20
parallelize Descriptions setup, fix response leak, use typed stream keys

Imadnajam:
- Dispose HttpResponseMessage in LoadFromUrlAsync (was leaking on every download)
- Run assembly reads and GHES downloads concurrently in GlobalSetup via Task.WhenAll;
  aggregate into _streams afterward to avoid concurrent Dictionary writes
- Replace string-keyed _streams with a private DescriptionSource enum for
  compile-time-checked, pre-sized lookups
- Drop the async/await wrapper on each [Benchmark] method so MemoryDiagnoser
  measures ParseDocumentAsync, not an extra state machine
- Derive the four pinned GHES URLs from one commit SHA constant
- Add an HttpClient timeout; simplify Assembly access; align field naming
Descriptions


baywet (approved):
Thanks for the contribution!

